PR #29, opened 2021-05-10
[Manta] #13 CI/CD - Create initial github action to check compilation and formatting

### LINKED ISSUE #13: CI/CD
Put CI/CD in the pipeline.

ghzlatarev:
closes #13 CI/CD : Create check.yml for compilation and formatting github action

* The check.yml first runs yamllint
* Then installs rustup and runs the following commands:
     cargo check --release
     cargo  fmt --all -- --check
     cargo clippy --release
     cargo test --release


.github/workflows/check.yml
```diff
@@ -0,0 +1,42 @@
+name: Check Set-Up & Build
+
+# Controls when the action will run.
+on:
+  # Triggers the workflow on push or pull request events but only for the manta (default) branch.
+  pull_request:
+    branches: [ manta ]
+  #push:
+  #  branches: [ manta ]
```
ghzlatarev: Do we want to enable an action when pushing to branches ?
The action is triggered when you push during a pull request so this seems enough for me.

.github/workflows/check.yml
```diff
@@ -0,0 +1,42 @@
+name: Check Set-Up & Build
```
ghzlatarev, line 3: https://github.com/substrate-developer-hub/substrate-node-template/blob/master/.github/workflows/check.yml
FYI using this one as a template.

ghzlatarev:
Not sure about the enhancement label that I used.
Maybe we should create a CI/CD label?

ghzlatarev:
The check is running but I'm not actually sure it will block me from merging if it fails. @grenade is there anything else required besides the check.yml file ?

Dengjianping:
I found that CI skips wasm build. 
```
SKIP_WASM_BUILD=1 cargo check --release
```
@stechu Do you think should skip this build, only for native build?
Maybe clippy check and testing should be supported.
```
cargo clippy --release
cargo test --release
```

ghzlatarev (commented):
> Maybe clippy check and testing should be supported.
> 
> ```
> cargo clippy --release
> cargo test --release
> ```

Agreed. Let me add them to this workflow and see how much time they add. If it becomes too much we can do this in a separate job maybe.

stechu:
yeah, we don't want to skip wasm in the build.

.github/workflows/check.yml
```diff
@@ -0,0 +1,61 @@
...
+
+name: Check Set-Up & Build
+
+# Controls when the action will run.
+on:
+  # Triggers the workflow on push or pull request events
+  # but only for the calamari (default) branch.
+  pull_request:
+    branches: [manta]
+  # push:
```
stechu: shall we also run the action again after pushed?

.github/workflows/check.yml
```diff
@@ -0,0 +1,61 @@
...
+          rustup component add rustfmt
+      - name: Check Formatting
+        run: |
+          cargo +nightly fmt --all -- --check
+      - name: Check Clippy
+        run: |
+          cargo +nightly clippy --release
+      - name: Check Build
+        run: |
+          SKIP_WASM_BUILD=1 cargo check --release
```
stechu: remove `SKIP_WASM_BUILD=1`

stechu: should use `cargo +nightly` for all command or simply set nightly as the default in rust toolchain.

stechu (changes requested):

.github/workflows/check.yml
ghzlatarev: I think it is enough that a push during a pull request will trigger the action. We never want to push directly to manta branch anyway.
Compilaiton is 15 minutes.
Clippy is 20
Tests are 30

ghzlatarev: Actually since we never should push directly into the manta branch, it won't make too much of a difference. And it will force the person to take a moment and think about why they are pushing directly without a pull request. So might as well enable it.

stechu:
@ghzlatarev  now we get some warnings from github:
githubw

Any idea?

stechu:
In general, LGTM. We could merge it once we figured out the warning: either fix it or we are confident that that it is fine.

ghzlatarev:
> In general, LGTM. We could merge it once we figured out the warning: either fix it or we are confident that that it is fine.

Yeah let's keep it like this for readability as mentioned in the other comment.

By the way I already fixed this in the code but how do I resolve it in the Conversation so I can merge ?
how_to_resolve
It seems stuck now...

ghzlatarev:
> @ghzlatarev now we get some warnings from github:
> githubw
> 
> Any idea?

"on" is a truthy value that is implicitly converted to "true" (what the warning is indicating) 
I'd keep it like this for readability.

stechu (approved):

grenade (approved):
lgtm

Issue #30, opened 2021-05-10
Remove clippy warnings


Dengjianping:
When I tried clippy check, there're some clippy warnings, we need to make clippy happy.
```
cargo +nightly clippy
```
And we might have a new convention that developers need do clippy check before push code.
And clippy can teach us how to write elegant code.

zhenfeizhang:
I second this.

zhenfeizhang:
Current issue with Clippy is caused by `pallet-manta-pay`.
This issue https://github.com/Manta-Network/pallet-manta-pay/issues/19 tracks the problem

zhenfeizhang:
Problem solved with #31

PR #31, opened 2021-05-10
[manta] [ready to merge] #30 fix linter

zhenfeizhang:
you should be able to run
```
cargo +nightly clippy
```
without any warnings or errors.

The complains are not fixed, but suppressed.
See comments in the code.

### DIFF
--- a/runtimes/manta/runtime/src/lib.rs
+++ b/runtimes/manta/runtime/src/lib.rs
@@ -1,3 +1,6 @@
+// suppressing linter on impl From
+// linter will trigger a complain on `frame-support-3.0.0` which we cannot fix
+#![allow(clippy::from_over_into)]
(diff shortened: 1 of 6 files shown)


runtimes/manta/runtime/src/lib.rs
```diff
@@ -350,7 +353,7 @@ impl pallet_timestamp::Config for Runtime {
 }
 
 parameter_types! {
-	pub const NativeTokenExistentialDeposit: u128 = 1 * MA;
+	pub const NativeTokenExistentialDeposit: u128 = MA;
```
stechu: Don't really like the change here, `1 * MA ` is more readable.

stechu: @zhenfeizhang  Can we disable Clippy for this line?

PR #34, opened 2021-05-11
[Manta] Update types for fixing balances showing on front-end

Dengjianping:
This is a workaround for fixing a problem on showing balances on front-end.
And it happened to someone, see this chat history:
https://matrix.to/#/!HzySYSaIhtyWrwiwEV:matrix.org/$162053226462902xjzST:matrix.org?via=matrix.parity.io&via=matrix.org&via=corepaper.org


stechu (approved):
Thanks for the fix! I update the token symbol as well.

PR #44, opened 2021-05-13
[Manta]Impl manta runtime mock, runtime constants tests(WIP)

### LINKED ISSUE #36: Fee related unit tests
Fee related unit tests, could be similar to:
https://github.com/paritytech/substrate/blob/master/bin/node/runtime/src/impls.rs

Dengjianping:
Still need to consider add more tests. And figure out what these runtime constants tests mean.


runtimes/manta/Cargo.toml
```diff
@@ -29,7 +29,6 @@ serde_json = "1.0"
 frame-benchmarking = '3.1.0'
 frame-benchmarking-cli = '3.0.0'
 pallet-transaction-payment-rpc = '3.0.0'
-pallet-contracts-rpc = '3.0.0'
```
stechu, line 32: good catch!

stechu:
@Dengjianping  any update on this?

Dengjianping:
There're some calculations I need to figure out in these test cases, some updates will be submitted before this weekend ends, I think.

stechu:
one additional thing that may worth testing is the existential deposit.

runtimes/manta/runtime/src/tests.rs
```diff
@@ -74,17 +74,47 @@ impl ExtBuilder {
...
+	let initial_authorities = vec![(
+		ALICE,
+		ALICE,
+		hex!["9becad03e6dcac03cee07edebca5475314861492cdfc96a2144a67bbe9699332"].unchecked_into(),
+		hex!["6e7e4eb42cbd2e0ab4cae8708ce5509580b8c04d11f6758dbf686d50fe9f9106"].unchecked_into(),
+	)];
+	let root_key = ALICE;
+	let stash = 9 * cMA;
+
+	let init_balances: Vec<(AccountId, Balance)> = vec![(ALICE, 99 * cMA), (BOB, 99 * cMA)];
```
stechu: here should be based on existential deposit rather than hard coded. Maybe we should put existential deposit as a configuration constant?

stechu (changes requested):
LGTM in general. Thanks for the PR!

runtimes/manta/runtime/src/tests.rs
Dengjianping: Removed this hardcoded value, just check it again.

stechu (approved):

PR #45, opened 2021-05-15
[Manta] Runtime benchmark

stechu:
This PR add the runtime benchmark function to Manta runtime.

### DIFF
--- a/runtimes/manta/runtime/Cargo.toml
+++ b/runtimes/manta/runtime/Cargo.toml
@@ -57,7 +57,7 @@ pallet-utility = { default-features = false, version = '3.0.0' }
 
 
 # Manta
-pallet-manta-pay = { git = 'https://github.com/manta-network/pallet-manta-pay', default-features = false }
+pallet-manta-pay = { branch="calamari", git = 'https://github.com/Manta-Network/pallet-manta-pay', default-features = false }
 manta-primitives = {path="../primitives", default-features = false }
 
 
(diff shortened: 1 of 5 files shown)


runtimes/manta/runtime/Cargo.toml
```diff
@@ -57,7 +57,7 @@ pallet-utility = { default-features = false, version = '3.0.0' }
 
 
 # Manta
-pallet-manta-pay = { branch = 'calamari', git = 'https://github.com/Manta-Network/pallet-manta-pay', default-features = false }
+pallet-manta-pay = { git = 'https://github.com/manta-network/pallet-manta-pay', default-features = false }
```
zhenfeizhang: can we revert this change and specify the branch?

zhenfeizhang: actually this pointer points to the master branch and will bug...
looks like cargo will pull from master rather than default

```
warning: fetching `master` branch from `https://github.com/manta-network/pallet-manta-pay` but the `HEAD` reference for this repository is not the `master` branch. This behavior will change in Cargo in the future and your build may break, so it's recommended to place `branch = "master"` in Cargo.toml when depending on this git repository to ensure that your build will continue to work.
```

stechu: I thought the `Cargo.lock` will specify the commit.

zhenfeizhang (approved):

PR #50, opened 2021-05-19
[Manta] Nodes deployment scripts (WIP)

Dengjianping:
Minimize the mistakes while configure keystore manually for manta node.


stechu:
This works well for deploying on a single machine, not so well in cloud :disappointed:

stechu:
close for now.

PR #57, opened 2021-05-20
[Manta] Update WeightInfo for FRAME pallets to use default substrate weights

### LINKED ISSUE #49: Benchmark and update weights of pallets used in Manta runtime
* Benchmark and update weights of pallets used in Manta runtime
* We need a weights folder where we can place the generated weights.rs file for each pallet. Like https://github.com/AcalaNetwork/Acala/tree/master/runtime/acala/src
* This is needed because generated weight functions are dependent on runtime configurations.
* Upload a script for each pallet.

ghzlatarev:
References #49 

* Update WeightInfo for FRAME pallets to use default substrate weights

### DIFF
--- a/runtimes/manta/runtime/src/lib.rs
+++ b/runtimes/manta/runtime/src/lib.rs
@@ -216,7 +216,7 @@ impl pallet_session::Config for Runtime {
 	type SessionHandler = <opaque::SessionKeys as OpaqueKeys>::KeyTypeIdProviders;
 	type Keys = opaque::SessionKeys;
 	type DisabledValidatorsThreshold = DisabledValidatorsThreshold;
-	type WeightInfo = ();
+	type WeightInfo = pallet_session::weights::SubstrateWeight<Runtime>;
 }
 
 impl pallet_session::historical::Config for Runtime {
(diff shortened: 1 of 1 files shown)


ghzlatarev:
> References #49
> 
> * Update WeightInfo for FRAME pallets to use default substrate weights

So I don't think anyone overrides the FRAME pallets weight functions. For reference:
1. https://github.com/PureStake/moonbeam/blob/master/runtime/moonriver/src/lib.rs
2. https://github.com/hicommonwealth/edgeware-node/blob/master/node/runtime/src/lib.rs
3. Acala actually have a weights folder but not for FRAME pallets - https://github.com/AcalaNetwork/Acala/tree/master/runtime/acala/src/weights

EDIT: asked in the substrate chat, seems our understanding is correct.
1 guy suggests that the default values are probably good enough for most runtime developers.
Will update if there are contradicting replies.

ghzlatarev:
Damn I wish we could link issues to PRs, without closing the issue once the PR is merged.
Now we gotta reopen them manually...

stechu:
Good catch!

stechu (approved):

PR #60, opened 2021-05-22
[Manta] update pallet-manta-pay

stechu:
Update `pallet-manta-pay` to the newest version.
Currently, runtime-benchmarks fails. I suspect because of newly added dependencies (`manta-asset` and `manta-crypto`).


zhenfeizhang:
fixed with https://github.com/Manta-Network/pallet-manta-pay/pull/46

PR #102, opened 2021-06-24
[Manta-PC]Bump manta-pc's deps to polkadot-v0.9.5(WIP)

### LINKED ISSUE #101: Test with Acala parachain
Acala is using polkadot-v0.9.5, so we have to follow it.

Dengjianping:
Closes #101 
Deps commits:
substrate: 9c572625f6557dfdb19f47474369a0327d51dfbc
polkadot: 3ce09da502dd74d88f24d76f579d19fd4e6c5624
cumulus: 05ab7a377d2c3b73566c0c3bd41b003486d6913b


Cargo.toml
```diff
@@ -7,3 +7,68 @@ members = [
 
 [profile.release]
 panic = "unwind"
+
+[patch."https://github.com/paritytech/substrate"]
```
ghzlatarev, line 11: Wait why do we need this patch ? And this new Cargo.toml file ?

Dengjianping: The only reason I have to do so is that we included a pallet `pallet-collator-selection`, but whose deps are from substrate's master branch.
But now, we have to bump all deps to `polkadot-v0.9.5`, it will cause compilation problem.
For example, `manta-xassets is using `sp-runtime` that is from `polkadot-v0.9.5`,
but if `pallet-collator-selection` uses sp-runtime` that is from `master`, now there're two version of `sp-runtime` in Cargo.lock. Then compilation will complain.

ghzlatarev: Thanks. Maybe add this as a comment.

ghzlatarev (approved):

stechu:
@Dengjianping  do we still need to merge this?

PR #108, opened 2021-06-29
[Manta-PC]Update impl_name() and description()

ghzlatarev:
* Update impl_name() and description() methods with new name "Manta Parachain Collator"


Dengjianping (approved):
LGTM!

PR #125, opened 2021-07-14
[Calamari-PC]Update para id and total issuance

Dengjianping:
This is a minor change.

1. Update para  id as 2084.
2. Update the total issuance of KMA from 1B to 10B.

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -97,7 +97,7 @@ pub fn manta_pc_development_config(id: ParaId) -> MantaPCChainSpec {
 
 	MantaPCChainSpec::from_genesis(
 		// Name
-		"Manta-Parachain Development",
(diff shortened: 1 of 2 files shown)


node/src/chain_spec.rs
```diff
@@ -563,7 +563,7 @@ fn calamari_testnet_genesis(
 		.cloned()
 		.map(|x| (x.0, ENDOWMENT))
 		.collect();
-	initial_balances.push((root_key.clone(), 500_000_000 * MA));
+	initial_balances.push((root_key.clone(), 5000_000_000 * MA));
```
stechu, line 566: should be KMA here?

stechu: May want to check the token symbol in other places as well.

stechu (changes requested):

PR #128, opened 2021-07-17
[Manta] update types json

stechu:
Update types json so that the Polkadot.js frontend works with the new runtime version.


Dengjianping (approved):
LGTM!

PR #135, opened 2021-07-22
[Calamari-PC]Adjust constants in calamari, and remove pallet-assets(WIP)

### LINKED ISSUE #134: Manta<>Shawn
1. setting constants:
a. AVERAGE_ON_INITIALIZE_RATIO

```rust
/// We assume that ~5% of the block weight is consumed by `on_initialize` handlers. This is
/// used to limit the maximal weight of a single extrinsic.
pub const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(5);
/// We allow `Normal` extrinsics to fill up the block up to 75%, the rest can be used by
/// Operational  extrinsics.
pub const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(75);

/// We allow for 0.5 seconds of compute with a 6 second average block time.
pub const MAXIMUM_BLOCK_WEIGHT: Weight = WEIGHT_PER_SECOND / 2;

// 5MB
pub RuntimeBlockLength: BlockLength =
		BlockLength::max_with_normal_ratio(5 * 1024 * 1024, NORMAL_DISPATCH_RATIO);
```

for example, Acala use a different value:
```rust
// TODO: somehow estimate this value. Start from a conservative value.
pub const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(10);
/// The ratio that `Normal` extrinsics should occupy. Start from a conservative value.
const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(70);
```

b.  `MaxInvulnerables` setting
```rust
parameter_types! {
	// Pallet account for record rewards and give rewards to collator.
	pub const PotId: PalletId = PalletId(*b"PotStake");
	// How many collator candidates is allowed.
	pub const MaxCandidates: u32 = 10;
	pub const MinCandidates: u32 = 3;
	// How many collators who cannot be slashed.
	pub const MaxInvulnerables: u32 = 10;
}
```

c. We make period shorter (from Acala's experience)
```rust
parameter_types! {
	pub const DisabledValidatorsThreshold: Perbill = Perbill::from_percent(33);
	// Rotate collator's spot each 3 hours.
	pub const Period: u32 = 3 * HOURS;
	pub const Offset: u32 = 0;
}
```

d. pallet_transaction_payment
```rust
parameter_types! {
	/// Relay Chain `TransactionByteFee` / 10
	pub const TransactionByteFee: Balance = mMA; // 10^12 / 1_000
}
```

Dengjianping:
Closes #134 

1. Adjust constants defined on calamari runtime, but wait for Shawn's comments.
2. Remove `pallet-asset`, as shell chain, this pallet is unnecessary for now. We can add it later by runtime upgrade.
> BTW, it's really heavy, the size of compact wasm blob is 1.2mb after remove pallet-assets, but previously it was 1.4mb.

### DIFF
--- a/runtime/calamari/src/lib.rs
+++ b/runtime/calamari/src/lib.rs
@@ -424,8 +424,8 @@ impl cumulus_pallet_dmp_queue::Config for Runtime {
 
 parameter_types! {
 	pub const DisabledValidatorsThreshold: Perbill = Perbill::from_percent(33);
-	// Rotate collator's spot each 3 hours.
-	pub const Period: u32 = 3 * HOURS;
+	// Rotate collator's spot each 6 hours.
+	pub const Period: u32 = 6 * HOURS;
 	pub const Offset: u32 = 0;
 }
 
@@ -456,7 +456,7 @@ parameter_types! {
 	pub const MaxCandidates: u32 = 50;
 	pub const MinCandidates: u32 = 3;
 	// How many collators who cannot be slashed.
-	pub const MaxInvulnerables: u32 = 10;
+	pub const MaxInvulnerables: u32 = 5;
 }
 
 /// We allow root and the Relay Chain council to execute privileged collator selection operations.
--- a/runtime/calamari/src/lib.rs
+++ b/runtime/calamari/src/lib.rs
@@ -99,10 +99,10 @@ pub fn native_version() -> NativeVersion {
 
 /// We assume that ~5% of the block weight is consumed by `on_initialize` handlers. This is
 /// used to limit the maximal weight of a single extrinsic.
-pub const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(5);
+pub const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(10);
 /// We allow `Normal` extrinsics to fill up the block up to 75%, the rest can be used by
 /// Operational  extrinsics.
-pub const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(75);
+pub const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(70);
 
 /// We allow for 0.5 seconds of compute with a 6 second average block time.
 pub const MAXIMUM_BLOCK_WEIGHT: Weight = WEIGHT_PER_SECOND / 2;
@@ -145,7 +145,6 @@ impl Filter<Call> for BaseFilter {
 			// Sudo also cannot be filtered because it is used in runtime upgrade.
 			_ => false,
 			// Filter System to prevent users from runtime upgrade without sudo privilege.
-			// Filter Assets and Balances to prevent users from creating assets or making transfers.
 			// Filter Utility and Multisig to prevent users from setting keys and selecting collator for parachain (couldn't use now).
 			// Filter Session and CollatorSelection to prevent users from utility operation.
 			// Filter XCM pallet.
@@ -204,7 +203,7 @@ impl pallet_authorship::Config for Runtime {
 }
 
 parameter_types! {
-	pub const NativeTokenExistentialDeposit: u128 = MA;
+	pub const NativeTokenExistentialDeposit: u128 = 10 * cMA; // 0.1 KMA
 	pub const MaxLocks: u32 = 50;
 	pub const MaxReserves: u32 = 50;
 }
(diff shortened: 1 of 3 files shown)


runtime/calamari/src/lib.rs
```diff
@@ -99,10 +99,10 @@ pub fn native_version() -> NativeVersion {
 
 /// We assume that ~5% of the block weight is consumed by `on_initialize` handlers. This is
 /// used to limit the maximal weight of a single extrinsic.
-pub const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(5);
+pub const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(10);
 /// We allow `Normal` extrinsics to fill up the block up to 75%, the rest can be used by
 /// Operational  extrinsics.
-pub const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(75);
+pub const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(70);
```
ghzlatarev, line 105: Can you write down why these values in the description ?

runtime/calamari/src/lib.rs
```diff
@@ -204,7 +203,7 @@ impl pallet_authorship::Config for Runtime {
 }
 
 parameter_types! {
-	pub const NativeTokenExistentialDeposit: u128 = MA;
+	pub const NativeTokenExistentialDeposit: u128 = 10 * cMA; // 0.1 KMA
```
ghzlatarev, line 206: Same comment as the previous.

runtime/calamari/src/lib.rs
```diff
@@ -484,7 +453,7 @@ parameter_types! {
 	// Pallet account for record rewards and give rewards to collator.
 	pub const PotId: PalletId = PalletId(*b"PotStake");
 	// How many collator candidates is allowed.
-	pub const MaxCandidates: u32 = 10;
+	pub const MaxCandidates: u32 = 50;
```
ghzlatarev, line 456: Again why this value.

runtime/calamari/src/lib.rs
Dengjianping: We followed acala's configuration, Shumo and I discussed about it, 5% is too small.

runtime/calamari/src/lib.rs
Dengjianping: 1MA to create a new account is really not reasonable.

runtime/calamari/src/lib.rs
Dengjianping: We followed acala's configuration as well, it should be not too small or too big.
Definitely we can modify these constants by runtime upgrade.

ghzlatarev: Great thanks. Just wanted to include the reasoning in the PR, if someone reads through it in the future.

ghzlatarev (approved):

runtime/calamari/src/lib.rs
```diff
@@ -99,10 +99,10 @@ pub fn native_version() -> NativeVersion {
 
 /// We assume that ~5% of the block weight is consumed by `on_initialize` handlers. This is
 /// used to limit the maximal weight of a single extrinsic.
-pub const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(5);
+pub const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(10);
 /// We allow `Normal` extrinsics to fill up the block up to 75%, the rest can be used by
 /// Operational  extrinsics.
```
ghzlatarev, line 104: Can we add a note about Operational extrinsics, that they have higher priority and can make a block larger than it should be, so they can fit and recover situations.
Also that decreasing the value is more conservative, should be self evident but still.

runtime/calamari/src/lib.rs
```diff
@@ -99,10 +99,10 @@ pub fn native_version() -> NativeVersion {
 
 /// We assume that ~5% of the block weight is consumed by `on_initialize` handlers. This is
 /// used to limit the maximal weight of a single extrinsic.
-pub const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(5);
+pub const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(10);
```
ghzlatarev, line 102: Can we add a note that increasing this is more conservative. Should be self evident but still.

PR #141, opened 2021-07-28
[Manta] Adapt to new pallet-manta-pay interface

stechu:
This PR adapt `pallet-manta-pay` version.

### DIFF
--- a/runtimes/manta/runtime/Cargo.toml
+++ b/runtimes/manta/runtime/Cargo.toml
@@ -3,7 +3,7 @@ authors = ['Manta Network']
 edition = '2018'
 license = 'GPL V3'
 name = 'manta-runtime'
-version = '3.0.0'
+version = '3.0.1'
 
 [package.metadata.docs.rs]
 targets = ['x86_64-unknown-linux-gnu']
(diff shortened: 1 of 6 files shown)


runtimes/manta/runtime/Cargo.toml
```diff
@@ -3,7 +3,7 @@ authors = ['Manta Network']
 edition = '2018'
 license = 'GPL V3'
 name = 'manta-runtime'
-version = '3.0.0'
+version = '3.0.1'
```
Dengjianping, line 6: Do we only bump runtime version?

runtimes/manta/runtime/src/lib.rs
```diff
@@ -107,8 +106,8 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	//   `spec_version`, and `authoring_version` are the same between Wasm and native.
 	// This value is set to 100 to notify Polkadot-JS App (https://polkadot.js.org/apps) to use
 	//   the compatible custom types.
-	spec_version: 2,
-	impl_version: 1,
+	spec_version: 3,
+	impl_version: 2,
```
Dengjianping: In my opinion, most of time, we don't need to bump `impl_version`.
Reference: https://substrate.dev/docs/en/knowledgebase/runtime/upgrades

runtimes/manta/runtime/Cargo.toml
stechu: yeah, we should bump other version as well. good call.

runtimes/manta/runtime/src/lib.rs
stechu: agreed. will revert it back.

Dengjianping (approved):
LGTM!

PR #149, opened 2021-08-11
[Manta-PC]Refactor MantaTransactorAdaptor

Dengjianping:
1. Refactor `MantaTransactorAdaptor`.
2. Add crate 'manta-xcm-support' that implements xcm related traits and data types.
3. Add a clippy to disable the usage of `unwrap` while do clippy check.
4. Define xcm events to `manta-xassets`.

### DIFF
--- a/runtime/primitives/src/currency_id.rs
+++ b/runtime/primitives/src/currency_id.rs
@@ -14,6 +14,10 @@ pub enum TokenSymbol {
 	KAR = 128,
 	// Shiden
 	SDN = 50,
+
+	// Relaychain Token
+	KSM,
+	DOT,
 }
 
 impl Default for TokenSymbol {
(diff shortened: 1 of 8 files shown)


runtime/primitives/src/currency_id.rs
```diff
@@ -33,3 +37,29 @@ impl Default for CurrencyId {
 		Self::Token(TokenSymbol::MA)
 	}
 }
+
+#[allow(dead_code)]
```
ghzlatarev, line 41: Why do we need ``#[allow(dead_code)]``

runtime/primitives/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 #![allow(non_upper_case_globals)]
 #![allow(clippy::unnecessary_cast)]
 #![allow(clippy::upper_case_acronyms)]
+#![forbid(clippy::unwrap_used)]
```
ghzlatarev, line 4: Coool!

runtime/primitives/src/currency_id.rs
Dengjianping: Compiler always reports we don't use it in anywhere.

stechu:
LGTM. I think we should make the `manta-pc-xcm` as the official branch and make PR to `manta-pc-xcm`?

Dengjianping:
> LGTM. I think we should make the `manta-pc-xcm` as the official branch and make PR to `manta-pc-xcm`?

Yes, we should create another pr to merge branch `manta-acala` into `manta-pc-xcm`.

PR #150, opened 2021-08-12
XCM Prototype

stechu:


ghzlatarev:
@stechu maybe we should make this a protected branch ?

Dengjianping:
> @stechu maybe we should make this a protected branch ?

We're going to take `manta-pc-xcm` as a branch for xcm related, but not production-ready.

Dengjianping:
LGTM!

PR #153, opened 2021-08-16
[Manta-PC]Fix unexpected block time

### LINKED ISSUE #147: Investigate why manta parachain produces block so slowly, but acala is more fast than us.
This issue would be good case to test integration test from @grenade .

Now, in `manta-acala` XCM experiment, manta parachain (using `manta-pc-dev` config)'s block time is significantly slower than Acala.

Dengjianping:
Closes #147 

Follow statemint' client codes from cumulus branch v0.9.8.

### DIFF
--- a/runtime/manta-pc/src/lib.rs
+++ b/runtime/manta-pc/src/lib.rs
@@ -36,8 +36,8 @@ use frame_system::{
 	EnsureOneOf, EnsureRoot,
 };
 use manta_primitives::{
-	currency::*, fee::WeightToFee, time::*, AccountId, AuraId, Balance, BlockNumber, Hash, Header,
-	Index, Signature,
+	currency::*, currency_id::CurrencyId, fee::WeightToFee, time::*, AccountId, AuraId, Balance,
+	BlockNumber, Hash, Header, Index, Signature,
 };
 use sp_runtime::Perbill;
 
@@ -487,7 +487,7 @@ pub type XcmOriginToTransactDispatchOrigin = (
 
 parameter_types! {
 	// One XCM operation is 1_000_000 weight - almost certainly a conservative estimate.
-	pub UnitWeightCost: Weight = 1_000_000_000;
+	pub UnitWeightCost: Weight = 200_000_000;
 }
 
 match_type! {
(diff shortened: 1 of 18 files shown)


runtime/manta-pc/src/lib.rs
```diff
@@ -500,23 +508,29 @@ match_type! {
...
+	>;
 	type OriginConverter = XcmOriginToTransactDispatchOrigin;
-	type IsReserve = NativeAsset;
+	type IsReserve = manta_xcm_support::TrustedParachains<TrustedChains>;
 	type IsTeleporter = NativeAsset; // <- should be enough to allow teleportation of DOT
 	type LocationInverter = LocationInverter<Ancestry>;
 	type Barrier = Barrier;
 	type Weigher = FixedWeightBounds<UnitWeightCost, Call>;
-	type Trader = UsingComponents<IdentityFee<Balance>, DotLocation, AccountId, Balances, ()>;
+	type Trader = UsingComponents<IdentityFee<Balance>, AcalaLocation, AccountId, Balances, ()>;
```
stechu, line 533: Can we make this location configurable?

stechu:
Thanks for the PR! LGTM in general. Can we use this to integrate with @grenade new build automation?

stechu (changes requested):

runtime/manta-pc/src/lib.rs
Dengjianping: It seems it has to be hardcoded. Look at the definition of `UsingComponents`:
```rust
pub struct UsingComponents<
	WeightToFee: WeightToFeePolynomial<Balance = Currency::Balance>,
	AssetId: Get<MultiLocation>,
	AccountId,
	Currency: CurrencyT<AccountId>,
	OnUnbalanced: OnUnbalancedT<Currency::NegativeImbalance>,
>(
	Weight,
	Currency::Balance,
	PhantomData<(WeightToFee, AssetId, AccountId, Currency, OnUnbalanced)>,
);
```
`AssetId` only supports single value instead of `Vec<MultiLocation>` or something else.

We might post this issue to polkadot.

Dengjianping: I just went through related xmc code from polkadot, I think we need to implement our own 'UsingComponents', just impl this trait `WeightTrader `.
I'll create an issue for this.

stechu:
@grenade @Dengjianping Any update on the runner integration?

Dengjianping:
> @grenade @Dengjianping Any update on the runner integration?

For my part, I'm still writing that tool which can measure block time.

PR #155, opened 2021-08-18
[Manta-PC] benchmark persistence workflow

grenade:
this workflow:
- uses self-hosted aws runners to produce a benchmark build and observe/record benchmark timings for that build
- adds benchmark persistence to a historical benchmarks table which will allow us to graph benchmark timings over time and across commits.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -211,6 +211,7 @@ jobs:
                 observed: ISODate(),
                 pallet: "${{ matrix.benchmark.pallet.name }}",
                 extrinsic: "${{ matrix.benchmark.extrinsic.name }}",
+                machine: "c5a.8xlarge",
                 actual: {
                   time: '${actual_time}',
                   unit: "'${actual_unit}'"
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,288 @@
...
+              id: pallet_manta_pay
+              name: pallet-manta-pay
+            iterations: 1000
+            expected:
+              time: 180000
+              unit: µs
+    steps:
+      -
+        name: install mongodb-mongosh
+        run: |
```
stechu, line 156: mind elaborating why we need mongodb here?

stechu (approved):

.github/workflows/benchmark.yml
grenade: not at all. each time we run benchmarks we [push a benchmark stats record](https://github.com/Manta-Network/Manta/blob/44a2ff9/.github/workflows/benchmark.yml#L204-L224) to a mongo db. this enables the historical benchmark stats to be available for analysis over time or across changes. eg: https://dev.stats.manta.systems

PR #166, opened 2021-08-25
[Calamari]Add more collators for calamari-local chain spec and update polkadot deps

### LINKED ISSUE #165: Let calamari-local chain spec support 5 collators

Dengjianping:
Closes #165 

1. Add more collators for calamari-local chain spec.
2. Update polkadot deps to the latest commit of branch `release-v0.9.9`.
> Tips: We're using the commit `67d539dd74768096cbe024dce2577560729b2341`, but the latest commit is `7a9f624777ad2d2adb3a1e6797a31f9d653c9587`.
And no need to update cumulus and substrate deps, they didn't update both repositories actually.

### DIFF
--- a/.github/workflows/publish-draft-releases.yml
+++ b/.github/workflows/publish-draft-releases.yml
@@ -604,7 +604,7 @@ jobs:
         name: run test suites
         run: |
           # todo: implement moonbeam-like js test suite triggers here
-          sleep 180
+          sleep 210
       -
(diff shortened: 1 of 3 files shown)


stechu:
I checked the error log, it appears that the test fall short due to the block producing is a little bit behind the expectation. The integration test expect greping `#6` block from the parachain at the time when the parachain is producing `#5` block.

PR #168, opened 2021-08-27
[Calamari] Calamari pre release

stechu:
Calamari Pre-Release.


Cargo.lock
```diff
@@ -720,7 +720,7 @@ dependencies = [
 [[package]]
 name = "bp-header-chain"
 version = "0.1.0"
-source = "git+https://github.com/paritytech/polkadot.git?branch=release-v0.9.9#7a9f624777ad2d2adb3a1e6797a31f9d653c9587"
```
Dengjianping: It seems you rollbacked the polkadot deps.
`67d539dd74768096cbe024dce2577560729b2341` is older than `7a9f624777ad2d2adb3a1e6797a31f9d653c9587`.
See this pr: https://github.com/Manta-Network/Manta/pull/166

Dengjianping (changes requested):

node/Cargo.toml
```diff
@@ -104,7 +104,7 @@ manta-primitives = { path = '../runtime/primitives' }
 substrate-build-script-utils = { git = 'https://github.com/paritytech/substrate.git', branch = "polkadot-v0.9.9" }
 
 [features]
-default = []
+default = ['calamari']
```
Dengjianping: I remembered why I didn't set any default feature here, the reason is that if we compile `calamari`, that's fine. But if we compile `manta-pc`, it will generate two binaries `calamari-pc` and `manta-pc`, because both features are enabled at the same time.

PR #172, opened 2021-08-30
Integrate runtime upgrade test into CI

ghzlatarev:
closes https://github.com/Manta-Network/pallet-manta-pay/issues/92

* Integrates a runtime upgrade test into the CI
* First the spec version from the current commit is compared to the spec verion in the base branch.
* If they don't match the runtime upgrade job is triggered.
* The code from the base branch is used to launch the tesntet
* The code from the current commit is used to build the wasm to upgrade the runtime
* The log for the runtime upgrade is parsed for success and also block production is checked.
* The log is also uploaded as artifact.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,25 @@
+<!-- < < < < < < < < < < < < < < < < < < < < < < < < < < < < < < < < < ☺
+v                               ✰  Thanks for creating a PR! ✰
+v    Before hitting that submit button please review the checkboxes.
+v    If a checkbox is n/a - please still include it but + a little note why
+☺ > > > > > > > > > > > > > > > > > > > > > > > > > > > > > > > > >  -->
+
+## Description
+
+<!-- Add a description of the changes that this PR introduces and the files that
+are the most critical to review.
+-->
+
+closes: #XXXX
+
+---
+
+Before we can merge this PR, please make sure that all the following items have been
+checked off. If any of the checklist items are not applicable, please leave them but
+write a little note why.
+
+- [ ] Targeted PR against correct branch (master)
+- [ ] Linked to Github issue with discussion and accepted design OR have an explanation in the PR that describes this work.
+- [ ] Wrote unit tests
+- [ ] Updated relevant documentation in the code
+- [ ] Re-reviewed `Files changed` in the Github PR explorer
--- a/.github/workflows/publish-draft-releases.yml
+++ b/.github/workflows/publish-draft-releases.yml
@@ -446,7 +446,7 @@ jobs:
       matrix:
         chain-spec:
           -
-            id: calamari-testnet
+            id: calamari-testnet-ci
             expected:
               block-count:
                 para: 6
--- a/.github/workflows/publish-draft-releases.yml
+++ b/.github/workflows/publish-draft-releases.yml
@@ -6,11 +6,11 @@ name: publish draft releases
 
 # yamllint disable-line rule:truthy
 on:
+  pull_request:
+      - manta-pc
   push:
     tags:
       - 'v[0-9]+.[0-9]+.[0-9]+*'
-    pull_request:
-      - manta-pc
 
 jobs:
 
--- a/.github/workflows/publish-draft-releases.yml
+++ b/.github/workflows/publish-draft-releases.yml
@@ -7,10 +7,10 @@ name: publish draft releases
 # yamllint disable-line rule:truthy
 on:
   push:
-    branch:
-      - manta-pc
     tags:
       - 'v[0-9]+.[0-9]+.[0-9]+*'
+    pull_request:
+      - manta-pc
 
 jobs:
 
--- a/.github/workflows/publish-draft-releases.yml
+++ b/.github/workflows/publish-draft-releases.yml
@@ -9,8 +9,8 @@ on:
   push:
     tags:
       - 'v[0-9]+.[0-9]+.[0-9]+*'
-    # branch:
-    #   - manta-pc
+    branch:
+      - ghzlatarev/rtu-calamari-test-net
 
 jobs:
 
@@ -564,7 +564,6 @@ jobs:
                     "flags": [
                       "--rpc-cors=all",
                       "--rpc-port=9971",
-                      "--rpc-methods=unsafe",
                       "--",
                       "--execution=wasm"
                     ]
@@ -575,7 +574,6 @@ jobs:
                     "flags": [
                       "--rpc-cors=all",
                       "--rpc-port=9972",
-                      "--rpc-methods=unsafe",
                       "--",
                       "--execution=wasm"
                     ]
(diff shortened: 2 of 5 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,25 @@
+<!-- < < < < < < < < < < < < < < < < < < < < < < < < < < < < < < < < < ☺
```
ghzlatarev, line 1: Do we want something like this ?

.github/workflows/publish-draft-releases.yml
```diff
@@ -7,10 +7,10 @@ name: publish draft releases
...
 on:
   push:
-    tags:
-      - 'v[0-9]+.[0-9]+.[0-9]+*'
-    # branch:
-    #   - manta-pc
+    branch:
+      - manta-pc
+    #tags:
+    #  - v[0-9]+.[0-9]+.[0-9]+*
```
ghzlatarev: Revert this when done.

@grenade  how do I trigger the CI with the tags ?

.github/workflows/publish-draft-releases.yml
```diff
@@ -340,11 +340,405 @@ jobs:
...
+  build-node-old:
+    needs:
+      - start-node-builder-old
+      - check-for-runtime-upgrade
+    runs-on: ${{ needs.start-node-builder-old.outputs.runner-label }}
+    env:
+      CARGO_TERM_COLOR: always
+    steps:
+      -
+        if: ${{ needs.check-for-runtime-upgrade.outputs.runtime-version != needs.check-for-runtime-upgrade.outputs.runtime-version-old }}
```
ghzlatarev: These conditionals need to be on each step, unfortunately.
There's no other way to exit early with success.
We need jobs to rely on the runtime upgrade stuff only if the runtime upgrade actually happens.

.github/workflows/publish-draft-releases.yml
```diff
@@ -340,11 +340,405 @@ jobs:
...
+          curl -vH 'Content-Type: application/json' --data '{ "jsonrpc":"2.0", "method":"author_insertKey", "params":["aura", "civil cigar remain hybrid glove symptom review what pole lock concert lamp", "0xb40aa6bd104d0260b60350c2fb30d4882437466d66135130b667799ea6c9f52b"],"id":1 }' localhost:9973
+          curl -vH 'Content-Type: application/json' --data '{ "jsonrpc":"2.0", "method":"author_insertKey", "params":["aura", "output evidence anger invest country opinion girl mouse direct double carbon usage", "0x4a3aa51469e802be6504422cd9dd03be638ac3f6dc3a7c0c85a6ace3e72f0048"],"id":1 }' localhost:9974
+          curl -vH 'Content-Type: application/json' --data '{ "jsonrpc":"2.0", "method":"author_insertKey", "params":["aura", "series disorder today argue interest pond flight guess asthma guilt road gadget", "0xa68feb4fe2ea3f8ff288af4254aad2284e1cd0da67cb9ea61c13632bad57eb40"],"id":1 }' localhost:9975
```
ghzlatarev, line 689: Generated my own and updated the ones in the chain spec.
Wasn't sure whether to use the originals and Deep-Manta.

.github/workflows/publish-draft-releases.yml
```diff
@@ -921,10 +1313,10 @@ jobs:
...
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          aws-instance-ssh-public-key: ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIPb24HEQ++aNFWaqVyMtIs6GotUB8R+q61XOoI2z6uMj
-          aws-region: us-east-1
-          aws-subnet-id: subnet-08c26caf0a52b7c19
-          aws-security-group-id: sg-0315bffea9042ac9b
+          # aws-instance-ssh-public-key: ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIPb24HEQ++aNFWaqVyMtIs6GotUB8R+q61XOoI2z6uMj
+          aws-region: eu-west-1
+          # aws-subnet-id: subnet-08c26caf0a52b7c19
+          # aws-security-group-id: sg-0315bffea9042ac9b
```
ghzlatarev: This is a workaround for an AWS error ``MaxSpotInstanceCountExceeded: Max spot instance count exceeded``

ghzlatarev: TODO: revert before merging.

runtime/calamari/src/lib.rs
```diff
@@ -82,7 +82,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("calamari"),
 	impl_name: create_runtime_str!("calamari"),
 	authoring_version: 1,
-	spec_version: 1,
+	spec_version: 2,
```
ghzlatarev, line 85: To trigger the runtime upgrade job.

TODO: revert before merging.

.github/workflows/publish-draft-releases.yml
```diff
@@ -340,11 +340,405 @@ jobs:
...
+      - check-for-runtime-upgrade
+    runs-on: ${{ needs.start-node-builder-old.outputs.runner-label }}
+    env:
+      CARGO_TERM_COLOR: always
+    steps:
+      -
+        if: ${{ needs.check-for-runtime-upgrade.outputs.runtime-version != needs.check-for-runtime-upgrade.outputs.runtime-version-old }}
+        uses: actions/checkout@v2
+        with:
+          ref: '4204135e90ce02fa28af0913c8f050b76bd3ea23'
```
ghzlatarev: This should be ``manta-pc``.
Using this commit for testing as it has my modified chain spec.

TODO: revert before merging.

.github/PULL_REQUEST_TEMPLATE.md
ghzlatarev: Actually this file needs to be added in the default branch of the repo.

@stechu @grenade

node/src/chain_spec.rs
```diff
@@ -589,6 +589,63 @@ pub fn calamari_testnet_config(id: ParaId) -> CalamariChainSpec {
...
+		),
+		(
+			hex!["a2057091a6d15a9d64c398f148395b69aaae99626cffa53ac5e6beaa5790f946"].into(),
+			hex!["a68feb4fe2ea3f8ff288af4254aad2284e1cd0da67cb9ea61c13632bad57eb40"]
+				.unchecked_into(),
+		),
+	];
+
+	let root_key: AccountId =
+		hex!["04ac7fe1675e10c0df9c136f53df912e6ae7d2c21778a293c534226f8a3e6a0c"].into();
```
ghzlatarev, line 626: I split this into its own method because we don't have a solution in place to handle the corresponding secrets yet.
Generated new ones and directly use them for CI.

ghzlatarev:
Going to reoppen this PR on a different branch, because I cannot push to it anymore. It's complaining that there's a tag with the same name when there isn't...

PR #189, opened 2021-09-10
[Manta-PC] Integrate runtime upgrade test into CI

ghzlatarev:
## Description

closes https://github.com/Manta-Network/pallet-manta-pay/issues/92

* Integrates a runtime upgrade test into the CI
* First the spec version from the current commit is compared to the spec verion in the base branch.
* If they don't match the runtime upgrade job is triggered.
* The code from the base branch is used to launch the tesntet.
* The code from the current commit is used to build the wasm to upgrade the runtime.
* The log for the runtime upgrade is parsed for success and also block production is checked.
* The log is also uploaded as artifact.

---

write a little note why.

- [x] Wrote unit tests.


ghzlatarev:
@stechu used the verified commit.

ghzlatarev (commented):
TODO: add benchmarks for the new pallets.

stechu (approved):

PR #221, opened 2021-10-19
Add calamari-vesting pallet

Dengjianping:
## Description

closes: #XXXX

---

write a little note why.

- [x] Wrote unit tests.


Cargo.lock
```diff
@@ -4403,6 +4404,25 @@ dependencies = [
  "sp-std",
 ]
 
+[[package]]
+name = "manta-vesting"
+version = "3.0.6"
```
ghzlatarev: You should merge manta-pc, it has the 3.0.8 release already.

Dengjianping: I merged latest manta-pc, it's 3.0.6. I think you didn't align it in your pr. But it's fine, I will bump version to 3.0.9.

Dengjianping:
Todo
1. Adapt new schedule.
2. Generate new weight files.

runtime/calamari/src/lib.rs
```diff
@@ -91,7 +91,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("calamari"),
 	impl_name: create_runtime_str!("calamari"),
 	authoring_version: 1,
-	spec_version: 4,
+	spec_version: 5,
```
stechu: please use the new semantic versioning convention of the spec. I think it should be 3100

runtime/manta-pc/Cargo.toml
```diff
@@ -5,7 +5,7 @@ homepage = 'https://manta.network'
 license = 'GPL-3.0'
 name = 'manta-pc-runtime'
 repository = 'https://github.com/Manta-Network/Manta/'
-version = '3.0.8'
+version = '3.0.9'
```
stechu: `3.1.0`

runtime/primitives/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 authors = ['Manta Network']
 name = "manta-primitives"
-version = "3.0.8"
+version = "3.0.9"
```
stechu: same here.

pallets/vesting/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
+authors    = ['Manta Network']
+name       = "manta-vesting"
```
stechu: Can you change the folder and package name to `calamari-vesting`? Manta's vesting schedule and policy will be different.

pallets/vesting/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+		BoundedVec<(Percent, Schedule), T::MaxScheduleLength>,
+		ValueQuery,
+		DefaultVestingSchedule<T>,
+	>;
+
+	#[pallet::type_value]
+	pub(super) fn DefaultVestingSchedule<T: Config>(
+	) -> BoundedVec<(Percent, Schedule), T::MaxScheduleLength> {
+		BoundedVec::try_from(sp_std::vec![
+			(Percent::from_percent(34), 1636329600),
```
stechu: Can you comment the date and time of these timestamps?

pallets/vesting/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+		/// The first round of vesting is not done yet.
+		ClaimTooEarly,
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Update vesting schedule.
+		///
+		/// - `new_schedule`: New schedule for vesting.
+		#[pallet::weight(10_000)]
```
stechu: weight need to be updated.

pallets/vesting/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+			ensure!(
+				new_schedule.as_slice().windows(2).all(|w| w[0] < w[1]),
+				Error::<T>::UnsortedSchedule
+			);
+
+			let now = T::Timestamp::now().as_secs();
+			ensure!(
+				new_schedule.iter().all(|&s| now <= s),
+				Error::<T>::InvalidTimestamp
+			);
```
stechu: The check is incorrect here, say if we want to update schedule after the first vesting happens. This check will fail.

We should check the following:
1. In a new schedule, if that is a past time, it should be consistent with the old_schedule
2. if that is a future time, it is fine. We also need to put some `buffer` of future time considering the block finality.

stechu: Also, please update the logic of this check to the rustdoc comments.

pallets/vesting/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+				{
+					schedule.1 = *newer_schedule;
+				}
+			});
+
+			Self::deposit_event(Event::VestingScheduleUpdated(new_schedule));
+			Ok(())
+		}
+
+		/// Unlock any vested funds of the sender account.
```
stechu:
```suggestion
		/// Unlock vested balance according to the schedule
```

pallets/vesting/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+		///
+		/// The dispatch origin for this call must be _Signed_ and the sender must have funds still
+		/// locked under this pallet.
+		///
+		/// Emits either `VestingCompleted` or `VestingUpdated`.
+		#[pallet::weight(10_000)]
```
stechu: update weight.

pallets/vesting/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+			// Ensure signer can claim once time is up to schedule.
+			ensure!(
+				Some(now) >= VestingSchedule::<T>::get().first().map(|v| v.1),
+				Error::<T>::ClaimTooEarly
+			);
+
+			Self::update_lock(&who)
+		}
+
+		/// Create a vested transfer.
```
stechu:
```suggestion
		/// Create a vested transfer: send `target` balance with the vesting schedule.
```

pallets/vesting/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+			Self::deposit_event(Event::<T>::VestingCompleted(who.clone()));
+		} else {
+			let reasons = WithdrawReasons::TRANSFER | WithdrawReasons::RESERVE;
+			T::Currency::set_lock(VESTING_ID, who, unvested, reasons);
+			Self::deposit_event(Event::<T>::VestingUpdated(who.clone(), unvested));
+		}
+		Ok(())
+	}
+
+	fn add_vesting_schedule(who: &T::AccountId, locked: BalanceOf<T>) -> DispatchResult {
```
stechu:
```suggestion
	fn new_vesting_account(who: &T::AccountId, locked: BalanceOf<T>) -> DispatchResult {
```

stechu: function comment.

pallets/vesting/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+	#[pallet::getter(fn vesting_balance)]
+	pub(super) type VestingBalances<T: Config> =
+		StorageMap<_, Blake2_128Concat, T::AccountId, BalanceOf<T>>;
+
+	/// Information regarding the vesting of a given account.
+	#[pallet::storage]
+	#[pallet::getter(fn vesting_schedule)]
+	pub(super) type VestingSchedule<T: Config> = StorageValue<
+		_,
+		BoundedVec<(Percent, Schedule), T::MaxScheduleLength>,
```
stechu, line 87: Why not just use `T:Timestamp` here? (Instead of `Schedule`, which is `u64`)

pallets/vesting/src/tests.rs
```diff
@@ -0,0 +1,415 @@
...
+		.existential_deposit(1)
+		.build()
+		.execute_with(|| {
+			let unvested = 100;
+			assert_ok!(MantaVesting::vested_transfer(
+				Origin::signed(ALICE),
+				BOB,
+				unvested
+			));
+			assert_eq!(Balances::free_balance(ALICE), 10_000 - unvested);
```
stechu: Can you replace `10_000` with a constant rather a magic number?

pallets/vesting/src/tests.rs
```diff
@@ -0,0 +1,415 @@
...
+			run_to_block(3);
+			let now = VestingSchedule::<Test>::get()[0].1 * 1000 + 1;
+			Timestamp::set_timestamp(now);
+
+			assert_ok!(MantaVesting::vest(Origin::signed(BOB)));
+			assert_eq!(Balances::free_balance(BOB), unvested);
+
+			// BOB cannot transfer more than 34 tokens.
+			// Bacause rest of 66 is locked now.
+			let vested = 34;
```
stechu: Can you calculate this in code? Since the final vesting schedule may change.

pallets/vesting/src/tests.rs
```diff
@@ -0,0 +1,415 @@
...
+				Balances::transfer(Origin::signed(BOB), ALICE, vested + 1),
+				pallet_balances::Error::<Test, _>::LiquidityRestrictions,
+			);
+
+			assert_ok!(Balances::transfer(Origin::signed(BOB), ALICE, vested));
+			assert_eq!(Balances::free_balance(ALICE), 10_000 - unvested + vested);
+			assert_eq!(Balances::free_balance(BOB), unvested - vested);
+
+			// Ensure Bob can claim all tokens once vesting is done.
+			let now = VestingSchedule::<Test>::get()[6].1 * 1000 + 1;
```
stechu: get rid of magic number here.

pallets/vesting/src/tests.rs
```diff
@@ -0,0 +1,415 @@
...
+			// Check vested done event
+			System::assert_has_event(MockEvent::MantaVesting(PalletEvent::VestingCompleted(BOB)));
+
+			// Now, Bob can transfer all his tokens.
+			assert_ok!(Balances::transfer(
+				Origin::signed(BOB),
+				ALICE,
+				unvested - vested
+			));
+			assert_eq!(Balances::free_balance(ALICE), 10_000);
```
stechu: get rid of the magic number here.

pallets/vesting/src/tests.rs
```diff
@@ -0,0 +1,415 @@
...
+			assert_eq!(Balances::free_balance(BOB), 0);
+
+			// Ensure vesting info is removed once vesting is done.
+			assert_eq!(VestingBalances::<Test>::get(BOB), None);
+		});
+}
+
+#[test]
+fn update_vesting_schedule_should_work() {
```
stechu: Can you split this into multiple unit tests?

stechu (changes requested):
Thanks for the PR! See the detailed comments.

stechu:
Also, please fix the conflicts.

pallets/vesting/src/lib.rs
Dengjianping: It could not be compiled with this definition:
```rust
type Timestamp: Parameter + UnixTime + From<u64> + Into<u64>;
```
Because `pallet-timestamp` doesn't implement `From/Into` trait. Lots of compilation errors will be prompted.

pallets/vesting/src/lib.rs
Dengjianping: I will update related weight once review is done. It doesn't make sense to update weight while this pr is under review.

pallets/vesting/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+				// n > o means new schedule is future schedule.
+				// n < o && n > now, also fine.
+				if *n == o.1 {
+					continue;
+				}
+				// This is an invalid schedule. New schedule cannot be past time.
+				if *n <= now {
+					return Err(Error::<T>::InvalidTimestamp.into());
+				}
+			}
```
stechu:
```suggestion
			for (n, o) in new_schedule.iter().zip(old_schedule.iter()) {
				if o.1 <= now {
				     ensure!(
				        *n == o.1,
				         Error::<T>::InvalidTimestamp
				      )
				}  
			}
```

Dengjianping: You have to consider new schedule is future time. If we don't add `else` branch, it still can be possible to update a invalid schedule.
For example:
```
old-schedule = [1,2,3,4,5,6,7];
now = 3.5;
new-schedule = [1,2,3,2,3,2,2];
```
After check `[1,2,3]`, we still need to check the rest of new schedule.

```rust
for (n, o) in new_schedule.iter().zip(old_schedule.iter()) {
    if o.1 <= now {
	    ensure!(*n == o.1, Error::<T>::InvalidSchedule);
    } else {
	    ensure!(*n >= now, Error::<T>::InvalidSchedule);
    }
}
```

Dengjianping: You have to consider new schedule is future time. If we don't add `else` branch, it still can be possible to update an invalid schedule.
For example:
```
old-schedule = [10, 20, 30, 100, 110, 120, 130];
now = 99;
new-schedule = [10, 20, 30, 98, 110, 120, 130];
```
After check `[10, 20, 30]`, `98` is smaller than now(99). So, we still need to check the rest of new schedule.

```rust
for (n, o) in new_schedule.iter().zip(old_schedule.iter()) {
    if o.1 <= now {
        ensure!(*n == o.1, Error::<T>::InvalidSchedule);
    } else {
        ensure!(*n >= now, Error::<T>::InvalidSchedule);
    }
}
```

pallets/vesting/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+
+		let unvested = (Percent::from_percent(100) - portion) * vesting;
+
+		if unvested.is_zero() {
+			T::Currency::remove_lock(VESTING_ID, who);
+			VestingBalances::<T>::remove(&who);
+			Self::deposit_event(Event::<T>::VestingCompleted(who.clone()));
+		} else {
+			let reasons = WithdrawReasons::TRANSFER | WithdrawReasons::RESERVE;
+			T::Currency::set_lock(VESTING_ID, who, unvested, reasons);
```
stechu, line 291: If `unvested == locked value`, we don't need to `set_lock` here.

stechu (changes requested):

stechu (approved):

PR #227, opened 2021-10-21
[Manta-PC] runtime upgrade hotfix

stechu:
## Description

closes: #XXXX

---

write a little note why.

- [ ] Wrote unit tests.

### DIFF
--- a/runtime/calamari/src/lib.rs
+++ b/runtime/calamari/src/lib.rs
@@ -142,7 +142,8 @@ impl Contains<Call> for BaseFilter {
 			| Call::ParachainSystem(_)
 			| Call::Authorship(_)
 			| Call::Sudo(_)
-			| Call::Multisig(_) => true,
+			| Call::Multisig(_) 
+			| Call::Balances(_)=> true,
 			// pallet-timestamp and parachainSystem could not be filtered because they are used in commuication between releychain and parachain.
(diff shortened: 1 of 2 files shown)


stechu:
Note: this PR is not intended to be merged.

runtime/calamari/src/lib.rs
```diff
@@ -525,6 +527,21 @@ construct_runtime!(
...
+impl OnRuntimeUpgrade for CalamariUpgradeHotFix {
+	fn on_runtime_upgrade() -> Weight {
+		let sudo = hex!["9b862dcfb8bc6b7d419e3ce659b3c1704b7dea3ec3ee2745de72e8948330af2d"].into();
+		let sudo_controller_1 = hex!["d607f740bb32a506c53fe81ddeabeab7333943ee4ef4c7adf0a6970a0c728833"].into();
+		<Balances as Currency<_>>::transfer(
+			&sudo,
+			&sudo_controller_1,
+			10_000_000_000_000_000u128,
+			ExistenceRequirement::AllowDeath,
+		).unwrap();
```
Dengjianping, line 540: Can you remove this unwrap? I know it won't fail, but just in case.

Dengjianping (changes requested):

PR #245, opened 2021-11-04
Unified client for Manta and Calamari

### LINKED ISSUE #232: Unified Manta and Calamari Client
To reduce the complexity of DevOp, we should have a unified runtime for Calamari and Manta, similar to Polkadot.

stechu:
## Description

closes: #232 

- [x]  Unify Manta/Calamari Client
- [x]  Change CI accordingly
- [x]  fix CI client version bug
- [x] update docker file generation 

---

write a little note why.

- [ ] Wrote unit tests.


node/src/chain_specs/calamari.rs
```diff
@@ -1,3 +1,19 @@
+// Copyright 2019-2021 Manta Network.
```
Dengjianping: Start year should be 2020?

node/src/command.rs
```diff
@@ -19,71 +37,64 @@ use std::{io::Write, net::SocketAddr};
...
 impl SubstrateCli for Cli {
 	fn impl_name() -> String {
-		cfg_if::cfg_if! {
-			if #[cfg(feature = "manta-pc")] {
-				"Manta Parachain Collator".into()
-			} else {
-				"Calamari Parachain Collator".into()
-			}
-		}
+		"Manta/Calamari Collator".into()
```
Dengjianping, line 96: We cannot change this name at will. It will affect whether polkadot-js can find our logo.
Just take a look at this place: https://github.com/polkadot-js/apps/blob/master/packages/apps-config/src/ui/logos/index.ts#L308

Dengjianping:
Should our manta runtime be renamed from `manta-pc` to `manta`?
https://github.com/Manta-Network/Manta/blob/2a8526df680dc20459866aa62005d15ad1200a79/runtime/manta/src/lib.rs#L104

But if we change this name, I'm not sure whether we need to submit a pr to polkadot-js.
We should take some time on testing with frontend, ensure polkadot-js can work well with our configuration.

BTW, what's the size of new client binary?

Dengjianping:
Please rename that dockerfile, because we're going to unify the binary, there's no need to keep that name of dockerfile.

node/src/command.rs
ghzlatarev: We should add a comment about this.

node/src/command.rs
```diff
@@ -19,71 +37,64 @@ use std::{io::Write, net::SocketAddr};
 
 pub type Block = generic::Block<Header, OpaqueExtrinsic>;
 
+// TODO: add Manta parachain ID if available
```
ghzlatarev, line 40: unlogged/unresolved TODO

node/src/command.rs
stechu: Yeah, I will put a switch here.

stechu:
> Please rename that dockerfile, because we're going to unify the binary, there's no need to keep that name of dockerfile.

I think we still need two docker file? since the genesis are different?

node/src/command.rs
stechu: > Should our manta runtime be renamed from `manta-pc` to `manta`?
> 
> https://github.com/Manta-Network/Manta/blob/2a8526df680dc20459866aa62005d15ad1200a79/runtime/manta/src/lib.rs#L104
> 
> But if we change this name, I'm not sure whether we need to submit a pr to polkadot-js. We should take some time on testing with frontend, ensure polkadot-js can work well with our configuration.
> 
> BTW, what's the size of new client binary?

Probably we need to. about 125 MB.

node/src/command.rs
stechu: I think there is a way to do this without affecting polkadot.js, you can see statemine's info does not match its `rpc.system.name()`.

It is not possible to add a switch here since it may not have a spec loaded yet.
 
Screen Shot 2021-11-04 at 11 39 07 AM

Dengjianping: So, there must be a way to find the logo respectively without depending on by `impl_name`. I found kusama and polkadot have the same `impl_name`, but polkadot-js shows the correct logo.

Anyway, we have to submit a pr to fix it if we're going to use this `impl_name`.

Dengjianping: Perhaps `Manta Collator` is enough. I don't know, but acala is using `Acala Node`.

Dengjianping:
> > Please rename that dockerfile, because we're going to unify the binary, there's no need to keep that name of dockerfile.
> 
> I think we still need two docker file? since the genesis are different?

Maybe we can get what suggestion from devops' perspective.

node/src/command.rs
stechu: `rpc.system.chain()` works well enough. PR already submitted: https://github.com/polkadot-js/apps/pull/6493

Dengjianping (approved):
LGTM!

PR #250, opened 2021-11-06
Manta shell chain runtime

stechu:
## Description

closes: #XXXX

---

- [x]  runtime configuration
- [x]  mainnet spec
- [x]  testnet spec
- [x] switch the json spec

write a little note why.

- [ ] Wrote unit tests.


runtime/manta/Cargo.toml
```diff
@@ -129,11 +128,9 @@ std = [
 	'frame-system/std',
 	'frame-system-rpc-runtime-api/std',
 	'frame-try-runtime/std',
-	'pallet-assets/std',
```
ghzlatarev, line 132: Why don't we remove these as dependencies altogether.
`pallet_assets` and `pallet_proxy`

runtime/manta/Cargo.toml
```diff
@@ -129,11 +128,9 @@ std = [
...
 	'pallet-authorship/std',
 	'pallet-balances/std',
 	'pallet-multisig/std',
-	'pallet-proxy/std',
 	'pallet-utility/std',
 	'pallet-transaction-payment-rpc-runtime-api/std',
```
ghzlatarev, line 133: seems like no-std is missing for `pallet_scheduler`

runtime/manta/src/lib.rs
```diff
@@ -121,10 +120,10 @@ pub fn native_version() -> NativeVersion {
 
 /// We assume that ~5% of the block weight is consumed by `on_initialize` handlers. This is
 /// used to limit the maximal weight of a single extrinsic.
```
ghzlatarev, line 122: Comment needs to be updated for new code.

runtime/manta/src/lib.rs
```diff
@@ -121,10 +120,10 @@ pub fn native_version() -> NativeVersion {
...
-pub const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(5);
+pub const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(10);
 /// We allow `Normal` extrinsics to fill up the block up to 75%, the rest can be used by
 /// Operational  extrinsics.
```
ghzlatarev, line 125: Comment needs update.

runtime/manta/src/lib.rs
```diff
@@ -151,12 +150,35 @@ parameter_types! {
...
+pub struct MantaFilter;
+impl Contains<Call> for MantaFilter {
+	fn contains(c: &Call) -> bool {
+		match c {
+			Call::Timestamp(_)
+			| Call::ParachainSystem(_)
+			| Call::Authorship(_)
+			| Call::Sudo(_)
+			| Call::Multisig(_)
+			| Call::Scheduler(_)
```
ghzlatarev: Actually why do we need the `Scheduler` when we don't have governace here yet.
Also is `Balances` needed in the shell chain ?

runtime/manta/src/lib.rs
```diff
@@ -224,7 +246,7 @@ impl pallet_balances::Config for Runtime {
 
 parameter_types! {
 	/// Relay Chain `TransactionByteFee` / 10
```
ghzlatarev, line 247: This comment is redundant, basically the same as the code.

runtime/manta/src/lib.rs
```diff
@@ -104,7 +103,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("manta"),
 	impl_name: create_runtime_str!("manta"),
 	authoring_version: 1,
-	spec_version: 1,
+	spec_version: 3090,
```
ghzlatarev, line 106: Shouldn't we start with `3000` in the manta runtime, since it's the shell chain runtime now.

runtime/manta/src/lib.rs
stechu: This is deliberate. The comments are from parity. Here we use a more conservative setting.

runtime/manta/src/lib.rs
stechu: see above.

runtime/manta/src/lib.rs
stechu: Yeah, Balance is needed to help testing (at least do no harm). We could ban `Scheduler`.

runtime/manta/src/lib.rs
stechu: No.

runtime/manta/src/lib.rs
stechu: This is to be consistent with the release version. We are doing a similar semantic versioning on different runtimes with polkadot.

runtime/primitives/src/constants.rs
```diff
@@ -14,10 +14,14 @@
...
 pub const CALAMARI_SS58PREFIX: u8 = 78;
-pub const MANTAPC_SS58PREFIX: u8 = 77;
-pub const MANTA_DECIMAL: u8 = 12;
+pub const CALAMARI_DECIMAL: u8 = 12;
 pub const CALAMARI_TOKEN_SYMBOL: &str = "KMA";
+
+// Manta constants 
+pub const MANTA_SS58PREFIX: u8 = 77;
+pub const MANTA_DECIMAL: u8 = 18;
 pub const MANTA_TOKEN_SYMBOL: &str = "MA";
```
Dengjianping: Are we really going to use `MA` as token name?

Dengjianping:
And resolve `fmt` and `clippy` warning.
Update readme, there're some of words still named `manta-pc`.

Dengjianping:
Because we removed some pallets, so there's some constant storages which are unused.
Like this part https://github.com/Manta-Network/Manta/blob/0ab48290b6ecc5c6b60b5b2272354b4c14d9f750/runtime/manta/src/lib.rs#L290
We can remove it as well. And there're some pallets that will be removed, so perhaps we have to check these `parameter_types` again.

runtime/primitives/src/constants.rs
stechu: good catch! need to use `MANTA`.

runtime/manta/Cargo.toml
stechu: See L138?

Dengjianping (approved):
LGTM!

PR #254, opened 2021-11-08
[Manta] CI fixes

### LINKED ISSUE #252: Re-enable runtime upgrade in CI pipeline after rococo issues are resolved.
Currently this test is disabled in order for the pipeline to complete. Performing runtime upgrades to 0.9.12 on rococo relay chain is problematic. There's no definite answer as to why but the general consensus between para teams is to test on Westrend.
Or perhaps patch the test to not use rococo chain spec for the relay chain.

ghzlatarev:
## Description

closes: #252 

* Revert the rococo runtime upgrade workaround
* Fix erroneous skipping of runtime upgrade test in CI. Now we parse both calamari and manta runtime versions. If any one of them differs from the base branch version, runtime upgrades are triggered for both. 
* Force fail the CI workflow if the new Calamari and Manta runtime versions don't match, as we want to enforce synchronized updates.
* Add a separate manual workflow for Manta runtime weights generation.
---

write a little note why.

- [ ] Wrote unit tests.

### DIFF
--- a/.github/workflows/publish-draft-releases.yml
+++ b/.github/workflows/publish-draft-releases.yml
@@ -1177,11 +1177,9 @@ jobs:
           echo "::set-output name=match::true"
           if [[ ${{ needs.parse-runtimes.outputs.manta-runtime-base }} != ${{ needs.parse-runtimes.outputs.manta-runtime-current }} ]]; then echo "::set-output name=match::false"; fi
           if [[ ${{ needs.parse-runtimes.outputs.calamari-runtime-base }} != ${{ needs.parse-runtimes.outputs.calamari-runtime-current }} ]]; then echo "::set-output name=match::false"; fi
-      -
-        name: print result
-        id: print-result
+      -  
+        name: force fail if new versions don't match
         run: |
-          echo ${{ steps.check-match.outputs.match }}
-          echo ....................
+          if [[  ${{ needs.parse-runtimes.outputs.manta-runtime-current }} !=  ${{ needs.parse-runtimes.outputs.calamari-runtime-current }} ]]; then exit 1; fi
 
 # yamllint enable rule:line-length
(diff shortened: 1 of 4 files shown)


runtime/calamari/src/lib.rs
```diff
@@ -107,7 +107,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("calamari"),
 	impl_name: create_runtime_str!("calamari"),
 	authoring_version: 1,
-	spec_version: 3090,
+	spec_version: 3091,
```
ghzlatarev: Just for the test, will revert.

.github/workflows/publish-draft-releases.yml
```diff
@@ -1132,41 +1140,49 @@ jobs:
...
-        name: check-match
+        name: check if runtime versions match
         id: check-match
         run: |
-          echo "::set-output name=match::false"
-          if [[ ${{ steps.get-runtime-version-base.outputs.runtime-ver-base }} == ${{ steps.get-runtime-version-current.outputs.runtime-ver-current }} ]]; then echo "::set-output name=match::true"
-          fi
+          echo "::set-output name=match::true"
+          if [[ ${{ needs.parse-runtimes.outputs.manta-runtime-base }} != ${{ needs.parse-runtimes.outputs.manta-runtime-current }} ]]; then echo "::set-output name=match::false"; fi
+          if [[ ${{ needs.parse-runtimes.outputs.calamari-runtime-base }} != ${{ needs.parse-runtimes.outputs.calamari-runtime-current }} ]]; then echo "::set-output name=match::false"; fi
```
ghzlatarev, line 1179: We've parsed 4 runtime versions. 2 old and 2 new.
The logic is that if either calamari's runtimes don't match or manta's runtime don't match the test will be triggered.

Dengjianping (approved):
LGTM!

.github/workflows/publish-draft-releases.yml
stechu: Can we fail the test in addition to the echo? (i.e. always require calamari and manta's runtime version are the same).

stechu (changes requested):

.github/workflows/publish-draft-releases.yml
ghzlatarev: Yep can do.

stechu:
@ghzlatarev  Thanks for the PR. Can you try a commit with different runtime version and see whether the CI will fail? Otherwise it is good to go.

stechu:
Also, it appears that you are not using verified commit?

ghzlatarev:
> Also, it appears that you are not using verified commit?



> @ghzlatarev Thanks for the PR. Can you try a commit with different runtime version and see whether the CI will fail? Otherwise it is good to go.

Weird, going to reopen it.

PR #258, opened 2021-11-09
[Manta] CI improvements

### LINKED ISSUE #182: CI/CD Improvements
Some issues concerning the CI/CD pipeline:

- [ ] pin a nightly rust for our CI/CD https://github.com/AcalaNetwork/Acala/blob/master/.github/workflows/test.yml#L39 (Jamie, P2)(Just use stable)
- [x] break up the draft release ci. pull the integration tests out of that workflow, since integration tests should be triggered on every push (Rob, P1)
- [ ] integrate the multi-sig in runtime upgrade test (Gerogi, Gab, P5)
- [x] make the CI checks faster, move everything to self-hosted github actions. (Rob, P2)
- [x] add kusama-calamari to status web  (P3)
- [ ] add columns with role description in status web (P3)
- [x] start a calamari testnet (self hosted)   (P1)
- [ ] add a client upgrade test
- [x] add a bullet in the PR template/description for benchmarking weights
- [ ] fix parsing in the benchmarking jobs to accommodate all pallets
- [x] extend weights generation job to accommodate all pallets
- [x] update workflows that use the block measuring tool because it has been changed
- [x] Fix erroneous skipping of runtime upgrade test in CI. 
- [x]  Force fail the CI workflow if the new Calamari and Manta runtime versions don't match, as we want to enforce synchronized updates.
- [x]  Add a separate manual workflow for Manta runtime weights generation.

ghzlatarev:
## Description

Reopen https://github.com/Manta-Network/Manta/pull/254 with verified commits

closes #182

* Fix erroneous skipping of runtime upgrade test in CI. Now we parse both calamari and manta runtime versions. If any one of them differs from the base branch version, runtime upgrades are triggered for both. 
* Force fail the CI workflow if the new Calamari and Manta runtime versions don't match, as we want to enforce synchronized updates.
* Add a separate manual workflow for Manta runtime weights generation.
* Use the new workflow to add custom weights for `pallet_session`, `frame_system`, `pallet_balances` and `pallet_timestamp` in the Manta runtime. https://github.com/Manta-Network/Manta/actions/runs/1446082289
* Add missing `pallet_session`, `frame_system` and `pallet_timestamp` custom weight file in Calamari runtime https://github.com/Manta-Network/Manta/actions/runs/1446082288

---

write a little note why.

- [ ] Wrote unit tests.


.github/workflows/generate_manta_weights_files.yml
```diff
@@ -0,0 +1,195 @@
...
+    name: benchmark (${{ matrix.benchmark.pallet.name }} ${{ matrix.benchmark.extrinsic.name }})
+    needs:
+      - start-node-builder-current
+      - build-benchmark
+    runs-on: ${{ needs.start-node-builder-current.outputs.runner-label }}
+    strategy:
+      matrix:
+        benchmark:
+          -
+            extrinsic:
```
stechu, line 108: why does the benchmark only contain scheduler and balance pallets?

ghzlatarev: Only added the pallets that overlap with the Calamari workflow, but let's just add all available pallets in both runtime.

ghzlatarev: Actually only `pallet_timestamp` isn't benchmarked at the moment in both runtimes. Everything else is filtered out and I'd suggest adding them as we go.
Running a job now and will add the missing weights:
https://github.com/Manta-Network/Manta/actions/runs/1446082289
https://github.com/Manta-Network/Manta/actions/runs/1446082288

Dengjianping: We need to update weight for `frame_system`. Because there's some calls that will be triggered by users, like `remark`.

ghzlatarev: Updated `frame_system` and `pallet_session` weights.

ghzlatarev:
This needs to be reopened again with verified commits...

PR #270, opened 2021-11-18
[Manta] Whitelist `frame_system` calls and Integrate custom `multisig` weights in Manta/Calamari runtimes

### LINKED ISSUE #269: Add `pallet_mutlisig` weights file to Manta/Calamari
`pallet_multisig` benchmarking was erroring out, because it calls `system.remark()` in its benchmarking code, and we filter out frame_system calls.
No other team filters these calls out so just add them to our whitelist.
Then run a benchmark job and integrate the weights file in both runtimes.

ghzlatarev:
## Description

closes: #269

* Add `frame_system` calls to our base filter in order for `pallet_multisig` benchmarks to not fail with `BadOrigin` error, as that code calls `system.remark()`
* Add `pallet_multisig` to our benchmark workflows.
* Execute the benchmark workflows and add the `pallet_multisig` weights files to Manta (https://github.com/Manta-Network/Manta/actions/runs/1473377864) and Calamari (https://github.com/Manta-Network/Manta/actions/runs/1473377866) runtimes

---

write a little note why.

- [x] Wrote unit tests.


Dengjianping:
I'm not sure why we change both runtimes at one pr, I mean both runtimes have its own schedule to enable functionalities or features.
I think we should dedicate one pr to one runtime instead of both runtimes. Otherwise, if both runtime have the same security issue, we can change them at the same pr.
@stechu @ghzlatarev Any suggestion?

ghzlatarev:
> I'm not sure why we change both runtimes at one pr, I mean both runtimes have its own schedule to enable functionalities or features. I think we should dedicate one pr to one runtime instead of both runtimes. Otherwise, if both runtime have the same security issue, we can change them at the same pr. @stechu @ghzlatarev Any suggestion?

Yeah we should probably roll out runtime changes on Kusama first, then Polkadot. But I wonder how we're going to keep the releases syncronized and have different code ?

stechu (approved):

PR #296, opened 2021-12-08
[manta] Adjust Calamari weight to fee calculation

### LINKED ISSUE #299: Increase tx fees to be at least as expensive as Polkadot/Kusama/Acala/Moonbeam dollar wise

ghzlatarev:
## Description

closes: #299

* Adjust weight to fee calculation so our tx fees are at least as expensive as Polkadot/Kusama/Moonbeam/Acala. Compared `balance.transfer` and `system.remark` :
* transfer on Polkadot is roughly $0.47
* transfer on Kusama is roughly $0.016
* transfer on Acala is roughly $0.01
* transfer on Moonriver is roughly $0.25
* transfer on Calamari was $0.000036 and now is roughly $0.036

---

write a little note why.

- [ ] Wrote unit tests.

### DIFF
--- a/runtime/calamari/src/fee.rs
+++ b/runtime/calamari/src/fee.rs
@@ -54,11 +54,11 @@ impl WeightToFeePolynomial for WeightToFee {
 		// daily_cost_to_fully_congest = 0.000281626 * coeff * 1134 * 7200 * 0.02 = 45.988399296 * coeff
 		// Assuming we want the daily cost to be around $250000 we get:
 		// 250000 = 45.988399296 * coeff
-		// coeff = 5436
+		// coeff = ~5436
 
 		// Keep in mind this is a rough worst-case scenario calculation.
-		// The length_fee could not be negligible, and the targeted_fee_adjustment will hike the fees,
-		// as the network gets congested more and more congested, and this will further increase the costs.
+		// The `length_fee` could not be negligible, and the `targeted_fee_adjustment` will hike the fees
+		// as the network gets more and more congested, which will further increase the costs.
 		smallvec![WeightToFeeCoefficient {
 			coeff_integer: 5000u32.into(),
 			coeff_frac: Perbill::zero(),
--- a/runtime/calamari/src/fee.rs
+++ b/runtime/calamari/src/fee.rs
@@ -40,16 +40,11 @@ pub struct WeightToFee;
 impl WeightToFeePolynomial for WeightToFee {
 	type Balance = Balance;
 	fn polynomial() -> WeightToFeeCoefficients<Self::Balance> {
-		// in Polkadot, extrinsic base weight (smallest non-zero weight) is mapped to 1/10 CENT:
-		// in Manta Parachain, we map to 1/10 of that, or 1/100 CENT
-		// TODO, revisit here to figure out why use this polynomial
-		let p = currency::cKMA;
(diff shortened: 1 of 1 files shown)


runtime/calamari/src/fee.rs
```diff
@@ -40,16 +40,11 @@ pub struct WeightToFee;
...
 	type Balance = Balance;
 	fn polynomial() -> WeightToFeeCoefficients<Self::Balance> {
-		// in Polkadot, extrinsic base weight (smallest non-zero weight) is mapped to 1/10 CENT:
-		// in Manta Parachain, we map to 1/10 of that, or 1/100 CENT
-		// TODO, revisit here to figure out why use this polynomial
-		let p = currency::cKMA;
-		let q = 100 * Balance::from(ExtrinsicBaseWeight::get());
-		smallvec![WeightToFeeCoefficient {
-			degree: 1,
+		smallvec!(WeightToFeeCoefficient {
```
Dengjianping: Can you use [] instead of ()? Like
```rust
smallvec![
    WeightToFeeCoefficient{...},
    WeightToFeeCoefficient{...},
    WeightToFeeCoefficient{...},
]
```
I think we will use different polynomial the future.
And put what exact current polynomial here.

ghost (approved):
:+1:

stechu:
For current block weight, this means a single block cause 43.2 USD to attack?

ghzlatarev:
> For current block weight, this means a single block cause 43.2 USD to attack?

Right, based on Jamie's results of 1134 transactions in a block I get ~41USD

Dengjianping:
> > For current block weight, this means a single block cause 43.2 USD to attack?
> 
> Right, based on Jamie's results of 1134 transactions in a block I get ~41USD

Once we change fees calculation, I think we should re-benchmark all related pallets.
And perhaps `1134` is not correct any more.

ghzlatarev:
> > > For current block weight, this means a single block cause 43.2 USD to attack?
> > 
> > 
> > Right, based on Jamie's results of 1134 transactions in a block I get ~41USD
> 
> Once we change fees calculation, I think we should re-benchmark all related pallets. And perhaps `1134` is not correct any more.

This should probably be the last PR to merge then.

stechu:
Per the discussion on Discord, we need a clear understanding of the fee dynamics under network congestion.  I personally don't think a fixed fee is a good design.

runtime/calamari/src/fee.rs
```diff
@@ -40,16 +38,12 @@ pub struct WeightToFee;
...
 	fn polynomial() -> WeightToFeeCoefficients<Self::Balance> {
-		// in Polkadot, extrinsic base weight (smallest non-zero weight) is mapped to 1/10 CENT:
-		// in Manta Parachain, we map to 1/10 of that, or 1/100 CENT
-		// TODO, revisit here to figure out why use this polynomial
-		let p = currency::cKMA;
-		let q = 100 * Balance::from(ExtrinsicBaseWeight::get());
+		// fee = coeff_integer * (weight ^ degree) + coeff_friction * (weight ^ degree)
 		smallvec![WeightToFeeCoefficient {
-			degree: 1,
+			coeff_integer: 5000u32.into(),
```
stechu, line 63: explain how does 5000 come from?

stechu (changes requested):

Dengjianping (approved):

stechu (approved):

PR #328, opened 2022-01-05
[Manta] Increase Calamari existential deposit

### LINKED ISSUE #327: Increase Calamari's existential deposit
Needs to be adjusted to reflect the tx fees increase in 3.1.1

This feature is used to prevent bloating of the network with spam accounts, which degrade the performance of the storage
So we should think about it like what is the appropriate cost to create some large number of accounts, 1 billion for example. 
The question is how to benchmark and measure how many spam accounts can be created before performance degrades a lot.
Furthermore the dollar value also shouldn't be over $5 or ideally over $1. A lot of threads about decreasing Polkadot's 1 DOT ($25+). But we have a lot of room to grow ours.

Another potential issue is some accounts may have more than enough existential deposit now, but not after the update, which can effectively make their accounts unusable but not reaped. Such accounts need to be identified and airdropped some funds. Or some other way of handling this needs to be figured out.

Existential deposit has been increased in the past by other chains like Kusama - https://github.com/paritytech/polkadot/pull/341/files#diff-0ec06ea58bd455f09ce6b3bb4c2c1c0d37bda51c1e1be2151c560c9c973959ecL189-L191 . Consult with them how they handle the situation if needed.

ghzlatarev:
## Description

closes: #327

* Transaction fees were greatly increased in 3.1.1 release but the existential deposit wasn't adjusted. Adjusting it to 1.5 KMA for now. (for reference a `balances.transferKeepAlive` is currently around 1.5KMA as well)

---

write a little note why.

- [ ] Wrote unit tests.

### DIFF
--- a/runtime/calamari/src/lib.rs
+++ b/runtime/calamari/src/lib.rs
@@ -285,7 +285,7 @@ impl pallet_authorship::Config for Runtime {
 }
 
 parameter_types! {
-	pub const NativeTokenExistentialDeposit: u128 = 10 * cKMA; // 0.1 KMA
+	pub const NativeTokenExistentialDeposit: u128 = 2 * KMA;
 	pub const MaxLocks: u32 = 50;
 	pub const MaxReserves: u32 = 50;
 }
--- a/runtime/calamari/src/lib.rs
+++ b/runtime/calamari/src/lib.rs
@@ -285,7 +285,7 @@ impl pallet_authorship::Config for Runtime {
 }
 
 parameter_types! {
-	pub const NativeTokenExistentialDeposit: u128 = 2 * KMA;
+	pub const NativeTokenExistentialDeposit: u128 = 150 * cKMA;
 	pub const MaxLocks: u32 = 50;
 	pub const MaxReserves: u32 = 50;
 }


runtime/calamari/src/lib.rs
```diff
@@ -285,7 +285,7 @@ impl pallet_authorship::Config for Runtime {
 }
 
 parameter_types! {
-	pub const NativeTokenExistentialDeposit: u128 = 10 * cKMA; // 0.1 KMA
+	pub const NativeTokenExistentialDeposit: u128 = 2 * KMA;
```
Dengjianping: I think it should be smaller due to LTO enabled.

ghzlatarev: @Dengjianping agreed let's make it 1.5KMA
For reference a balances keep alive transfer is around 1.5KMA now as well.

runtime/calamari/src/lib.rs
```diff
@@ -285,7 +285,7 @@ impl pallet_authorship::Config for Runtime {
 }
 
 parameter_types! {
-	pub const NativeTokenExistentialDeposit: u128 = 10 * cKMA; // 0.1 KMA
+	pub const NativeTokenExistentialDeposit: u128 = 150 * cKMA;
```
ghzlatarev, line 288: More research needed for minimal value.
Initial thoughts between 1 and 10 KMA.

ghzlatarev:
```
						 _______________________________________________________________________________________________________________________________________________________________
									|                                                                       |                                                       |
						 Network    |   						Existential Deposit							|						Transfer						|		Market Cap
						 ___________|_______________________________________________________________________|_______________________________________________________|_____________________
						 Calamari  	|		0.1  KMA 		* 	$0.01194752 	= 		$0.001194752  		|		630.3619 milli KMA 	=	 $0.007531261407488		|		$118,103,100
						 Phala  	|		0.01 PHA 		* 	$0.359294 		= 		$0.00359294			|		1.5807 milli PHA   	=	 $0.0005679360258		|		$359,293,991
						 Karura  	|		0.1  KAR 		* 	$3.03 			= 		$0.303				|		2.4700 milli KAR   	=	 $0.0074841				|		$303,258,074
						 KILT  		|		0.01 KILT 		* 	$2.89 			= 		$0.0289				|		4.2263 milli KILT  	=	 $0.012214007			|		$842,979,173
						 KSM  		|		0.000033 KSM 	* 	$246.97 		= 		$0.0082323 			|		52.3328 micro KSM  	=	 $0.012924631616		|		$2,460,597,905
```

PR #337, opened 2022-01-09
[Manta] Add a `congested_chain_simulation` test in Calamari

### LINKED ISSUE #305: Add congestion test in Calamari
Something similar to:
https://github.com/paritytech/substrate/blob/master/bin/node/runtime/src/impls.rs#L213

This can be used to explore how modifying TargetedFeeAdjustment affects how quickly TX fees are hiked. https://paritytech.github.io/substrate/master/pallet_transaction_payment/struct.TargetedFeeAdjustment.html

ghzlatarev:
## Description

closes: #305 

* Test to calculate how much it would cost someone to fully congest Calamari for 1 day (7200 blocks). The test fails if the total cost is less than $250k (currently comes out to ~$315392)
* The test takes the total total block weight and runs it through the weight to fee routine and then adjusts it with the variable adjustment variable. This step is repeated 7200 and the total fee is accumulated.
* In time the `kma_price` may need to be adjusted depending on market fluctuations and `target_daily_congestion_cost_usd` can be adjusted depending on our preference.

Example output:
congested_test

---

write a little note why.

- [x] Wrote unit tests.

### DIFF
--- a/runtime/calamari/src/fee.rs
+++ b/runtime/calamari/src/fee.rs
@@ -83,6 +83,20 @@ mod multiplier_tests {
 		>::convert(fm)
 	}
 
+	fn fetch_kma_price() -> Result<f32, &'static str> {
+		let body = reqwest::blocking::get(
+			"https://api.coingecko.com/api/v3/simple/price?ids=calamari-network&vs_currencies=usd",
+		)
+		.unwrap();
+		let json_reply: serde_json::Value = serde_json::from_reader(body).unwrap();
+		if let Some(price) = json_reply["calamari-network"]["usd"].as_f64() {
+			// CG API return: {"calamari-network":{"usd": 0.01092173}}
+			Ok(price as f32)
+		} else {
+			Err("KMA price not found in reply from Coingecko. API changed? Check https://www.coingecko.com/en/api/documentation")
+		}
+	}
+
 	// Consider the daily cost to fully congest our network to be defined as:
 	// `target_daily_congestion_cost_usd = inclusion_fee * blocks_per_day * kma_price`
 	// Where:
@@ -94,9 +108,10 @@ mod multiplier_tests {
 	// This test loops 1 day of parachain blocks (7200) and calculates accumulated fee if every block is almost full
 	#[test]
 	fn congested_chain_simulation() {
-		// Configure these values depending on current prices and target cost.
-		let kma_price = 0.012f32;
+		// Configure the target cost depending on the current state of the network.
 		let target_daily_congestion_cost_usd = 250000;
+		let kma_price = fetch_kma_price().unwrap();
+		println!("KMA/USD price as read from CoinGecko = {}", kma_price);
 		let target_daily_congestion_cost_kma =
 			(target_daily_congestion_cost_usd as f32 * kma_price * KMA as f32) as u128;
 
--- a/runtime/calamari/src/fee.rs
+++ b/runtime/calamari/src/fee.rs
@@ -38,32 +38,114 @@ pub struct WeightToFee;
 impl WeightToFeePolynomial for WeightToFee {
 	type Balance = Balance;
 	fn polynomial() -> WeightToFeeCoefficients<Self::Balance> {
-		// Consider the daily cost to fully congest our network to be defined as:
-		// daily_cost_to_fully_congest = inclusion_fee * txs_per_block * blocks_per_day * kma_price
-		// The weight fee is defined as:
-		// weight_fee = coeff_integer * (weight ^ degree) + coeff_friction * (weight ^ degree)
-		// The inclusion fee is defined as:
-		// inclusion_fee = base_fee + length_fee + [targeted_fee_adjustment * weight_fee]
-		// As of the day of writing this code a single `balances.transfer` is benchmarked at 156626000 weight.
-		// Let's assume worst case scenarios where the `length_fee` of a transfer is negligible,
-		// and that `targeted_fee_adjustment` is simply 1, as if the network is not congested.
-		// Furthermore we know the `base_fee` is 0.000125KMA defined in our runtime. So:
-		// inclusion_fee = 0.000125 * coeff + 0.000156626 * coeff = 0.000281626 * coeff
-		// We have profiled `txs_per_block` to be around 1134 and `blocks_per_day` is known to be 7200.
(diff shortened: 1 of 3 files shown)


runtime/calamari/src/fee.rs
```diff
@@ -67,3 +47,104 @@ impl WeightToFeePolynomial for WeightToFee {
...
+	// Where:
+	// `fee_adjustment` and `weight_to_fee_coeff` are configurable in a runtime via `FeeMultiplierUpdate` and `WeightToFee`
+	// `fee_adjustment` is also variable depending on previous block's fullness
+	// We are also assuming `length_fee` is negligible for small TXs like a remark or a transfer.
+	// This test loops 1 day of parachain blocks (7200) and calculates accumulated fee if every block is almost full
+	#[test]
+	fn congested_chain_simulation() {
+		// Configure these values depending on current prices and target cost.
+		let kma_price = 0.012f32;
+		let target_daily_congestion_cost_usd = 250000;
```
Garandor, line 112: If i get this correctly, this test is a sanity check for the fee-related config options in the runtime.
The test logic seems sound, to reduce maintenance effort , i would propose

1. Scale the target usd cost with kma price like so:

```
let current_kma_price = 0.012f32
let orig_kma_price = 0.012f32;
let target_daily_congestion_cost_usd = 250000 * (std::cmp::max(1,current_kma_price / orig_kma_price) as u32) ;
```

That way the USD price target doesn't have to be manually adjusted to reflect the (growing) economic value of the network.
At current marketcap the 250k target equals 0.2% of the network and assuming KMA price is an accurate reflection of network value, it will stay at 0.2% that way.
Censoring a $100 mil network is likely not worth 250k, but a $10 billion network might very well be.

( current price ofc still has to be manually adjusted )

2. there should be a pre-release check if the current changeset has modified the kma_price line  ( and/or a pre-commit hook on release branches that fetches KMA price and updates it automatically) though, otherwise we risk this whole thing going out of date and not detecting no longer acceptable fee parameters.

If we don't do pre-commit hooks, i would recommend we annotate code lines that should be updated with each release with a standardized comment, so they are easy to grep for, something like 
` let kma_price = 0.012f32;  // CHECK_ON_REL `
or keep them in a dedicated constants.rs

runtime/calamari/src/fee.rs
```diff
@@ -67,3 +47,104 @@ impl WeightToFeePolynomial for WeightToFee {
...
+		let base_fee = <Runtime as pallet_transaction_payment::Config>::WeightToFee::calc(
+			&frame_support::weights::constants::ExtrinsicBaseWeight::get(),
+		);
+
+		run_with_system_weight(block_weight, || {
+			// initial value configured on module
+			let mut fee_adjustment = Multiplier::one();
+			assert_eq!(fee_adjustment, TransactionPayment::next_fee_multiplier());
+			let mut accumulated_fee: u128 = 0;
+			// Simulates 1 day of parachian blocks (12 seconds each)
```
Garandor: typo: parachian

runtime/calamari/src/fee.rs
ghzlatarev: What's the ceil for ? https://docs.rs/libmath/0.1.4/math/round/fn.ceil.html

Otherwise it's not a bad idea to scale it with the total value of the network. But then again $250k (actually $310k) is still going to be a lot of money for someone even if network is valuable.
What do you guys think @Dengjianping @stechu

runtime/calamari/src/fee.rs
ghzlatarev: Thx

runtime/calamari/src/fee.rs
Garandor: std::cmp::max was what i intended to use there

Garandor: the coingecko API can fetch KMA price using below REST call 
https://api.coingecko.com/api/v3/simple/price?ids=calamari-network&vs_currencies=usd
and returns
```
{
  "calamari-network": {
    "usd": 0.01092173
  }
}
```
I'll try to whip up a call integrated into this test using reqwest

Dengjianping (approved):
LGTM!

runtime/calamari/src/fee.rs
```diff
@@ -67,3 +47,119 @@ impl WeightToFeePolynomial for WeightToFee {
...
+		println!("KMA/USD price as read from CoinGecko = {}", kma_price);
+		let target_daily_congestion_cost_kma =
+			(target_daily_congestion_cost_usd as f32 * kma_price * KMA as f32) as u128;
+
+		// `cargo test --package calamari-runtime --lib -- fee::multiplier_tests::congested_chain_simulation --exact --nocapture` to get some insight.
+		// almost full. The entire quota of normal transactions is taken.
+		let block_weight = BlockWeights::get()
+			.get(DispatchClass::Normal)
+			.max_total
+			.unwrap() - 10;
```
Garandor, line 123: Why the -10?

runtime/calamari/src/fee.rs
```diff
@@ -67,3 +47,119 @@ impl WeightToFeePolynomial for WeightToFee {
...
+		run_with_system_weight(block_weight, || {
+			// initial value configured on module
+			let mut fee_adjustment = Multiplier::one();
+			assert_eq!(fee_adjustment, TransactionPayment::next_fee_multiplier());
+			let mut accumulated_fee: u128 = 0;
+			// Simulates 1 day of parachain blocks (12 seconds each)
+			for iteration in 0..7200 {
+				let next = runtime_multiplier_update(fee_adjustment);
+				// if no change, panic. This should never happen in this case.
+				if fee_adjustment == next {
```
Garandor: can next not be returned as something smaller than the last iteration?

Garandor:
Added 2 questions, LGTM otherwise

runtime/calamari/src/fee.rs
ghzlatarev: So it's almost full. It doesn't matter actually.

runtime/calamari/src/fee.rs
ghzlatarev: No, but let's just put a sanity check anyway.

Garandor (approved):

PR #358, opened 2022-01-20
[Manta] Auto-evict low-performing collators each session

### LINKED ISSUE #335: Evicting Underperforming Collators
# Goal 

We want to modify `manta-collator-selection` pallet to:

- Monitor block producing data of collators 
- Use this data to kick off candidate collators (excluding `invulerables`, which is a set of collators ran by ourselves) based on certain conditions

# Block Producing Data
A simple and effective block producing data is how many block this collator signed during in a section. It is easy to implement and is a good metric to measure how well a collator does in block production.

# Evicting Condition

Evicting condition need to set to remove `abnormal` collators, this condition should be relative robust, for example:

1. When all the collators are all doing well, we should not evict any collators. Thus, a simple percentile based approach would not work. We don't want to create a hostile environment for collator runners.

2. It should detect outliers reliably, so that the block production is healthy. 

# Reference Design

Acala Collator Evicting Logic: https://github.com/AcalaNetwork/Acala/blob/master/modules/collator-selection/src/lib.rs#L529

Garandor:
## Description

This PR modifies the manta-collator-selection pallet by adding three new storage items and corresponding setter extrinsics, that - if set to anything other than their default - automatically remove collators at the end of a session that significantly ( x% ) underperformed the y-th percentile of the total collator set over the current session.

A good initial value (and we intend to set this on Dolphin) is
- x = 10 ( default value on genesis: 100)
- y = 80 ( default value on genesis: 0)

Main changes in: pallets/collator-selection/src/lib.rs

- Substitutes the kick_threshold and last_block_authored storage items with two new storage items, defining the percentile and the percentage above
- Adds a setter extrinsic for each of the new items
- Adds a storage map that records the number of blocks produced by each collator in the current session

closes: #335

---

write a little note why.

- [x] Wrote unit tests.
- ~[ ] If needed, notify the committer this is a draft-release and a tag is needed after merging the PR.~
- ~[ ] If needed, bump `version` for every crate.~
- ~[ ] If needed, update our Javascript/Typescript APIs. These APIs are offcially used by exchanges or community developers.~
- ~[ ] If we're going to issue a new release, freeze the code one week early(it depends, but usually it's one week), ensure we have enough time for related testing.~


pallets/collator-selection/src/lib.rs
```diff
@@ -178,11 +190,14 @@ pub mod pallet {
...
+	#[pallet::type_value]
+	pub(super) fn StartingBlockCount() -> BlockCount {
+		0u32.into()
+	}
 	#[pallet::storage]
-	#[pallet::getter(fn last_authored_block)]
-	pub type LastAuthoredBlock<T: Config> =
-		StorageMap<_, Twox64Concat, T::AccountId, T::BlockNumber, ValueQuery>;
+	pub(super) type BlocksPerCollatorThisSession<T: Config> =
+		StorageMap<_, Blake2_128Concat, T::AccountId, BlockCount, ValueQuery, StartingBlockCount>; // RAD: Note: AccountId is user-selectable, but protected by ED, is that enough?
```
Garandor: AccountId is user-selectable, but protected by ED, is that enough to use the non cryptographic hashing algo instead of Blake2_128Concat?

pallets/collator-selection/src/lib.rs
```diff
@@ -499,29 +512,83 @@ pub mod pallet {
...
+			if collator_perf_this_session.is_empty() {
+				return Vec::new();
+			}
+			// 1. Sort collator performance list
+			collator_perf_this_session.sort_unstable_by_key(|k| k.1); // XXX: don't like the tuple accessor, could this be a struct?
+														  // collator_perf_this_session.reverse();
+			let no_of_candidates = collator_perf_this_session.len();
+
+			// 2. get percentile by _exclusive_ nearest rank method https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method (rust percentile API is feature gated)
+			// RAD: Doublecheck that f64 as usize actually rounds up
```
Garandor: It actually truncates, so there needs to be a non_std equivalent of math::round::ceil here to satisfy https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method , any recommendations?

pallets/collator-selection/src/lib.rs
```diff
@@ -499,29 +512,83 @@ pub mod pallet {
...
+			let ordinal_rank = (((T::PerformancePercentileToConsiderForKick::get() as f64) / 100.0
+				* no_of_candidates as f64) as usize)
+				.saturating_sub(1); // Note: -1 to accomodate 0-index counting // RAD: Is there API to saturate but notify on non-overflow so we can log a warning?
```
Garandor: Is there API to saturate but notify on a prevented underflow so we can log a warning?

pallets/collator-selection/src/lib.rs
```diff
@@ -499,29 +512,83 @@ pub mod pallet {
...
+				.collect::<Vec<_>>();
+			let mut removed_account_ids: Vec<T::AccountId> = Vec::new();
+			let kick_candidates = collator_perf_this_session[..ordinal_rank] // ordinal-rank exclusive, the collator with percentile perf is safe
+				.iter()
+				.map(|acc_info| acc_info.0.clone())
+				.collect::<Vec<_>>();
+			kick_candidates.into_iter().for_each(|acc_id| {
+				// If we're not a candidate we're invulnerable or already kicked
+				if current_candidate_ids.contains(&acc_id) {
+					let my_blocks_this_session = <BlocksPerCollatorThisSession<T>>::get(&acc_id); // RAD: read storage or find by iterating collator_perf_this_session vec, which is faster
```
Garandor: Should I read from storage or find by iterating through the collator_perf_this_session vec, which is faster
Storage is slow, but it's element access.
Stackvar is fast, but it's iteration until the key is found

pallets/collator-selection/src/lib.rs
```diff
@@ -540,7 +607,11 @@ pub mod pallet {
 			// `reward` is half of pot account minus ED, this should never fail.
 			let _success = T::Currency::transfer(&pot, &author, reward, KeepAlive);
 			debug_assert!(_success.is_ok());
-			<LastAuthoredBlock<T>>::insert(author, frame_system::Pallet::<T>::block_number());
+
+			// increment blocks this node authored
+			let mut authored_blocks = <BlocksPerCollatorThisSession<T>>::get(&author); // RAD: having this be an Option Query and throwing NotACandidate error would mess up testing code, hjow to resolve this?
```
Garandor: With a ValueQuery, this returns 0 if the validator doesnt exist, so it gets added.

Having this be an Option Query and throwing NotACandidate error would mess up testing code, how to resolve this?
(tests.rs code authors all blocks by origin 4, which is not part of the candidate or validator set at genesis)

pallets/collator-selection/src/lib.rs
```diff
@@ -564,22 +636,36 @@ pub mod pallet {
...
+			// Build performance map for the currently active validatorset in the just started session
+			Self::reset_collator_performance();
 			Some(result)
 		}
 		fn start_session(_: SessionIndex) {
 			// we don't care.
 		}
 		fn end_session(_: SessionIndex) {
 			// we don't care.
+			// RAD: Should we do kicking on end_session instead?
```
Garandor: Should we do kicking on end_session instead?

There's no real difference in function, it might make it clearer though that we're using performance data of the just ended session.

Not sure how to add `frame_system::Pallet::<T>::register_extra_weight_unchecked` to `end_session` though if we were to go that route (kick_stale_candidates mutates storage)

pallets/collator-selection/src/lib.rs
```diff
@@ -499,29 +512,83 @@ pub mod pallet {
...
-				})
-				.collect::<Vec<_>>()
+				}
+			});
+			removed_account_ids
+		}
+
+		/// Reset the performance map to the currently active validators at 0 blocks
+		pub fn reset_collator_performance() {
+			// XXX: intersect BPC map with validators and try_mutate to 0 oder or drop and recreate from scratch?
```
Garandor: Should we intersect the BlocksPerCollatorThisSession map with validators() and try_mutate to 0 or drop the map with remove_all and recreate from scratch?

Latter is much more concise, don't know about performance of the former
According to my research, rotate_session() uses up all the block, so high weight should not be an issue

pallets/collator-selection/src/lib.rs
```diff
@@ -499,29 +512,86 @@ pub mod pallet {
...
+
+		/// Removes collators with unsatisfactory performance
+		/// Returns the removed AccountIds
 		pub fn kick_stale_candidates(
 			candidates: Vec<CandidateInfo<T::AccountId, BalanceOf<T>>>,
 		) -> Vec<T::AccountId> {
-			let now = frame_system::Pallet::<T>::block_number();
-			let kick_threshold = T::KickThreshold::get();
-			candidates
+			use num_traits::float::FloatCore; // For no_std compatible f64::ceil()
```
Garandor: This import is needed to pull the f64::ceil() fn into scope but I am getting a

```
warning: unused import: `num_traits::float::FloatCore`
   --> pallets/collator-selection/src/lib.rs:521:8
    |
521 |             use num_traits::float::FloatCore; 
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_imports)]` on by default
```
Probably because in cases where `std` is available, ceil from `Float` instead of `FloatCore` is preferred.
Any hints on how to squelch that?

Garandor:
This PR is functional but not fully ready - i have added comments where I still have open questions myself - please have a look and advise how to best resolve them

pallets/collator-selection/src/lib.rs
Garandor: resolved by using f64::ceil() from the FloatCore trait

pallets/collator-selection/src/lib.rs
```diff
@@ -499,29 +512,86 @@ pub mod pallet {
...
+			if collator_perf_this_session.is_empty() {
+				return Vec::new();
+			}
+			// 1. Sort collator performance list
+			collator_perf_this_session.sort_unstable_by_key(|k| k.1); // XXX: don't like the tuple accessor, could this be a struct?
+														  // collator_perf_this_session.reverse();
+			let no_of_candidates = collator_perf_this_session.len();
+
+			// 2. get percentile by _exclusive_ nearest rank method https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method (rust percentile API is feature gated)
+			let ordinal_rank = (num_traits::float::FloatCore::ceil(
```
Garandor: There is actually one "problem" with this algorithm in that for validator counts below 5 it uses the highest performing validator for the 80-th percentile. That's not necessarily a bad thing, but something to keep in mind, as for low validator counts this approach is equivalent to "kick everyone 10% worse than the best collator" ( unless invunerable )

I think we already run 5+ nodes on calamari ourselves, so it shouldn't be an issue

image

pallets/collator-selection/src/lib.rs
bhgomes: Suggested way is something like the following:

```rust
let consider_for_kick_percentile = T::PerformancePercentileToConsiderForKick::get();
let ordinal_rank = match (((no_of_candidates * (consider_for_kick_percentile as usize) as f64) * 0.01)
    .ceil() as usize)
    .checked_sub(1)
{
    Some(ordinal_rank) => ordinal_rank,
    _ => {
        log::info!(...);
        0
    }
};
```

and then you can use `consider_for_kick_percentile` later on [line 550](https://github.com/Manta-Network/Manta/pull/358/files#diff-a0359a0add814d66aac74493f7e17aefdcee21d3de1ae592e97059082b6feedfR550) instead of doing another database GET.

pallets/collator-selection/src/lib.rs
Garandor: resolved by using libm::ceil() and f64

pallets/collator-selection/src/lib.rs
Garandor: thanks, I added your  `consider_for_kick_percentile` suggestion.

Turns out we don't need checked_sub because the saturation case can not happen (due to new precondition check)

pallets/collator-selection/src/lib.rs
```diff
@@ -499,29 +512,96 @@ pub mod pallet {
...
-			let now = frame_system::Pallet::<T>::block_number();
-			let kick_threshold = T::KickThreshold::get();
-			candidates
+			// 0. Storage reads and precondition checks
+			if candidates.is_empty() {
+				return Vec::new(); // No candidates means we're running invulnerables only
+			}
+			let percentile_for_kick = T::PerformancePercentileToConsiderForKick::get();
+			if percentile_for_kick == 0 {
+				return Vec::new(); // Selecting 0-th percentile disables kicking. Upper bound check in fn build()
```
Dengjianping, line 571: How about `return None`, and the return can be defined as `Option<Vec<T::AccountId>>`?

pallets/collator-selection/src/lib.rs
```diff
@@ -499,29 +512,96 @@ pub mod pallet {
...
 		) -> Vec<T::AccountId> {
-			let now = frame_system::Pallet::<T>::block_number();
-			let kick_threshold = T::KickThreshold::get();
-			candidates
+			// 0. Storage reads and precondition checks
+			if candidates.is_empty() {
+				return Vec::new(); // No candidates means we're running invulnerables only
+			}
+			let percentile_for_kick = T::PerformancePercentileToConsiderForKick::get();
+			if percentile_for_kick == 0 {
```
Dengjianping: I suggest to use `Zero::zero()` and `One::one()` if there's 0 or 1.

pallets/collator-selection/src/lib.rs
```diff
@@ -136,19 +141,26 @@ pub mod pallet {
 		/// Used only for benchmarking.
 		type MaxInvulnerables: Get<u32>;
 
-		// Will be kicked if block is not produced in threshold.
-		type KickThreshold: Get<Self::BlockNumber>;
+		// n-th Percentile of lowest-performing collators to be checked for kicking
+		type PerformancePercentileToConsiderForKick: Get<u8>;
```
Dengjianping: If this is for calculating percentage, I suggest you to use `Percent` or `Permill` instead of u8.
```rust
let _ = Percent::zero(); // 0%
let _ = Percent::one(); // 100%
let p = Percent::from_perent(40); // 40%
let _: u32 = p.saturated_mul(30u32); // the result is still u32
```
Is it a fixed value, or can we add a call to modify it?

pallets/collator-selection/src/lib.rs
Garandor: good point, thanks!
I didnt think of changing the signature of kick_stale_candidates

Garandor: added in [341b9e9](https://github.com/Manta-Network/Manta/pull/358/commits/341b9e95bce0a18f9c7afeed791013379281a2a1)

pallets/collator-selection/src/lib.rs
Garandor: My intuition is that iterating a stack var will be faster for small/medium validator sets, as i recall a storage read takes ~25ms

pallets/collator-selection/src/lib.rs
Garandor: I'll leave this resolved with the drop-and-recreate solution due to above reasoning

pallets/collator-selection/src/lib.rs
Garandor: changing to vector operation

pallets/collator-selection/src/lib.rs
Garandor: sticking with Blake2 as the storage item will be small in size ( < 1000 ) so performance impact should be negligible

pallets/collator-selection/src/lib.rs
Garandor: as validators are constant over a session and we're adding all validators to the map at the beginning of a new session, block production by a validator not in the map will not occur

pallets/collator-selection/src/lib.rs
Garandor: done

Garandor:
Only one unresolved open question left and that one is strictly optional to implement.

This is now ready for final review

P.S. I'll see on the side if i can get rid of the pulled in ValidatorSet trait as an optional refactor

pallets/collator-selection/src/lib.rs
Garandor: oh yeah, makes sense to do it in the type definition instead of during the computation.

pallets/collator-selection/src/lib.rs
Garandor: resolved in the most recent commit

pallets/collator-selection/src/lib.rs
Garandor: > Is it a fixed value, or can we add a call to modify it?

At the moment it is a runtime config variable that is hard-set at genesis.
I agree that it would make sense to allow for later changes by adding a callable to mutate it.
Working on it

ghzlatarev (commented):
@Garandor The CI is failing.

Garandor:
Reworking this to add extrinsics to change the percentile and threshold.
Ended the day yesterday with benchmarking problems

runtime/calamari/Cargo.toml
```diff
@@ -17,6 +17,7 @@ smallvec = "1.6.1"
 
 # Substrate primitives
 sp-api = { git = 'https://github.com/paritytech/substrate.git', default-features = false, branch = "polkadot-v0.9.15" }
+sp-arithmetic      = { git = 'https://github.com/paritytech/substrate.git', default-features = false, branch = "polkadot-v0.9.15" }
```
ghzlatarev: Formatting

runtime/dolphin/Cargo.toml
```diff
@@ -17,6 +17,7 @@ smallvec = "1.6.1"
 
 # Substrate primitives
 sp-api = { git = 'https://github.com/paritytech/substrate.git', default-features = false, branch = "polkadot-v0.9.15" }
+sp-arithmetic      = { git = 'https://github.com/paritytech/substrate.git', default-features = false, branch = "polkadot-v0.9.15" }
```
ghzlatarev: formatting

runtime/dolphin/src/lib.rs
```diff
@@ -791,6 +792,8 @@ parameter_types! {
 
 parameter_types! {
 	pub const ExecutiveBody: BodyId = BodyId::Executive;
+	pub const PerformancePercentileToConsiderForKick : Percent = Percent::from_percent(80);
+	pub const UnderperformPercentileByPercentToKick  : Percent = Percent::from_percent(10);
```
ghzlatarev: fmt

runtime/calamari/src/lib.rs
```diff
@@ -792,6 +793,8 @@ parameter_types! {
 
 parameter_types! {
 	pub const ExecutiveBody: BodyId = BodyId::Executive;
+	pub const PerformancePercentileToConsiderForKick : Percent = Percent::from_percent(80);
+	pub const UnderperformPercentileByPercentToKick  : Percent = Percent::from_percent(10);
```
ghzlatarev: fmt

pallets/collator-selection/src/lib.rs
ghzlatarev: This looks resolved now ?

runtime/calamari/Cargo.toml
Garandor: it's kinda strange how `cargo fmt` is executed autmatically on file save in code, but somehow, sometimes it just finishes without actually reformatting.

runtime/calamari/src/lib.rs
ghzlatarev: Can we rename `UnderperformPercentileByPercentToKick` to something without `percentile` in the name.
`UnderperformancePrecentToKick` maybe.
Maybe it's just me but mixing percent and percentile got me confused a bit.

runtime/calamari/Cargo.toml
ghzlatarev: Yeah also not sure why this is not failing the CI fmt check...

runtime/calamari/src/lib.rs
Garandor: I've been unhappy with the wording too.
I'm currently leaning towards calling it `EvictionPercentile` and `EvictionThreshold`
Tradeoff is that it's not immediately clear what number format these expect

Garandor:
Kicking config is now updateable and disabled by default on genesis.

The verify step of the benchmark is still failing (https://github.com/Manta-Network/Manta/runs/5211310990?check_suite_focus=true#step:9:1687), looking into it, but the code should now be fairly stable.

@ghzlatarev @bhgomes

pallets/collator-selection/src/lib.rs
Garandor: Is resolved, new setters are here: https://github.com/Manta-Network/Manta/pull/358/files#diff-a0359a0add814d66aac74493f7e17aefdcee21d3de1ae592e97059082b6feedfR374

runtime/calamari/Cargo.toml
Garandor: is resolved

runtime/dolphin/src/lib.rs
Garandor: removed those config params, resolved

runtime/dolphin/Cargo.toml
Garandor: resolved

runtime/calamari/src/lib.rs
Garandor: removed config params, resolved

runtime/calamari/src/weights/manta_collator_selection.rs
```diff
@@ -74,6 +76,13 @@ impl<T: frame_system::Config> manta_collator_selection::WeightInfo for Substrate
 		(10_523_000 as Weight)
 			.saturating_add(T::DbWeight::get().writes(1 as Weight))
 	}
+
+	fn set_eviction_percentile() -> Weight{
+		(0 as Weight)
```
ghzlatarev: These 0 weights are yet to be benchmarked, right ?

runtime/calamari/src/lib.rs
ghzlatarev: Not a big deal I guess, it becomes clear if the person reads the algorithm.

runtime/calamari/src/weights/manta_collator_selection.rs
Garandor: Yeah, is was trying to figure out how to run benchmark locally, but - even if I did - my machine is not up to spec.
Once CI succeeds i hope I can just use the CI benchmark artifact

runtime/calamari/src/lib.rs
Garandor: Modified as above in the latest commit

runtime/calamari/src/weights/manta_collator_selection.rs
ghzlatarev: Cool

ghzlatarev: @Garandor please search for `Search for Generating weight files with CI workflows` in the new benchmarking doc in Notion, to see how to use the CI for benchmarking.

Garandor: Added CI-computed weights.
Most extrinisics are cheaper at the tradeoff that the note-author hook and new_session access more storage items.
new_session might be cheaper for large N of collators because the map is only read once ( weights file shows twice for some reason I don't see ) instead of the old solution that had N reads.

Garandor:
Do Not Merge until resolution of

https://substrate.stackexchange.com/questions/504/reclaim-used-space-of-a-removed-storageitem-on-a-live-chain

Resolution came quick - A migration to kill the old storage item is needed

Garandor:
@stechu Worked in your comments. 
As for adding #transactional, I think the two set_* extrinsics really don't have the complexity to warrant adding it.
Parity also hasnt marked set_candidacy_bond transactional: https://paritytech.github.io/cumulus/src/pallet_collator_selection/lib.rs.html#335 

This stil is not ready to merge as it needs a storage migration to clear the removed storage item

pallets/collator-selection/src/lib.rs
```diff
@@ -500,29 +554,87 @@ pub mod pallet {
...
+				return Vec::new(); // tolerating 100% underperformance disables kicking
+			}
+			let mut collator_perf_this_session =
+				<BlocksPerCollatorThisSession<T>>::iter().collect::<Vec<_>>();
+			if collator_perf_this_session.is_empty() {
+				return Vec::new(); // no validator performance recorded ( should not happen )
+			}
+
+			// 1. Ascending sort of collator performance list by number of produced blocks
+			collator_perf_this_session.sort_unstable_by_key(|k| k.1);
```
bhgomes, line 584: Is unstable sort correct here? Is there any loss of fairness by doing this? It's going to depend on the ordering that comes out of `BlocksPerCollatorThisSession::iter`.

bhgomes (approved):

pallets/collator-selection/src/lib.rs
Garandor: We apply a tolerance and kick _below_ it (even if it is 0%) due to this line:
https://github.com/Manta-Network/Manta/pull/358/files#diff-a0359a0add814d66aac74493f7e17aefdcee21d3de1ae592e97059082b6feedfR610

To my knowledge the only effect of unstable sorting is that a collators that produced equally many blocks as the one selected for the percentile will be randomly sorted above or below it, and are thus randomly added to the kick_candidates set and checked for underperformance.
As the checked nodes have the same created blocks count as the one at percentile, they will never be removed, so there is no impact to fairness in my view.

bhgomes: Ok great, just checking.

ghzlatarev (approved):

Garandor:
Tentatively final version of this PR:

- Calamari is set in Cargo.toml to use the last pre-merge version of manta_collator_selection on manta branch.
No migration is needed as it continues to use the old KickThreshold-based eviction mechamism.

NOTE: This will go well for a while, until we upgrade substrate dependencies to v0.9.17, as this hard-set rev of manta_collator_selection can't easily have its deps updated along with the rest of the code.
When we _do_ decide to upgrade calamari to the new mechanism, a migration is needed to remove the following storage items from the chain.

1. LastAuthoredBlock
2. KickThreshold

- Dolphin is set to use the pallet from local filesystem
No migration is needed as the chain/testnet has not yet been launched

pallets/collator-selection/src/lib.rs
Garandor: Resolving as no-action as this is a strictly optional change

Garandor:
~updated weights from CI benchmark missing, don't merge yet~

Garandor:
~Do not merge until resoution of #437~

~After #437  is on manta:~
- [x] ~Retarget collator selection to the merge commit of #437  in https://github.com/Manta-Network/Manta/pull/358/files#diff-c27ffa7e0d850a31b480f6aefffae5ad060b4022c5c9bc2cc1e2550bb7be2082R85~ (not upgrading as waiting for v0.9.18)
- [x] Retest on `polkadot-launch`
- [x] Merge

Garandor (commented):
Finished refactoring the test code @stechu 
It should be easier to follow along now.

LGTM

stechu (approved):

PR #359, opened 2022-01-21
Update upstream to v0.9.15

### LINKED ISSUE #351: Update substrate dependencies to `v9.1.5`
Recommended by W3F.

Dengjianping:
## Description

Update upstream to v0.9.15

closes: #351 

### Calamari/Manta Runtime
Hook order execution is now more explicit [hook change](https://github.com/paritytech/substrate/pull/10043), The pr tells more details about the change.
For now we will use `AllPalletsReversedWithSystemFirst` which will preserve old behaviour and we don't have to reorder our pallets in consstruct_runtime!
At some point we should migrate to [AllPalletsWithSystem](https://github.com/Manta-Network/Manta/pull/359/files#diff-56417537fb560c5d1d8018408551510b0434860ead99a3ff28512ebbb4f669ddR918) - https://github.com/Manta-Network/Manta/issues/378

### Calamari/Manta Client
None.

### Upstream Changes
Polkadot: 
* https://github.com/paritytech/polkadot/pull/4655
* https://github.com/paritytech/polkadot/pull/4656
* https://github.com/paritytech/substrate/pull/10043

https://github.com/paritytech/polkadot/compare/release-v0.9.13...release-v0.9.15

Substrate:
* https://github.com/paritytech/substrate/pull/10043
* https://github.com/paritytech/substrate/pull/10324

https://github.com/paritytech/substrate/compare/polkadot-v0.9.13...polkadot-v0.9.15

Cumulus: 
None.

https://github.com/paritytech/cumulus/compare/polkadot-v0.9.13...polkadot-v0.9.15

---

write a little note why.

- [ ] Wrote unit tests.

### DIFF
--- a/runtime/calamari/src/lib.rs
+++ b/runtime/calamari/src/lib.rs
@@ -114,7 +114,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_version: 3120,
 	impl_version: 1,
 	apis: RUNTIME_API_VERSIONS,
-	transaction_version: 4,
+	transaction_version: 3,
 };
 
 /// The version information used to identify this runtime when compiled natively.
@@ -859,11 +859,8 @@ construct_runtime!(
 		TechnicalMembership: pallet_membership::<Instance2>::{Pallet, Call, Storage, Event<T>, Config<T>} = 18,
 
 		// Collator support. the order of these 5 are important and shall not change.
-		// Authorship and Session use CollatorSelection functions in their on_initialize hooks.
-		// So CollatorSelection's on_initialize hook has to be executed first.
-		// Since we are using AllPalletsWithSystem the declared order should be respected.
-		CollatorSelection: manta_collator_selection::{Pallet, Call, Storage, Event<T>, Config<T>} = 20,
-		Authorship: pallet_authorship::{Pallet, Call, Storage} = 21,
+		Authorship: pallet_authorship::{Pallet, Call, Storage} = 20,
+		CollatorSelection: manta_collator_selection::{Pallet, Call, Storage, Event<T>, Config<T>} = 21,
 		Session: pallet_session::{Pallet, Call, Storage, Event, Config<T>} = 22,
 		Aura: pallet_aura::{Pallet, Storage, Config<T>} = 23,
 		AuraExt: cumulus_pallet_aura_ext::{Pallet, Storage, Config} = 24,
--- a/runtime/calamari/src/lib.rs
+++ b/runtime/calamari/src/lib.rs
@@ -916,7 +916,6 @@ pub type Executive = frame_executive::Executive<
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllPalletsWithSystem,
-	()
 >;
 
 impl_runtime_apis! {
--- a/runtime/calamari/src/lib.rs
+++ b/runtime/calamari/src/lib.rs
@@ -859,7 +859,7 @@ construct_runtime!(
 		TechnicalMembership: pallet_membership::<Instance2>::{Pallet, Call, Storage, Event<T>, Config<T>} = 18,
 
 		// Collator support. the order of these 5 are important and shall not change.
-		// Authorship and Session uses CollatorSelection functions in their on_initialize hooks.
+		// Authorship and Session use CollatorSelection functions in their on_initialize hooks.
 		// So CollatorSelection's on_initialize hook has to be executed first.
 		// Since we are using AllPalletsWithSystem the declared order should be respected.
 		CollatorSelection: manta_collator_selection::{Pallet, Call, Storage, Event<T>, Config<T>} = 20,
--- a/runtime/calamari/src/lib.rs
+++ b/runtime/calamari/src/lib.rs
@@ -859,7 +859,6 @@ construct_runtime!(
 		TechnicalMembership: pallet_membership::<Instance2>::{Pallet, Call, Storage, Event<T>, Config<T>} = 18,
 
 		// Collator support. the order of these 5 are important and shall not change.
-				// Collator support. the order of these 5 are important and shall not change.
 		// Authorship and Session uses CollatorSelection functions in their on_initialize hooks.
 		// So CollatorSelection's on_initialize hook has to be executed first.
 		// Since we are using AllPalletsWithSystem the declared order should be respected.
--- a/runtime/calamari/src/lib.rs
+++ b/runtime/calamari/src/lib.rs
@@ -114,7 +114,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_version: 3120,
 	impl_version: 1,
 	apis: RUNTIME_API_VERSIONS,
-	transaction_version: 3,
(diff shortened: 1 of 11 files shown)


runtime/calamari/src/lib.rs
```diff
@@ -915,7 +915,8 @@ pub type Executive = frame_executive::Executive<
 	Block,
 	frame_system::ChainContext<Runtime>,
 	Runtime,
-	AllPallets,
+	AllPalletsWithSystem,
+	()
```
stechu: Why `()` here?

runtime/manta/src/lib.rs
```diff
@@ -643,7 +643,8 @@ pub type Executive = frame_executive::Executive<
 	Block,
 	frame_system::ChainContext<Runtime>,
 	Runtime,
-	AllPallets,
+	AllPalletsWithSystem,
+	()
```
stechu: same.

runtime/calamari/src/lib.rs
Dengjianping: I will remove it. I think you still remember this hooker `OnRuntimeUpgrade`. It's empty by default `OnRuntimeUpgrade = ()`.

runtime/calamari/src/lib.rs
```diff
@@ -915,7 +915,7 @@ pub type Executive = frame_executive::Executive<
 	Block,
 	frame_system::ChainContext<Runtime>,
 	Runtime,
-	AllPallets,
+	AllPalletsWithSystem,
```
ghzlatarev: This basically reverses the execution of the hooks in construct_runtime! hook. Are you sure this is fine ?

Dengjianping: I followed acala's and polkadot's change, both are the same. And I did runtime upgrade, so it looks fine.
But I think we should be careful while we customize these hooks. Because the execution of order is fixed.

ghzlatarev (approved):

runtime/calamari/src/lib.rs
Garandor: Mind explaining why you resolved this? 

According to https://github.com/paritytech/substrate/pull/10043
> Though it is recommended to be careful about the order of definition of the pallet in construct_runtime. and probably reorganize and use AllPalletsWithSystem, so the order is the expected normal order and not in reversed (as before this PR).

We would also need to reverse the order of at least the session pallets block below to keep the hook order consistent
https://github.com/Manta-Network/Manta/blob/c4859d124453c52b27ef7144edb1b25eccc47059/runtime/calamari/src/lib.rs#L861

Dengjianping: @stechu @ghzlatarev What do you think about this change?
Originally, it was
`
Before this PR AllPallets represented all pallets without system in reversed order
`

I did runtime upgrade, it looks fine.

Dengjianping: As @Garandor said, we should use `AllPalletsWithoutSystemReversed` if we keep hooks order consistent.

Dengjianping: @Garandor 
`// Collator support. the order of these 5 are important and shall not change. `
Generally, the module index is fixed, we don't change it once it's set. Pallet order is different from hook order.

ghzlatarev: I think it's safe to use `AllPalletsWithSystem` because we don't have any logic that relies on the order of the hooks staying the same as before this PR.
And using the same order as the pallets are defined is definitely saner option - https://github.com/paritytech/substrate/pull/10043

ghzlatarev: If a pallet A has a function which is used in pallet B's  `on_initialize` hook, then pallet A's `on_initialize` hook needs to be executed first. According to a problem the Kilt guys ran into and presented at Sub0 - https://youtu.be/uKwEBE77JEI?t=1285

ghzlatarev: This needs to be looked into some more.

runtime/calamari/src/lib.rs
```diff
@@ -755,7 +755,7 @@ impl cumulus_pallet_dmp_queue::Config for Runtime {
 
 parameter_types! {
 	// Rotate collator's spot each 6 hours.
-	pub const Period: u32 = 6 * HOURS;
+	pub const Period: u32 = 2 * MINUTES;
```
Dengjianping: It should be 6 hours, right?

ghzlatarev: Yeah, this is forgotten from testing. Will fix.

stechu (approved):
LGTM.

stechu:
@Garandor  Can you take a look whether the concern you raised is addressed? It looks to be addressed to me.

Garandor:
My main concern is that using `AllPalletsWithSystemFirst` would in effect duplicate System, as it is *also* explicitly listed in  https://github.com/Manta-Network/Manta/blob/manta/runtime/manta/src/lib.rs#L583

If system is running any hooks, its invocation could be duplicated - e.g. logs could be written twice if executed on hook.
I haven't looked into hooks enough to determine if this concern is valid or not and I will look into this more, but if any of you know this isn't a problem I would defer to your opinion.

The rest LGTM

ghzlatarev:
> My main concern is that using `AllPalletsWithSystemFirst` would in effect duplicate System, as it is _also_ explicitly listed in https://github.com/Manta-Network/Manta/blob/manta/runtime/manta/src/lib.rs#L583
> 
> If system is running any hooks, its invocation could be duplicated - e.g. logs could be written twice if executed on hook. I haven't looked into hooks enough to determine if this concern is valid or not and I will look into this more, but if any of you know this isn't a problem I would defer to your opinion.
> 
> The rest LGTM

What you are looking for are the diffs in [frame/executive/src/lib.rs](https://github.com/paritytech/substrate/pull/10043/files?diff=split&w=0#diff-71647e596916530d84fd02546feb833ff703dbf19397d076c499ab7dbf265da2) and [frame/support/procedural/src/construct_runtime/mod.rs](https://github.com/paritytech/substrate/pull/10043/files?diff=split&w=0#diff-901d5a6348e65983f0d7bf165b4d30681aabe9c8429242cf63f92dc784c0d5fc)
You can see how `construct_runtime` is expanded:
```
	let all_pallets_with_system_reversed = names
		.iter()
		.fold(TokenStream2::default(), |combined, name| quote!((#name, #combined)));
```
and then how it is used in Executive:
```
                       let on_initialize_weight =
				<AllPalletsWithSystem as OnInitialize<u64>>::on_initialize(block_number);
```
When before it was actually referring to system pallet explicitly:
```
			let frame_system_on_initialize_weight =
				frame_system::Pallet::<Runtime>::on_initialize(block_number);
			let on_initialize_weight =
				<AllPallets as OnInitialize<u64>>::on_initialize(block_number);
```

This is why there is no double execution. System is no longer called explicitly. but as part of the set from `construct_runtime!`.

Garandor:
Ok got it, since the behavior of executive changed to not explicitly execute System, it is not a concern.

Thanks @ghzlatarev

Garandor (approved):

PR #373, opened 2022-01-29
Expose more ports in dockerfile

### LINKED ISSUE #372: Expose more ports in docker container
9615 for prometheus.
30334 for relachain port(libp2p).

Dengjianping:
## Description
30334 is a port for connecting relaychain by default, I think we should expose this port in case there's any connection issue between relaychain and parachain.

closes: #372 

---

write a little note why.

- [ ] Wrote unit tests.

### DIFF
--- a/docker/calamari.Dockerfile
+++ b/docker/calamari.Dockerfile
@@ -32,8 +32,8 @@ RUN $PARA_BINARY_PATH --version
 ADD $PARA_GENESIS_URL $PARA_GENESIS_PATH
 ADD $RELAY_GENESIS_URL $RELAY_GENESIS_PATH
 
-# Expose three ports by default
-EXPOSE 30333 9933 9944
+# Expose 5 ports by default
+EXPOSE 30333 30334 9933 9944 9615
 
 ENTRYPOINT [\
   "/usr/local/bin/manta",\


ghzlatarev (changes requested):
Please say why we're doing this change in the description.
We need to be able to easily realize what's happening in this PR if we open it after 6 months.

Dengjianping:
> Please say why we're doing this change in the description. We need to be able to easily realize what's happening in this PR if we open it after 6 months.

Updated.

ghzlatarev (approved):

Garandor:
> ## Description
> 30334 is a port for connecting relaychain by default, I think we should expose this port in case there's any connection issue between relaychain and parachain.

As relay- and parachain both run on the same device ( or in case of docker more specifically: in the same container ) since they are the same binary, they will be able to find each other even without exposed ports. Also I'm not sure why the parachain should talk on the relaychain p2p port, i would imagine they exchange messages on websocket or rpc.

The problem begins when external nodes want to reach the relaychain node on 30334 which won't be available unless exposed. See also the discussion in https://github.com/Manta-Network/docs/pull/15/files#r800349071

docker/calamari.Dockerfile
```diff
@@ -32,8 +32,8 @@ RUN $PARA_BINARY_PATH --version
 ADD $PARA_GENESIS_URL $PARA_GENESIS_PATH
 ADD $RELAY_GENESIS_URL $RELAY_GENESIS_PATH
 
-# Expose three ports by default
-EXPOSE 30333 9933 9944
+# Expose 5 ports by default
+EXPOSE 30333 30334 9933 9944 9615
```
Garandor, line 36: are you sure the ws/rpc ports should be exposed by default?
I'm thinking interacting with them in this way is a dev usecase and devs can just invoke it with
`docker run -p 9933:9933 imagename`

Garandor, line 36: Also, I'm not sure hardcoding the expose in the dockerfile is the way to go, this way it won't be possible to run two containers on the same machine in parallel because the first one will claim the host ports.

I would recommend not having an expose section in the Dockerfile and instead ask the user to invoke 
`docker run -p 30333:30333 -p 30333:30334 <+other needed ports> imagename`

That way the second container can be invoked with 
`docker run -p 30333:30343 -p 30333:30344 <+other needed ports> imagename`
and peering will still work.

Garandor: turns out i misunderstood the meaning of EXPOSE

> Exposing ports is a way of documenting which ports are used, but does not actually map or open any ports. Exposing ports is optional.
](https://docs.docker.com/engine/reference/builder/#expose)

resolving this

Garandor: > Exposing ports is a way of documenting which ports are used, but does not actually map or open any ports. Exposing ports is optional.
](https://docs.docker.com/engine/reference/builder/#expose)

resolving this

Dengjianping: Yeah, you still have explicitly to open the port while run a container even though the port is exposed in dockerfile.

Dengjianping:
> > ## Description
> > 30334 is a port for connecting relaychain by default, I think we should expose this port in case there's any connection issue between relaychain and parachain.
> 
> As relay- and parachain both run on the same device ( or in case of docker more specifically: in the same container ) since they are the same binary, they will be able to find each other even without exposed ports. Also I'm not sure why the parachain should talk on the relaychain p2p port, i would imagine they exchange messages on websocket or rpc.
> 
> The problem begins when external nodes want to reach the relaychain node on 30334 which won't be available unless exposed. See also the discussion in https://github.com/Manta-Network/docs/pull/15/files#r800349071

I think ws and rpc are for applications, but exchanging messages between parachain and relaychain is by p2p(libp2p).
This p2p port is open by default, https://github.com/Manta-Network/Manta/blob/manta/node/src/command.rs#L395

Parachain should have two peer sets, one saves parachain peer addresses, another one has relaychain peer addresses.
I don't know, just guess

Garandor:
I'm still digging around the code, but I'm positive they don't talk p2p or RPC or WS for that matter.
They are coupled tightly in code by the cumulus pallet -> https://github.com/paritytech/cumulus

There seems to be work to split the relaychain node out from the parachain in https://github.com/paritytech/cumulus/issues/545
and the relaychain has a trait interface already merged in https://github.com/paritytech/cumulus/pull/835 
but it looks like having this running in separate processes and communicating via IPC is a work in progress: https://github.com/paritytech/cumulus/pull/963

stechu (approved):

stechu:
@Garandor  do we have any conclusion on the port issue?

stechu (approved):

PR #382, opened 2022-02-05
[Manta] Asset-Manager and XCM

### LINKED ISSUE #297: Manta Token/XCM Trait Design
We would like to design a unified token layer for Manta/Calamari/Dolphin, with the following goals:

- [x] a unified token abstraction that can handle different kinds of relay chain / parachain assets
- [x] handle XCM routing/fee mechanism elegantly
- [x] work with Manta's XCM facility library

References:
1. Moonbeam's XCM Transactor: https://github.com/PureStake/moonbeam/tree/master/pallets/xcm-transactor
2. Moonbeam's XToken: https://github.com/PureStake/moonbeam/tree/master/precompiles/xtokens
3. Acala's XToken: https://github.com/open-web3-stack/open-runtime-module-library/tree/1f520348f31b5e94b8a5dd7f8e6b8ec359df4177/xtokens 
4. Acala's Token Traits: https://github.com/open-web3-stack/open-runtime-module-library/tree/1f520348f31b5e94b8a5dd7f8e6b8ec359df4177/tokens

stechu:
## Description

closes: #297 

Manta Asset-Manager and XCM code.

**Manta/Calamari Asset Design:**

1. Each asset has a unique identifier within Manta/Calamari, which is a `u32` typed `AssetId`, this include both XCM assets, Manta/Calamari’s native token, and customized assets in Manta/Calamari (in the future)
2. `Pallet-Asset-Manager` is the centralized interface for asset management, it supports:
    1. Create a new asset
    2. Update the metadata of current asset 
    3. Freeze/Unfreeze an asset (Native asset cannot be frozen)
    4. Change `AssetLocation` of an existing asset
    5. Change `UnitPerSecond` of an existing asset

**Manta/Calamari XCM Fee Payment Mechanism:**

1. When transferring Manta/Calamari native asset, we charged a fixed rate fee on KMA/MANTA. This is defined on the first element of `Trader` of `XcmConfig` (`FixedRateOfFungible`)
2. When transferring other asset, we always charge the first kind of assets (when calling `buy_weight(weight, payment)` , the first asset in `payment`). We maintain a `UnitPerSecond` rate for each asset we support in `pallet-asset-manager`. During a single XCM execution, we expect that the user should use the same kind of assets to buy weight.  If a user use the different kind of asset to buy weights during a single XCM execution, we only handling the refund of the first kind of asset being used to buy weights. (see `FirstAssetTrader` code)

---

write a little note why.

- [x] Wrote unit tests.


runtime/calamari/tests/xcm_tests.rs
```diff
@@ -0,0 +1,1791 @@
...
+			parachain::Origin::signed(ALICE.into()),
+			parachain::CurrencyId::MantaCurrency(a_currency_id),
+			amount,
+			1,
+			Box::new(VersionedMultiLocation::V1(dest)),
+			dest_weight,
+		));
+		assert_eq!(
+			parachain::Balances::free_balance(&ALICE.into()),
+			INITIAL_BALANCE - amount - fee
```
ghzlatarev: Can this `amount` get refunded if the fee is not enough on the receiving chain ?

runtime/calamari/tests/xcm_tests.rs
```diff
@@ -0,0 +1,1791 @@
...
+			Box::new((Here, amount).into()),
+			0,
+		));
+		assert_eq!(
+			relay_chain::Balances::free_balance(&para_account_id(1)),
+			INITIAL_BALANCE + amount
+		);
+	});
+
+	// parachain should not have received assets
```
ghzlatarev: Why is the transfer supposed to fail here actually ?

runtime/calamari/tests/xcm_tests.rs
```diff
@@ -0,0 +1,1791 @@
...
+	ParaA::execute_with(|| {
+		println!("fresh account bal: {}", parachain::Assets::balance(relay_asset_id, &new_account.into()));
+	});
+
+}
+
+#[test]
+fn receive_asset_with_is_sufficient_true(){
```
ghzlatarev: Isn't this pretty much the same as `reserve_transfer_relaychain_to_parachain_a`

pallets/asset-manager/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+				Ok(id)
+			})
+		}
+
+		/// The account ID of AssetManager
+		pub fn account_id() -> T::AccountId {
+			PALLET_ID.into_account()
+		}
+	}
+}
```
ghzlatarev, line 374: The formatting in this page is kind of off. Also in other pages but less so.

runtime/primitives/src/xcm.rs
```diff
@@ -0,0 +1,233 @@
...
+					debug_assert!(ok, "`mint_into` cannot generally fail; qed");
+				}
+			}
+			Err(_) => log::debug!(
+				target: "xcm",
+				"take revenue failed matching fungible"
+			),
+		}
+	}
+}
```
ghzlatarev, line 244: The formatting in this page is kind of off. Also in other pages but less so.

runtime/calamari/tests/xcm_mock/mod.rs
```diff
@@ -0,0 +1,121 @@
+// Copyright 2021 Parity Technologies (UK) Ltd.
+// This file is part of Polkadot.
+
+// Polkadot is free software: you can redistribute it and/or modify
```
ghzlatarev: Some of the footers are still referring to Polkadot and Parity Technologies.

runtime/calamari/tests/xcm_tests.rs
stechu: No, the amount will stuck (by the current XCM design).

runtime/calamari/tests/xcm_tests.rs
stechu: When receiving this assets, `ParaA` need to create a new account for `ALICE`. Due to `is_sufficient` set to `false`, parachain cannot do so. 

To the first approximation, you can think of `is_sufficient` means whether this asset can be treated as reserve asset. The logic here is, despite the receiving asset is greater than the `min_balance` of this particular asset. Since this receiving asset is not treated as reserve asset and the`ALICE` doesn't have any other kind of balances (thus the account is not `active`), this transaction should fail. 

This failure also DOES NOT trigger a refund so need to be take care of cautiously.

runtime/calamari/tests/xcm_tests.rs
stechu: yes, this is just a dual test to previous test to make sure under the exact same condition the test passes when `is_sufficient` is `true`.

runtime/primitives/src/assets.rs
```diff
@@ -0,0 +1,108 @@
...
+// along with Manta.  If not, see <http://www.gnu.org/licenses/>.
+
+use codec::{Decode, Encode};
+use scale_info::TypeInfo;
+use sp_std::{borrow::Borrow, marker::PhantomData};
+///! Manta/Calamari/Dolphin Asset
+use xcm::{
+	v1::{Junctions, MultiLocation},
+	VersionedMultiLocation,
+};
```
bhgomes: nit: formatting

pallets/asset-manager/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+		) -> DispatchResult;
+	}
+
+	/// The AssetMetadata trait:
+	pub trait AssetMetadata<T: Config> {
+		/// Returns the minimum balance to hold this asset
+		fn min_balance(&self) -> T::Balance;
+
+		/// Returns a boolean value indicating whether this asset needs an existential deposit
+		fn is_sufficient(&self) -> bool;
```
bhgomes, line 96: Are these both about existential deposits? If so then I think this is a better interface:

```rust
fn minimum_required_balance(&self) -> Option<T::Balance>;
```

where `None` means that no existential deposit is enforced. We could even drop the `Option` if `T::Balance: Default` in which case the default value of "zero" means exactly that no existential deposit is needed since zero is the minimum possible balance anyway.

runtime/primitives/src/assets.rs
```diff
@@ -0,0 +1,108 @@
...
+/// Note: This does not guaranttee the `AssetLocation` is registered (i.e. have an AssetId)
+impl From<MultiLocation> for AssetLocation {
+	fn from(location: MultiLocation) -> Self {
+		AssetLocation(VersionedMultiLocation::V1(location))
+	}
+}
+
+/// Convert an `AssetLocation` to a MultiLocation
+/// If Native, return none.
+impl Into<Option<MultiLocation>> for AssetLocation {
```
bhgomes: Implementing `From` is always prefered when possible, which should be possible here:

```rust
impl From<AssetLocation> for Option<MultiLocation> {
    fn from(location: AssetLocation) -> Self {
        match location {
            AssetLocation(VersionedMultiLocation::V1(location)) => Some(location),
            _ => None,
        }
    }
}
```

but semantically `TryFrom` is more correct here.

runtime/primitives/src/xcm.rs
```diff
@@ -0,0 +1,233 @@
...
+// UnitToWeightRatio trait.
+pub struct FirstAssetTrader<
+	AssetId: Clone,
+	AssetLocation: From<MultiLocation> + Clone,
+	AssetIdInfoGetter: UnitsToWeightRatio<AssetId> + AssetIdLocationGetter<AssetId, AssetLocation>,
+	R: TakeRevenue,
+>(	
+	Weight,
+	Option<(MultiLocation, u128, u128)>,
+	sp_std::marker::PhantomData<(AssetId, AssetLocation, AssetIdInfoGetter, R)>);
```
bhgomes: Can we use a `struct` with named fields here? This is a rather important implementation to get right.

runtime/primitives/src/xcm.rs
```diff
@@ -0,0 +1,233 @@
...
+		AccountId: Clone,
+		ReceiverAccount: Get<AccountId>,
+	> TakeRevenue for XcmFeesToAccount<Assets, Matcher, AccountId, ReceiverAccount>
+{
+	fn take_revenue(revenue: MultiAsset) {
+		match Matcher::matches_fungibles(&revenue) {
+			Ok((asset_id, amount)) => {
+				if !amount.is_zero() {
+					let ok = Assets::mint_into(asset_id, &ReceiverAccount::get(), amount).is_ok();
+					debug_assert!(ok, "`mint_into` cannot generally fail; qed");
```
bhgomes: More canonical to use an `expect` directly:

```rust
Assets::mint_into(asset_id, &ReceiverAccount::get(), amount).expect("`mint_into` cannot generally fail");
```

pallets/asset-manager/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+		/// # </weight>
+		#[pallet::weight(50_000_000)]
+		pub fn register_asset(
+			origin: OriginFor<T>,
+			location: T::AssetLocation,
+			metadata: T::AssetRegistrarMetadata,
+		) -> DispatchResult {
+			T::ModifierOrigin::ensure_origin(origin)?;
+			ensure!(
+				LocationAssetId::<T>::get(&location).is_none(),
```
bhgomes: We should use the same precondition for all the storage modifications which is `AssetIdLocation` in the other extrinsics. In this case, we should change it to:

```rust
ensure!(!AssetIdLocation::<T>::contains_key(&asset_id), Error::<T>::AssetAlreadyRegistered);
```

and the error should reflect the actual state we are conditioning on, i.e. `LocationAlreadyExists` -> `AssetAlreadyRegistered`.

On second thought, for future-proofing, we should probably change all the storage precondition matching to `AssetIdMetadata`, and have that be the primary map. The other two are "convenience" maps.

pallets/asset-manager/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+				)
+			}
+			// write to the ledger state.
+			if let Some(location) = location_option.clone() {
+				let old_location =
+					AssetIdLocation::<T>::get(&asset_id).ok_or(Error::<T>::UpdateNonExistAsset)?;
+				LocationAssetId::<T>::remove(&old_location);
+				LocationAssetId::<T>::insert(&location, &asset_id);
+				AssetIdLocation::<T>::insert(&asset_id, &location);
+			}
```
bhgomes: The bodies of these two `if` statements should be merged and `location_option` can be matched by reference instead of cloning.

pallets/asset-manager/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+		/// * `asset_id`: AssetId to be updated.
+		/// * `metadata_option`: `Some(meta)` to update the metadata to `meta`, `None` means no update on metadata.
+		/// * `location_option`: `Some(location)` to update the asset location, `None` means no update on location.
+		///
+		/// # <weight>
+		/// TODO: get actual weight
+		/// # </weight>
+		#[pallet::weight(50_000_000)]
+		#[transactional]
+		pub fn update_asset(
```
bhgomes: This extrinsic should be split into two depending on the arguments, since both `None` does nothing, and you pay the weight of both even if you only use one. Some API name ideas:

```rust
pub fn update_asset_location(origin: OriginFor<T>, asset_id: T::AssetId, location: T::AssetLocation);
pub fn update_asset_metadata(origin: OriginFor<T>, asset_id: T::AssetId, metadata: T::AssetRegistrarMetadata);
```

If there is a scenario where you want to edit both at the same time, we can add a third extrinsic. This is along the same vein as `set_units_per_second`, which could have been part of the metadata or `update_asset` extrinsic but was not, because it is usually modified for different reasons. I think metadata and location are also independent for the same reason.

runtime/calamari/tests/xcm_mock/parachain.rs
```diff
@@ -0,0 +1,660 @@
...
+	CurrencyAdapter as XcmCurrencyAdapter, EnsureXcmOrigin, FixedRateOfFungible, FixedWeightBounds,
+	FungiblesAdapter, IsConcrete, LocationInverter, ParentIsDefault, SiblingParachainAsNative,
+	SiblingParachainConvertsVia, SignedAccountId32AsNative, SignedToAccountId32,
+	SovereignSignedViaLocation,
+};
+use xcm_executor::{traits::JustTry, Config, XcmExecutor};
+
+pub type AccountId = AccountId32;
+pub type Balance = u128;
+pub type AssetId = u32; // Manta plans to use u32 as AssetId
```
bhgomes: It would be nice to consolidate all the common types among all the repos. We can think about it after testnetv2.

runtime/calamari/tests/xcm_mock/parachain.rs
```diff
@@ -0,0 +1,660 @@
...
+	pub enum Event<T: Config> {
+		// XCMP
+		/// Some XCM was executed OK.
+		Success(Option<T::Hash>),
+		/// Some XCM failed.
+		Fail(Option<T::Hash>, XcmError),
+		/// Bad XCM version used.
+		BadVersion(Option<T::Hash>),
+		/// Bad XCM format used.
+		BadFormat(Option<T::Hash>),
```
bhgomes: In what cases do we not know the hash?

runtime/calamari/tests/xcm_mock/parachain.rs
```diff
@@ -0,0 +1,660 @@
...
+				let mut data_ref = data;
+				let _ = XcmpMessageFormat::decode(&mut data_ref)
+					.expect("Simulator encodes with versioned xcm format; qed");
+
+				let mut remaining_fragments = &data_ref[..];
+				while !remaining_fragments.is_empty() {
+					if let Ok(xcm) = VersionedXcm::<T::Call>::decode(&mut remaining_fragments) {
+						let _ = Self::handle_xcmp_message(sender, sent_at, xcm, max_weight);
+					} else {
+						debug_assert!(false, "Invalid incoming XCMP message data");
```
bhgomes: This should just be a `panic!`, or without the `if`:

```rust
let xcm = VersionedXcm::<T::Call>::decode(&mut remaining_fragments).expect("Invalid incoming XCMP message data.");
let _ = Self::handle_xcmp_message(sender, send_at, xcm, max_weight);
```

Also, why is the `handle_xcmp_message` return value dropped?

runtime/calamari/tests/xcm_mock/parachain.rs
```diff
@@ -0,0 +1,660 @@
...
+					}
+					Ok(Err(())) => {
+						Self::deposit_event(Event::UnsupportedVersion(id));
+					}
+					Ok(Ok(x)) => {
+						let outcome = T::XcmExecutor::execute_xcm(Parent, x.clone(), limit);
+						<ReceivedDmp<T>>::append(x);
+						Self::deposit_event(Event::ExecutedDownward(id, outcome));
+					}
+				}
```
bhgomes: nit: removing the `;` after each `deposit_event` allows you to compress the vertical space, as the formatter removes 4 redundant lines

runtime/calamari/tests/xcm_mock/parachain.rs
```diff
@@ -0,0 +1,660 @@
...
+	pub evm_address: Option<H160>,
+	pub is_frozen: bool,
+	pub min_balance: Balance,
+	/// `is_sufficient`: Whether a non-zero balance of this asset is deposit of sufficient
+	/// value to account for the state bloat associated with its balance storage. If set to
+	/// `true`, then non-zero balances may be stored without a `consumer` reference (and thus
+	/// an ED in the Balances pallet or whatever else is used to control user-account state
+	/// growth).
+	/// For example, if is_sufficient set to `false`, a fresh account cannot receive XCM tokens.
+	pub is_sufficient: bool,
```
bhgomes: See the other comment about `min_balance` being enough information to enforce existential deposit.

runtime/primitives/src/assets.rs
```diff
@@ -0,0 +1,108 @@
...
+	}
+}
+
+/// Defines the trait to obtain a generic AssetId
+pub trait AssetIdLocationGetter<AssetId, AssetLocation> {
+	// get AssetLocation from AssetId
+	fn get_asset_location(asset_id: AssetId) -> Option<AssetLocation>;
+
+	// get AssetId from AssetLocation
+	fn get_asset_id(loc: AssetLocation) -> Option<AssetId>;
```
bhgomes: I think this can take location by reference. See the actual implementation of this trait which uses storage getter to access this data.

runtime/primitives/src/assets.rs
```diff
@@ -0,0 +1,108 @@
...
+	AssetId: Clone,
+	AssetLocation: From<MultiLocation> + Into<Option<MultiLocation>> + Clone,
+	AssetInfoGetter: AssetIdLocationGetter<AssetId, AssetLocation>,
+{
+	fn convert_ref(loc: impl Borrow<MultiLocation>) -> Result<AssetId, ()> {
+		if let Some(asset_id) = AssetInfoGetter::get_asset_id(loc.borrow().clone().into()) {
+			Ok(asset_id)
+		} else {
+			Err(())
+		}
```
bhgomes:
```rust
AssetInfoGetter::get_asset_id(loc.borrow().clone().into()).ok_or(())
```

runtime/primitives/src/assets.rs
```diff
@@ -0,0 +1,108 @@
...
+	fn reverse_ref(id: impl Borrow<AssetId>) -> Result<MultiLocation, ()> {
+		if let Some(asset_loc) = AssetInfoGetter::get_asset_location(id.borrow().clone()) {
+			if let Some(location) = asset_loc.into() {
+				Ok(location)
+			} else {
+				Err(())
+			}
+		} else {
+			Err(())
+		}
```
bhgomes:
```rust
AssetInfoGetter::get_asset_location(id.borrow().clone()).and_then(Into::into).ok_or(())
```

runtime/primitives/src/xcm.rs
```diff
@@ -0,0 +1,233 @@
...
+			let parents = location.parent_count();
+			match (parents, first_interior.clone()) {
+				(0, Some(Parachain(id))) => Some(MultiLocation::new(0, X1(Parachain(id.clone())))),
+				(1, Some(Parachain(id))) => Some(MultiLocation::new(1, X1(Parachain(id.clone())))),
+				(1, _) => Some(MultiLocation::parent()),
+				_ => None,
+			}
+		} else {
+			None
+		}
```
bhgomes: Are all these clones needed? I suspect at least the clones on lines 49 and 50 are not needed.

runtime/primitives/src/xcm.rs
```diff
@@ -0,0 +1,233 @@
...
+/// reserve is same with `origin`.
+pub struct MultiNativeAsset;
+impl FilterAssetLocation for MultiNativeAsset {
+	fn filter_asset_location(asset: &MultiAsset, origin: &MultiLocation) -> bool {
+		if let Some(ref reserve) = asset.reserve() {
+			if reserve == origin {
+				return true;
+			}
+		}
+		false
```
bhgomes:
```rust
asset.reserve().map(|r| r == origin).unwrap_or(false)
```

runtime/primitives/src/xcm.rs
```diff
@@ -0,0 +1,233 @@
...
+				// Due to the trait bound, we can only refund one asset.
+				if let Some(new_asset) = new_asset {
+					self.0 = self.0.saturating_add(weight);
+					self.1 = Some(new_asset);
+				};
+				return Ok(unused);
+			}
+			_ => Err(XcmError::TooExpensive)
+		}
+	}
```
bhgomes: Are all these clones needed?

runtime/primitives/src/xcm.rs
```diff
@@ -0,0 +1,233 @@
...
+				units_per_second,
+			));
+			Some(MultiAsset {
+				fun: Fungibility::Fungible(amount),
+				id: xcmAssetId::Concrete(id.clone()),
+			})
+		} else {
+			None
+		} 
+	}
```
bhgomes: Are all of these clones needed?

runtime/calamari/tests/xcm_tests.rs
```diff
@@ -0,0 +1,1791 @@
...
+//  1. ReserveAssetDeposited(assets.clone()),
+//  2. ClearOrigin,
+//  3. BuyExecution { fees, weight_limit: Limited(0) },
+//  4. DepositAsset { assets: Wild(All), max_assets, beneficiary },
+//  each instruction's weight is 1000, thus, the total weight is 4000
+const RESERVE_TRANSFER_WEIGHT: u64 = 4000;
+
+fn calculate_fee(units_per_seconds: u128, weight: u64) -> u128 {
+	units_per_seconds * (weight as u128) / (WEIGHT_PER_SECOND as u128)
+}
```
bhgomes: We should make this a library function, since it is also used by [runtime/primitives/src/xcm.rs](https://github.com/Manta-Network/Manta/pull/382/files?diff=unified&w=0#diff-3366367f9696d8f74ba385cd02fe3c1e4b292086096acb4894da9cc12ce1fa3d).

runtime/calamari/tests/xcm_tests.rs
```diff
@@ -0,0 +1,1791 @@
...
+		));
+		assert_ok!(AssetManager::set_units_per_second(
+			parachain::Origin::root(), 
+			a_currency_id, 
+			0u128));
+		assert_eq!(
+			Some(a_currency_id),
+			parachain::AssetManager::location_asset_id(source_location.clone())
+		);
+	});
```
bhgomes: There are many similar blocks like this, can we refactor this to a function to make the tests shorter to read?

bhgomes (changes requested):

pallets/asset-manager/src/lib.rs
stechu: Actually no, the comments here is inaccurate (will fix that). `is_sufficient` (to the first approximation) means whether this asset can be used as reserve asset.

pallets/asset-manager/src/lib.rs
stechu: I deliberately did this here to have a clear separation between precondition checking and writing to the ledger.

runtime/calamari/tests/xcm_mock/parachain.rs
stechu: I should use the values defined in `manta-primitives` here instead actually.

runtime/calamari/tests/xcm_mock/parachain.rs
stechu: Honestly this code is copied from Polkadot. Not entirely sure here.

runtime/calamari/tests/xcm_mock/parachain.rs
stechu: Yeah, if you read the comments carefully. `min_balance` is not enough. I answered the question in your previous comments.

pallets/asset-manager/src/lib.rs
bhgomes: Yeah, but the `if` matches don't have to do with the ledger precondition, they are just parsing the input to see if it's a `Some` variant.

pallets/asset-manager/Cargo.toml
```diff
@@ -0,0 +1,42 @@
+[package]
+authors    = ['Manta Network']
+name = "pallet-asset-manager"
+version = "3.1.0"
```
Dengjianping: Currently, it's `3.1.3` in our default branch.

pallets/asset-manager/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+codec = { package = "parity-scale-codec", version = "2.3.1", default-features = false }
+scale-info = { version = "1.0", default-features = false, features = ["derive"] }
+sp-runtime = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.15", default-features = false }
+frame-support = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.15", default-features = false }
+frame-system = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.15", default-features = false }
+sp-std = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.15", default-features = false }
+frame-benchmarking = { git = 'https://github.com/paritytech/substrate.git', branch = "polkadot-v0.9.15", default-features = false, optional = true }
+manta-primitives = { path = '../../runtime/primitives', default-features = false}
+
+[dev-dependencies]
```
Dengjianping, line 22: Remove this section, because all tests are moved to calamari runtime.

pallets/asset-manager/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+	#[pallet::pallet]
+	pub struct Pallet<T>(PhantomData<T>);
+
+	/// The AssetManagers's pallet id
+	pub const PALLET_ID: PalletId = PalletId(*b"asstmngr");
+
+	/// The registrar trait: defines the interface of creating an asset in the asset implementation layer.
+	/// We may revisit this interface design (e.g. add change asset interface). However, change StorageMetadata
+	/// should be rare.
+	pub trait AssetRegistrar<T: Config> {
```
Dengjianping, line 61: Can we move the definition of both traits to a place like `primitives/traits.rs`? Just a suggestion.
It doesn't 'reach' the goal of decoupling pallets. 
For example, if another pallet wants to access to `asset-manager`, but which still need to be imported.
But both traits are inside `manta-primitives`, wee only import primitives pallet.

pallets/asset-manager/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+repository = 'https://github.com/Manta-Network/Manta/'
+
+
+[dependencies]
+codec = { package = "parity-scale-codec", version = "2.3.1", default-features = false }
+scale-info = { version = "1.0", default-features = false, features = ["derive"] }
+sp-runtime = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.15", default-features = false }
+frame-support = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.15", default-features = false }
+frame-system = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.15", default-features = false }
+sp-std = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.15", default-features = false }
```
Dengjianping: Seems this pallet is not under use.

pallets/asset-manager/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+		/// The asset id type, this have to be consistent with pallet-manta-pay
+		type AssetId: Member
+			+ Parameter
+			+ Default
+			+ Copy
+			+ HasCompact
+			+ CheckedAdd
+			+ Bounded
+			+ One
+			+ AtLeast32BitUnsigned
```
Dengjianping, line 127: I think you can remove these traits like `One`, `HasCompact `, `Bounded `, `CheckedAdd `.
You can find how these traits are derived: 
`One`, `HasCompact `, `Bounded `, `CheckedAdd ` => `BaseArithmetic` => `AtLeast32Bit` => `AtLeast32BitUnsigned`.
https://github.com/paritytech/substrate/blob/master/primitives/arithmetic/src/traits.rs#L156

pallets/asset-manager/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+			T::ModifierOrigin::ensure_origin(origin)?;
+			ensure!(
+				LocationAssetId::<T>::get(&location).is_none(),
+				Error::<T>::LocationAlreadyExists
+			);
+			let asset_id = Self::get_next_asset_id()?;
+			T::AssetRegistrar::create_asset(
+				asset_id,
+				metadata.min_balance(),
+				metadata.clone().into(),
```
Dengjianping, line 255: I kind of don't like `clone` here, is it possible to define this param as reference?

pallets/asset-manager/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+				metadata.is_sufficient(),
+			)
+			.map_err(|_| Error::<T>::ErrorCreatingAsset)?;
+			AssetIdLocation::<T>::insert(&asset_id, &location);
+			AssetIdMetadata::<T>::insert(&asset_id, &metadata.clone());
```
Dengjianping: is it possible to remove the `clone`?

pallets/asset-manager/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+			location_option: Option<T::AssetLocation>,
+			metadata_option: Option<T::AssetRegistrarMetadata>,
+		) -> DispatchResult {
+			// check validity.
+			T::ModifierOrigin::ensure_origin(origin)?;
+			ensure!(
+				AssetIdLocation::<T>::contains_key(&asset_id),
+				Error::<T>::UpdateNonExistAsset
+			);
+			if let Some(location) = location_option.clone() {
```
Dengjianping: This `if let` can be simplified as 
```rust
ensure!(
    location_option.as_ref().map(LocationAssetId::<T>::contains_key) == Some(true),
    Error::<T>::LocationAlreadyExists
);
```
Or remove this clone
```rust
if let Some(ref location) = location_option {}
```

pallets/asset-manager/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+			}
+			// write to the ledger state.
+			if let Some(location) = location_option.clone() {
+				let old_location =
+					AssetIdLocation::<T>::get(&asset_id).ok_or(Error::<T>::UpdateNonExistAsset)?;
+				LocationAssetId::<T>::remove(&old_location);
+				LocationAssetId::<T>::insert(&location, &asset_id);
+				AssetIdLocation::<T>::insert(&asset_id, &location);
+			}
+			if let Some(metadata) = metadata_option.clone() {
```
Dengjianping: Remove the clone.

pallets/asset-manager/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+		/// * `units_per_second`: units per second for `asset_id`
+		/// # <weight>
+		/// TODO: get actual weight
+		/// # </weight>
+		#[pallet::weight(50_000_000)]
+		#[transactional]
+		pub fn set_units_per_second(
+			origin: OriginFor<T>,
+			asset_id: T::AssetId,
+			units_per_second: u128,
```
Dengjianping: `#[pallet::compact]`

runtime/primitives/Cargo.toml
```diff
@@ -12,20 +12,28 @@ targets = ['x86_64-unknown-linux-gnu']
...
 
 [features]
 default = ["std"]
 std = [
 	'codec/std',
 	'sp-consensus-aura/std',
+	'scale-info/std',
 	'sp-io/std',
 	'sp-std/std',
+	'log/std',
```
Dengjianping, line 38: Add `'xcm-executor/std'` here.

runtime/primitives/src/assets.rs
```diff
@@ -0,0 +1,108 @@
+// Copyright 2020-2021 Manta Network.
```
Dengjianping: 2022

runtime/primitives/src/xcm.rs
```diff
@@ -0,0 +1,233 @@
+// Copyright 2020-2021 Manta Network.
```
Dengjianping: 2022

runtime/calamari/tests/xcm_mock/relay_chain.rs
```diff
@@ -0,0 +1,231 @@
+// Copyright 2021 Parity Technologies (UK) Ltd.
```
Dengjianping: @ghzlatarev mentioned it.

Dengjianping (changes requested):

runtime/calamari/tests/xcm_tests.rs
stechu: Not sure how to do that honestly. It need to be a closure to capture the context.

pallets/asset-manager/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+				Error::<T>::UpdateNonExistAsset
+			);
+			if let Some(location) = location_option.clone() {
+				ensure!(
+					!LocationAssetId::<T>::contains_key(&location),
+					Error::<T>::LocationAlreadyExists
+				)
+			}
+			// write to the ledger state.
+			if let Some(location) = location_option.clone() {
```
Dengjianping: clone

runtime/calamari/tests/xcm_tests.rs
ghzlatarev: This can be a good first issue for someone new actually. To understand, dedupe and comment this code more.

runtime/primitives/src/assets.rs
```diff
@@ -0,0 +1,108 @@
...
+impl From<MultiLocation> for AssetLocation {
+	fn from(location: MultiLocation) -> Self {
+		AssetLocation(VersionedMultiLocation::V1(location))
+	}
+}
+
+/// Convert an `AssetLocation` to a MultiLocation
+/// If Native, return none.
+impl Into<Option<MultiLocation>> for AssetLocation {
+	fn into(self: Self) -> Option<MultiLocation> {
```
Dengjianping:
```rust
fn into(self)
```

runtime/primitives/src/assets.rs
Dengjianping: `and_then` will return `Option<Option< MultiLocation>>`, `ok_or` returns a type like `Err(())` or `Ok(None) ` or `Ok(Some(_))`.
```rust
AssetInfoGetter::get_asset_location(id.borrow().clone()).and_then(Into::into).flatten().ok_or(())
```
or 
```rust
AssetInfoGetter::get_asset_location(id.borrow().clone()).map(Into::into).flatten().ok_or(())
```

runtime/primitives/src/assets.rs
stechu: It is not wrong to have type annotation.

runtime/primitives/src/assets.rs
stechu: without `flatten` seems works as well.

runtime/calamari/tests/xcm_tests.rs
bhgomes: I mean the blocks that do almost exactly this code. Something like this:

```rust
fn assert_valid_register<P>(source_location: ..., asset_metadata: ..., currency_id: ...) 
where
    P: ...
{
     P::execute_with(|| {
        assert_ok!(parachain::AssetManager::register_asset(
            parachain::Origin::root(),
            source_location.clone(),
            asset_metadata.clone(),
        ));
        assert_ok!(AssetManager::set_units_per_second(
            parachain::Origin::root(),
            currency_id,
            0
        ));
        assert_eq!(
            Some(currency_id),
            parachain::AssetManager::location_asset_id(source_location.clone())
        );
    })
}
```

where `...` fills in the type names.

runtime/primitives/src/xcm.rs
stechu: removed unnecessary ones.

runtime/primitives/src/xcm.rs
stechu: same here.

stechu:
Update: the code needs to be adapted to the breaking change of XCM re-anchoring logic brought by `v0.9.16` (https://github.com/paritytech/polkadot/pull/4470). The change seems to be non-trivial.

runtime/dolphin/tests/xcm_mock/relay_chain.rs
```diff
@@ -0,0 +1,233 @@
+// Copyright 2021 Parity Technologies (UK) Ltd.
```
Dengjianping: Please use our own header

runtime/primitives/src/xcm.rs
```diff
@@ -0,0 +1,243 @@
...
+pub trait Reserve {
+	/// Returns assets reserve location.
+	fn reserve(&self) -> Option<MultiLocation>;
+}
+
+// Takes the chain part of a MultiAsset
+impl Reserve for MultiAsset {
+	fn reserve(&self) -> Option<MultiLocation> {
+		// We only care about concrete location now.
+		if let xcmAssetId::Concrete(location) = self.id.clone() {
```
Dengjianping, line 49: Remove clone

runtime/primitives/src/xcm.rs
```diff
@@ -0,0 +1,243 @@
+// Copyright 2020-2022 Manta Network.
```
Dengjianping, line 1: Can we move these xcm related code to a separated crate like `xcm-support` or `?

stechu: Let's leave this to another PR.

bhgomes (approved):
Fine from me. Still not clear on the distinction between `min_balance` and `is_sufficient` and there are still some open comments by Jamie.

runtime/calamari/tests/xcm_tests.rs
stechu: This logic is not in our code. So I would like to get a stand alone code here.

runtime/calamari/tests/xcm_tests.rs
stechu: I think it worth-while to make the tests more abstract. But let's leave that to the next PR.

runtime/primitives/src/xcm.rs
stechu: This is necessary since `self.id` is a reference and it cannot be dereferenced here.

stechu:
> Fine from me. Still not clear on the distinction between `min_balance` and `is_sufficient` and there are still some open comments by Jamie.

The semantics is pretty subtle here, I added it to the new comments:

```rust
/// * `is_sufficient`: whether this asset can be used as reserve asset,
/// 	to the first approximation. More specifically, Whether a non-zero balance of this asset is deposit of sufficient
/// 	value to account for the state bloat associated with its balance storage. If set to
/// 	`true`, then non-zero balances may be stored without a `consumer` reference (and thus
/// 	an ED in the Balances pallet or whatever else is used to control user-account state
/// 	growth).
```

pallets/asset-manager/Cargo.toml
```diff
@@ -0,0 +1,51 @@
+[package]
+authors    = ['Manta Network']
+name = "pallet-asset-manager"
+version = "3.1.3"
```
Dengjianping: 3.1.4

pallets/asset-manager/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+edition = "2021"
+homepage   = 'https://manta.network'
+license    = 'GPL-3.0'
+repository = 'https://github.com/Manta-Network/Manta/'
+
+
+[dependencies]
+# codec has to be 2.0.0 for now
```
Dengjianping: Why 2.0? We're using 2.3.1 in other crates.

pallets/asset-manager/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+		#[pallet::weight(50_000_000)]
+		#[transactional]
+		pub fn register_asset(
+			origin: OriginFor<T>,
+			location: T::AssetLocation,
+			metadata: T::AssetRegistrarMetadata,
+		) -> DispatchResult {
+			T::ModifierOrigin::ensure_origin(origin)?;
+			ensure!(
+				LocationAssetId::<T>::get(&location).is_none(),
```
Dengjianping: Brandon‘s comment you didn't resolve.
`get` -> `contains_key`

Dengjianping:
Please resolve some clippy warnings.

pallets/asset-manager/Cargo.toml
stechu: I get compilation error before. But since now I upgrade to `v0.9.16`, will give it a try now.

stechu: Oh, I remember now. `2.3.1` works but `3.0.0` doesn't.

pallets/asset-manager/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+			// change the ledger state.
+			let old_location =
+				AssetIdLocation::<T>::get(&asset_id).ok_or(Error::<T>::UpdateNonExistAsset)?;
+			LocationAssetId::<T>::remove(&old_location);
+			LocationAssetId::<T>::insert(&location, &asset_id);
+			AssetIdLocation::<T>::insert(&asset_id, &location);
+			// deposit event.
+			Self::deposit_event(Event::<T>::AssetLocationUpdated {
+				asset_id,
+				location: location,
```
Dengjianping: `location: location` -> `location`

pallets/asset-manager/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+			let asset_id = Self::get_next_asset_id()?;
+			T::AssetRegistrar::create_asset(
+				asset_id,
+				metadata.min_balance(),
+				metadata.clone().into(),
+				metadata.is_sufficient(),
+			)
+			.map_err(|_| Error::<T>::ErrorCreatingAsset)?;
+			AssetIdLocation::<T>::insert(&asset_id, &location);
+			AssetIdMetadata::<T>::insert(&asset_id, &metadata.clone());
```
Dengjianping: Remove clone

ghzlatarev (approved):

pallets/asset-manager/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+			ensure!(
+				!LocationAssetId::<T>::contains_key(&location),
+				Error::<T>::LocationAlreadyExists
+			);
+			// change the ledger state.
+			let old_location =
+				AssetIdLocation::<T>::get(&asset_id).ok_or(Error::<T>::UpdateNonExistAsset)?;
+			LocationAssetId::<T>::remove(&old_location);
+			LocationAssetId::<T>::insert(&location, &asset_id);
+			AssetIdLocation::<T>::insert(&asset_id, &location);
```
Dengjianping, line 300: why insert new one instead of modifying it?
And if insert new one, you should remove previous one.

pallets/asset-manager/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+			origin: OriginFor<T>,
+			#[pallet::compact] asset_id: T::AssetId,
+			metadata: T::AssetRegistrarMetadata,
+		) -> DispatchResult {
+			T::ModifierOrigin::ensure_origin(origin)?;
+			ensure!(
+				AssetIdLocation::<T>::contains_key(&asset_id),
+				Error::<T>::UpdateNonExistAsset
+			);
+			AssetIdMetadata::<T>::insert(&asset_id, &metadata);
```
Dengjianping, line 323: The same as previous comment: why insert new one instead of modifying it?
And I think you should add some checks here, new metadata should be checked as well.

pallets/asset-manager/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+		///
+		/// * `origin`: Caller of this extrinsic, the acess control is specfied by `ForceOrigin`.
+		/// * `asset_id`: AssetId to be updated.
+		/// * `units_per_second`: units per second for `asset_id`
+		/// # <weight>
+		/// TODO: get actual weight
+		/// # </weight>
+		#[pallet::weight(50_000_000)]
+		#[transactional]
+		pub fn set_units_per_second(
```
Dengjianping, line 338: How do we modify `UnitsPerSecond`?
Should we add one more extrinsic here to modify it?

pallets/asset-manager/src/tests.rs
```diff
@@ -0,0 +1,184 @@
...
+		assert_ok!(AssetManager::register_asset(
+			Origin::root(),
+			source_location.clone(),
+			asset_metadata.clone()
+		));
+		assert_eq!(
+			AssetIdLocation::<Runtime>::get(counter),
+			Some(source_location.clone())
+		);
+		counter += 1;
```
Dengjianping, line 107: You should check whether asset id is increased.
And check `AssetIdMetadata`/`LocationAssetId`.

Should we check `min_balance` and the length of `name`/`symbol`?

pallets/asset-manager/src/tests.rs
```diff
@@ -0,0 +1,184 @@
...
+			Origin::root(),
+			source_location.clone(),
+			asset_metadata.clone()
+		));
+		assert_noop!(
+			AssetManager::update_asset_location(Origin::root(), 1, new_location),
+			crate::Error::<Runtime>::LocationAlreadyExists
+		);
+	})
+}
```
Dengjianping, line 191: Add one more case for `set_units_per_second` even its logic is really simple?

Dengjianping:
And do fmt.

pallets/asset-manager/src/lib.rs
stechu: You should also use this one to modify it.

pallets/asset-manager/src/tests.rs
stechu: I would also put this to future PR. Since currently, only governance can access asset-manager.

Dengjianping (approved):
LGTM!

PR #393, opened 2022-02-09
[manta] CI runtime upgrade test and github templates improvements

### LINKED ISSUE #300: Improve runtime upgrade CI test scenarios
* Right now our test scenario is old client + genesis runtime --> upgrade to new runtime
* Also have a test starting with the new client
* Also another runtime upgrade test on top of the first one

ghzlatarev:
## Description

closes: #300 

* Runtime upgrade test changes:
  - Before the CI runtime upgrade test scenario started the network with `old client` + `old runtime` upgrading the runtime only. 
  - This is problematic because sometimes there are breaking changes to the client introduced from Parity's updates, which automatically fail our test.
  - From now on the CI runtime upgrade test should just start with `new client`. We always upgrade our clients first before we do runtime upgrades on our production environment anyway.
* Removed some of the PR template bullets, that don't apply anymore or are more appropriate for the release template.

---

write a little note why.

- [x] Wrote unit tests.
- [x] If needed, bump `version` for every crate.
- [x] If we're going to issue a new release, freeze the code one week early(it depends, but usually it's one week), ensure we have enough time for related testing.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -18,7 +18,6 @@ Before we can merge this PR, please make sure that all the following items have
 checked off. If any of the checklist items are not applicable, please leave them but
 write a little note why.
 
-- [ ] Targeted PR against correct branch (`manta` or `dolphin`) with right title (start with [Manta] or [Dolphin]),
```
Garandor, line 21: > PR Title starts with [Manta] [Kalamari] [Dolphin]

I think this info is still valuable to see which RT the change impacts.
If no title tag is given, it must be a client change only - i.e. no RT is impacted

.github/ISSUE_TEMPLATE/release.md
```diff
@@ -24,18 +23,23 @@ candidate branch.
     runtime logic.
 - [ ] Check for any upstream storage migrations and perform tests with `try-runtime`, if any.
 
-The following checks can be performed after we have forked off to the release branch.
+The following checks can be performed after we have frozen our release candidate:
+
+- [ ] Code freeze should typically happen one week prior to rlease, to ensure we have enough time for related testing.
```
Garandor: rlease typo

.github/PULL_REQUEST_TEMPLATE.md
ghzlatarev: I thought we used this convention to simply specify the merging branch, not where the changes are.
What do you guys think ? @stechu @Dengjianping

.github/ISSUE_TEMPLATE/release.md
ghzlatarev: Thanks.

.github/PULL_REQUEST_TEMPLATE.md
stechu: I am all for drop it. Just for better ergonomics.

.github/workflows/publish-draft-releases.yml
```diff
@@ -496,17 +402,17 @@ jobs:
...
               peer-count:
                 relay: 5
                 para: 2
           -
             id: calamari-dev
             expected:
               block-count:
-                relay: 25
-                para: 6
+                relay: 20
```
stechu: Any reason for this change?

.github/workflows/publish-draft-releases.yml
```diff
@@ -175,98 +175,6 @@ jobs:
...
-        name: strip
-        run: |
-          strip target/release/manta
-      -
-        if: ${{ needs.check-for-runtime-upgrade.outputs.do-versions-match == 'false' }}
-        name: upload
-        uses: actions/upload-artifact@v2
-        with:
-          name: manta-base
-          path: target/release/manta
```
stechu: Can you explain why dropping this section? (It is not obvious to me)

.github/PULL_REQUEST_TEMPLATE.md
Garandor: ok true, as in the case of my #358 there is no value to leading it with manta, as it actually only affects Dolphin.

If y'all don't think seeing in the release notes https://github.com/Manta-Network/Manta/releases/tag/v3.1.4 which RT was affected by which PR provides value, then yeah feel free to drop it. My proposal would be to change 

> [manta] Add a congested_chain_simulation test in Calamari by @ghzlatarev https://github.com/Manta-Network/Manta/pull/337

to

> [Calamari] Add a congested_chain_simulation test by @ghzlatarev https://github.com/Manta-Network/Manta/pull/337

or we could invent little shorthands like
MA/CA/DO + ALL if verbosity is the concern

stechu: The reason we had this before is that we used to have multiple branches. It turns out to be very bad for both productivity and maintenance. Reducing complexity in any part of engineering at no other cost is always preferred.

.github/workflows/publish-draft-releases.yml
ghzlatarev: This test is sometimes failing for some reason.
My hunch was that it didn't have enough time to reach these blocks, so I tried reducing it.
But it seems this might not be the original reason, as it still failed once.
Going to revert this.

.github/workflows/publish-draft-releases.yml
ghzlatarev: Ah yes let me also modify the description a bit in that case, as this is the main point of the PR.
We used to test the runtime upgrade by starting the network with old client and old runtime and upgrading the runtime.
This means if there's a breaking client change (which sometimes Parity does) the test will automatically fail.
That's why the network should just start with new client. And so we don't need this secondary build from the `manta` branch. 
We always upgrade our clients first anyway, before we do the runtime upgrade on production.

.github/workflows/publish-draft-releases.yml
ghzlatarev: Ok I just realized the actual reason. Sometimes blocks are not logged as finalized and their finalization is implied from succeeding blocks being finalized.
And sometimes this is the block number we're look to parse.
Anyway i'm going to add a fix for this in a separate PR. https://github.com/Manta-Network/Manta/issues/420

.github/PULL_REQUEST_TEMPLATE.md
Garandor: My idea here was to substitute the useless tag with one that actually gives valuable insight, instead of simply removing it.
I'm fine with it being gone too tho

.github/ISSUE_TEMPLATE/release.md
```diff
@@ -24,18 +23,23 @@ candidate branch.
     runtime logic.
 - [ ] Check for any upstream storage migrations and perform tests with `try-runtime`, if any.
 
-The following checks can be performed after we have forked off to the release branch.
+The following checks can be performed after we have frozen our release candidate:
+
+- [ ] Code freeze should typically happen one week prior to release, to ensure we have enough time for related testing.
```
Garandor, line 28: I'd say "Notify everyone, especially people with merge rights to `manta` (stechu, Dengjianping , ghzlatarev) that a release is ongoing and no more merges to `manta` should happen until told otherwise."

More specific than just "code freeze"

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -29,10 +28,7 @@ write a little note why.
...
-- [ ] If needed, notify the committer this is a draft-release and a tag is needed after merging the PR.
 - [ ] Verify benchmarks & weights have been updated for any modified runtime logics
-- [ ] If needed, bump `version` for every crate.
 - [ ] If import a new pallet, choose a proper module index for it, and allow it in `BaseFilter`. Ensure **every** extrinsic works from front-end. If there's corresponding tool, ensure both work for each other.
 - [ ] If needed, update our Javascript/Typescript APIs. These APIs are offcially used by exchanges or community developers.
-- [ ] If we're going to issue a new release, freeze the code one week early(it depends, but usually it's one week), ensure we have enough time for related testing.
-- [ ] Check if inheriting any upstream runtime storage migrations. If any, perform tests with `try-runtime`.
+- [ ] When integrating upstream changes check if you are inheriting runtime storage migrations. If any, perform tests with `try-runtime`.
```
Garandor: +
- [ ] Search code changes for added/removed StorageItems ( grep for "StorageValue", "StorageMap" in the diffs and find changes to #[pallet::config] sections ), ensure those changes have migrations and test them with `try-runtime`

Garandor (approved):
Two things I would add/change, LGTM otherwise

.github/PULL_REQUEST_TEMPLATE.md
ghzlatarev: There are more storage items than those two but I get your point.

Garandor (approved):
LGTM!

Dengjianping (approved):
LGTM!

PR #394, opened 2022-02-10
[manta] Dziabko/dolphin ci workflows

### LINKED ISSUE #376: CI workflows for Dolphin runtime
We need a benchmarking workflow to generate custom weights.
Maybe add to the draft-release workflow as well down the line.

dziabko:
## Description

Adds a new dolphin runtime CI workflow

closes: #376 

---

write a little note why.

- [ ] Targeted PR against correct branch (`manta` or `dolphin`) with right title (start with [Manta] or [Dolphin]),
- [ ] Wrote unit tests.

### DIFF
--- a/.github/workflows/generate_dolphin_weights_files.yml
+++ b/.github/workflows/generate_dolphin_weights_files.yml
@@ -107,14 +107,6 @@ jobs:
           -
             extrinsic:
               id: '*'
-              name: calamari_vesting
-            pallet:
-              id: calamari_vesting
-              name: calamari_vesting
-            iterations: 20
-          -
-            extrinsic:
-              id: '*'
               name: frame_system
             pallet:
               id: frame_system
--- a/.github/workflows/generate_dolphin_weights_files.yml
+++ b/.github/workflows/generate_dolphin_weights_files.yml
@@ -1,0 +1,291 @@
+---
+
+# yamllint disable rule:line-length
+
+name: Benchmark Dolphin Runtime & Generate Weights Files
+
+# yamllint disable-line rule:truthy
+on:
+  workflow_dispatch:
+
+env:
+  AWS_INSTANCE_SSH_PUBLIC_KEY: ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIPb24HEQ++aNFWaqVyMtIs6GotUB8R+q61XOoI2z6uMj
+  AWS_REGION: us-east-1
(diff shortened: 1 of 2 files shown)


.github/workflows/generate_dolphin_weights_files.yml
```diff
@@ -0,0 +1,283 @@
...
+              name: pallet_treasury
+            iterations: 20
+          -
+            extrinsic:
+              id: '*'
+              name: manta_collator_selection
+            pallet:
+              id: manta_collator_selection
+              name: manta_collator_selection
+            iterations: 20
```
ghzlatarev, line 210: Could you check if we can make the matrix dynamic and pass a list of the pallets you want to benchmark as input.
If there's no input we can have this as default. Kind of out of scope, but we'd be able to dedupe the 3 workflows if we can do that.

stechu:
@dziabko  Any follow up on this PR?

.github/workflows/generate_dolphin_weights_files.yml
```diff
@@ -0,0 +1,183 @@
...
+    needs:
+      - start-node-builder-current
+      - build-benchmark
+    runs-on: ${{ needs.start-node-builder-current.outputs.runner-label }}
+    strategy:
+      matrix:
+        benchmark:
+          extrinsic:
+            id: '*'
+            name: [frame_system, pallet_balances, pallet_collective, pallet_democracy, pallet_membership, pallet_multisig, pallet_scheduler, pallet_session, pallet_tx_pause, pallet_utility, pallet_timestamp, pallet_treasury, manta_collator_selection]
```
ghzlatarev: These are the pallet names not the extrinsics.

.github/workflows/generate_dolphin_weights_files.yml
ghzlatarev: Nvm this won't work

.github/workflows/publish-draft-releases.yml
```diff
@@ -1129,6 +1133,8 @@ jobs:
             name: calamari
           -
             name: manta
+          -
+            name: dolphin
```
ghzlatarev, line 1137: You need to define some more output variables for the dolphin runtime like `calamari-runtime-current`, `manta-runtime-current`, etc See how they are used and do the same thing. Basically the runtimes version must be equal or the job fails.

PR #383, opened 2022-02-06
[All] Split up manta_primitives lib.rs & [Manta&Calamari] Port parity's fast-runtime cargo feature

Garandor:
## Description
Closes #357 
Port of https://github.com/paritytech/polkadot/pull/4332

Allows building Manta/Calamari runtimes with session and democracy times set to much shorter 2 or 5 minutes instead of their production values and for customizable timings using environment variables.
Env variables are ignored unless the RT is built with `--features=fast-runtime` in which case they overwrite the (shortened) defaults.

Build fast RTs with
`cargo build --release --features=fast-runtime`

or customize, to e.g. 1 minute period interval with env variable
`MANTA_PERIOD=5 cargo build --release --features=fast-runtime`

Variables are named as capitalized versions of their name in the parameter_types! macro prefixed with MANTA_ or CALAMARI_ respectively

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [x] Targeted PR against correct branch (`manta` or `dolphin`) with right title (start with [Manta] or [Dolphin]),
- [x] Linked to Github issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [x] ~Wrote unit tests.~ N/A, buildtime changes only
- [x] ~Updated relevant documentation in the code.~ N/A, buildtime changes only
- [x] Re-reviewed `Files changed` in the Github PR explorer.
- [x] ~If runtime changes, need to update the version numbers properly~ N/A, buildtime changes only 
- [x] ~If needed, notify the committer this is a draft-release and a tag is needed after merging the PR.~ N/A, is feature PR
- [x] ~Verify benchmarks & weights have been updated for any modified runtime logics~ N/A, buildtime changes only
- [x] ~If needed, bump `version` for every crate.~ N/A, buildtime changes only
- [x] ~If import a new pallet, choose a proper module index for it, and allow it in `BaseFilter`. Ensure **every** extrinsic works from front-end. If there's corresponding tool, ensure both work for each other.~ N/A, buildtime changes only
- [x] ~If needed, update our Javascript/Typescript APIs. These APIs are offcially used by exchanges or community developers.~ N/A, buildtime changes only
- [x] ~If we're going to issue a new release, freeze the code one week early(it depends, but usually it's one week), ensure we have enough time for related testing.~ N/A, Is feature PR

PR #401, opened 2022-02-15
Customize cargo profiles

### LINKED ISSUE #295: Utilize profile feature in cargo.toml
After [rust 1.57](https://blog.rust-lang.org/2021/12/02/Rust-1.57.0.html#cargo-support-for-custom-profiles), we can separate `profile` into these parts:
```toml
[profile.release]
opt-level = 3

[profile.bench]
opt-level = 3
debug = false
lto = "thin"
incremental = true
debug-assertions = false

[profile.dev]
opt-level = 3

[profile.test]
opt-level = 3
debug = true
```

https://doc.rust-lang.org/nightly/cargo/reference/profiles.html#profiles

Dengjianping:
## Description

Rust 1.57 supports to customize cargo profile, so we can move rustc flags(like lto, code-unit) to customized profile.

- codegen-units = 1: https://doc.rust-lang.org/nightly/cargo/reference/profiles.html#codegen-units
- opt-level = 3(3: all optimizations)
`
The opt-level setting controls the [-C opt-level flag](https://doc.rust-lang.org/nightly/rustc/codegen-options/index.html#opt-level) which controls the level of optimization. Higher optimization levels may produce faster runtime code at the expense of longer compiler times. Higher levels may also change and rearrange the compiled code which may make it harder to use with a debugger.
`
- lto = true: https://doc.rust-lang.org/nightly/cargo/reference/profiles.html#lto

Add a new profile `production` which inherits from `release`.
How to use this profile:
```
cargo b --profile production
```

closes: #295

---

write a little note why.

- [ ] Wrote unit tests.


Dengjianping:
Before merge this pr, I need to check whether changes in benchmark should be minor.

ghzlatarev (changes requested):
You have the same description for `opt-level` and `codegen-units`

Dengjianping:
Do we need another profile? Like
```
[profile.testnet]
inherits = "release"
panic = "unwind"
```

Current `profile.release` is for production, it will take too much time for compiling while development.

Cargo.toml
```diff
@@ -9,4 +9,10 @@ members = [
 ]
 
 [profile.release]
-panic = "unwind"
\ No newline at end of file
+codegen-units = 1
+lto = true
+opt-level = 3
+panic = "unwind"
```
ghzlatarev: I think we should have a `production` profile with the optimizations and keep the old release profile as is.

.github/workflows/generate_calamari_weights_files.yml
```diff
@@ -84,7 +84,7 @@ jobs:
           CARGO_TERM_COLOR: always
         run: |
           source ${HOME}/.cargo/env
-          CARGO_PROFILE_RELEASE_LTO=true RUSTFLAGS="-C codegen-units=1" cargo build --release --verbose --features=runtime-benchmarks
```
ghzlatarev, line 87: There's one more place this is used - in the `publish_draft_release.yml` workflow to create the binary we release officially.

Dengjianping: > There's one more place this is used - in the `publish_draft_release.yml` workflow to create the binary we release officially.

Thanks! I forgot to `git add` it.

Cargo.toml
Dengjianping: Yeah, I thought about it. The binary will generated under `target/production`, and some path in CI need to be adapted.

Garandor: We could add `--features=fast-runtime` to a new `profile.dev` and maybe `profile.release` to make testnets deploy with fast times by default when https://github.com/Manta-Network/Manta/pull/383 is merged (it's ready to)

Garandor: > I think we should have a `production` profile with the optimizations and keep the old release profile as is.

@ghzlatarev is the production profile used in CI and for the deployed testnets
Otherwise i'm thinking it might be dangerous to test with `release` and deploy `production` that was never actually tested

ghzlatarev: Internal testnets are tested with the draft-release artifacts. Sometimes I use those for local testing as well. That binary is optimized.

Dengjianping: > We could add `--features=fast-runtime` to a new `profile.dev` and maybe `profile.release` to make testnets deploy with fast times by default when #383 is merged (it's ready to)

cargo profile cannot include features, only rustc flags.

ghzlatarev: This looks resolved now, no ?

ghzlatarev (changes requested):
Please mention mention in the description the name of the new profile explicitly, and how to use it and the new target folder.

Garandor:
https://github.com/Manta-Network/Manta#build-mantacalamari-node
build instructions need update to `--prod`

Cargo.toml
ghzlatarev: Please resolve the conversations when you're done.

ghzlatarev (approved):

.github/workflows/generate_calamari_weights_files.yml
```diff
@@ -84,7 +84,7 @@ jobs:
...
+          cargo b --profile production --verbose --features=runtime-benchmarks
```
Garandor: tiny remark, but can we maybe use `build` instead of `b` in all the scripts? it's just more readable, the `b` shorthand makes more sense in console.

Garandor (approved):

stechu:
Is there a way that we can control the rust version when people use `cargo b`?

Dengjianping:
> Is there a way that we can control the rust version when people use `cargo b`?

Thanks for the suggestion! But actually we has specified msrv in our readme, seems it's not enough.

There're three ways to tell people msrv:
1. rust-version = '1.57' in `Cargo.toml`.
2. `msrv = "1.57.0"` in `clippy.toml`.
3. Add a toolchain file `rust-toolchain` which specify a msrv.

I prefer option 1, it's a new feature in rust 2021.
And it will pop a hint like:
```rust-version = '1.61'
cannot be built because it requires rustc 1.61 or newer, while the currently active rustc version is 1.58.1
```

stechu (approved):

PR #402, opened 2022-02-15
[manta] Substrate integration tests

### LINKED ISSUE #370: Better Integration Test
Many things currently we do in QA test can be performed in an automated fashion, for example:

https://github.com/PureStake/moonbeam/blob/master/runtime/moonbeam/tests/integration_test.rs

Related to #347

ghzlatarev:
## Description

closes: #370 

Adding new integration tests for:

- [x] democracy, council, technical committee
- [x] session management
- [x] collator selection, joining and leaving
- [x] tx fee distribution to block authors and treasury
- [x] balances operations
- [x] weights sanity checks
- [x] xcm version snaity checks
- [x] pallet prefixes and indices sanity checks

---

write a little note why.

- [x] Wrote unit tests.
- [x] If needed, bump `version` for every crate.
- [x] If we're going to issue a new release, freeze the code one week early(it depends, but usually it's one week), ensure we have enough time for related testing.


stechu:
Thanks for PR. Let's organize a PR review meeting for this.

runtime/calamari/tests/integration_tests.rs
```diff
@@ -0,0 +1,950 @@
...
+	is_pallet_prefix::<calamari_runtime::Treasury>("Treasury");
+	is_pallet_prefix::<calamari_runtime::Scheduler>("Scheduler");
+	is_pallet_prefix::<calamari_runtime::XcmpQueue>("XcmpQueue");
+	is_pallet_prefix::<calamari_runtime::PolkadotXcm>("PolkadotXcm");
+	is_pallet_prefix::<calamari_runtime::CumulusXcm>("CumulusXcm");
+	is_pallet_prefix::<calamari_runtime::DmpQueue>("DmpQueue");
+	is_pallet_prefix::<calamari_runtime::Utility>("Utility");
+	is_pallet_prefix::<calamari_runtime::Multisig>("Multisig");
+	is_pallet_prefix::<calamari_runtime::Sudo>("Sudo");
+	is_pallet_prefix::<calamari_runtime::CalamariVesting>("CalamariVesting");
```
ghzlatarev, line 820: @stechu actually I'm not clear what your suggestion here was.

runtime/calamari/src/impls.rs
```diff
@@ -31,11 +37,19 @@ impl OnUnbalanced<NegativeImbalance> for DealWithFees {
 	fn on_unbalanceds<B>(mut fees_then_tips: impl Iterator<Item = NegativeImbalance>) {
 		if let Some(fees) = fees_then_tips.next() {
 			// for fees, 40% to treasury, 60% to author
-			let mut split = fees.ration(40, 60);
```
ghzlatarev, line 34: @stechu i know you had a suggestion to make these part of the `DealWithFee` struct, however there wouldn't be a way to access them, because I can't change the signature of `on_unbalanceds` trait function.

runtime/calamari/tests/integration_tests.rs
Garandor: i think the idea was to wrap each individual line into a rust macro, so that the line reads simply e.g.
> pallet_storage_prefix_is!("Multisig")

abstracting away the verbosity of the rust syntax.
One param should be enough for now as the names and identifiers all seem to be identical

ghzlatarev: Not sure how to reduce the verbosity further. I gotta pass a generic for the `name` function. And I gotta pass the string to compare it to in the assertion.

stechu: Ideally the macro takes a list of module name and generate this code automatically, we can leave this to future PRs.

runtime/calamari/src/impls.rs
stechu: yeah, makes sense. The function didn't pass self in.

stechu (approved):

runtime/calamari/tests/integration_tests.rs
ghzlatarev: So for example if I pass it a string token `AuraExt` the macro can generate a generic `calamari_runtime::AuraExt>` ?

Cargo.lock
```diff
@@ -899,8 +899,10 @@ dependencies = [
  "cumulus-pallet-xcm",
  "cumulus-pallet-xcmp-queue",
  "cumulus-primitives-core",
+ "cumulus-primitives-parachain-inherent",
```
Garandor, line 902: why did we get added dependencies in this PR?

Garandor: nvm, just saw the changed cargo.toml

runtime/calamari/tests/common/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+		Scheduler::on_initialize(System::block_number());
+		Democracy::on_initialize(System::block_number());
+	}
+}
+
+pub fn last_event() -> Event {
+	System::events().pop().expect("Event expected").event
+}
+
+pub struct ExtBuilder {
```
Garandor: Wouldn't this be better placed in a mock.rs ?

ghzlatarev: Yeah makes sense.

runtime/calamari/tests/integration_tests.rs
```diff
@@ -0,0 +1,950 @@
...
+
+	ExtBuilder::default()
+		.with_collators(vec![alice.clone()], desired_candidates)
+		.with_balances(vec![
+			(alice.clone(), INITIAL_BALANCE),
+			(bob.clone(), INITIAL_BALANCE),
+		])
+		.build()
+		.execute_with(|| {
+			// Alice is in the invulnerables set, so not part of the candidates set.
```
Garandor, line 550: Just FYI, this assertion is only true for invulnerables at genesis, a candidate that is made invulnerable can be part of both sets.
(tbh. it's a bit strange to me that the latter is not always the case)

Garandor (approved):
Checked the collator_selection tests and everything LGTM!

PR #411, opened 2022-02-23
Add a corner case for collator-selection

### LINKED ISSUE #360: Add a corner case for increasing/decreasing candidate bond
Specifically, one of the checks should be the following:
1. set low bond
2. register candidate
3. increase bond
4. unregister candidate
5. assert unreserved bond is the original (lower) value

Refer to this: https://github.com/paritytech/cumulus/pull/906/files#diff-3ffb0282ef5c5fd986a407ccac39f35a7d44a293b1719fab3ac28b85ad168e9fR87-R112

Dengjianping:
## Description

closes: #360 

---

write a little note why.

- [x] Wrote unit tests.
- [ ] Check if inheriting any upstream runtime storage migrations. If any, perform tests with `try-runtime`.


pallets/collator-selection/src/tests.rs
```diff
@@ -457,3 +456,63 @@ fn remove_collator_should_work() {
...
+		assert_eq!(0, Balances::reserved_balance(candidate_1));
+		assert_eq!(100, Balances::free_balance(candidate_1));
+
+		// remove candidate_2
+		assert_ok!(CollatorSelection::remove_collator(
+			Origin::signed(RootAccount::get()),
+			candidate_2
+		));
+		// check candidate_2
+		assert_eq!(0, Balances::reserved_balance(candidate_2));
```
ghzlatarev: It might be better to reserve some more of the balance in another way in order not check against 0.

Garandor (approved):
LGTM

pallets/collator-selection/src/tests.rs
```diff
@@ -457,3 +456,63 @@ fn remove_collator_should_work() {
 		);
 	});
 }
+
+#[test]
+fn increase_bond_after_register_candidate() {
+	// It's a corner case:
+	// 1. Set orignal bond as 100KMA.
```
stechu: `10` from the code?

pallets/collator-selection/src/tests.rs
```diff
@@ -457,3 +456,63 @@ fn remove_collator_should_work() {
...
+	// 2. Register candidate.
+	// 3. increase bond to 150KMA.
```
stechu: `15` from the code?

pallets/collator-selection/src/tests.rs
```diff
@@ -457,3 +456,63 @@ fn remove_collator_should_work() {
...
+	// 3. Unregister candidate.
+	// 4. The owner should get 100KMA back instead of 150KMA.
```
stechu: See previous comments.

stechu (changes requested):

pallets/collator-selection/src/tests.rs
```diff
@@ -457,3 +456,63 @@ fn remove_collator_should_work() {
...
+		assert_eq!(new_bond, Balances::reserved_balance(candidate_2));
+
+		// remove candidate_1
+		assert_ok!(CollatorSelection::remove_collator(
+			Origin::signed(RootAccount::get()),
+			candidate_1
+		));
+		// check candidate_1
+		assert_eq!(0, Balances::reserved_balance(candidate_1));
+		assert_eq!(100, Balances::free_balance(candidate_1));
```
ghzlatarev: Can you write a comment that they start off with 100 in this mock runtime.

ghzlatarev (approved):

stechu (approved):

Issue #386, opened 2022-02-07
Revisit deposit configurations in Manta/Calamari


ghzlatarev:
- [x] pallet_preimage - BaseDeposit
- [x] pallet_preimage - ByteDeposit
- [ ] pallet_balances - ExistentialDeposit
- [ ] pallet_multisig - DepositBase
- [ ] pallet_multisig - DepositFactor
- [ ] pallet_democracy - MinimumDeposit
- [ ] pallet_democracy - PreimageByteDeposit
- [ ] pallet_treasury- ProposalBond
- [x] pallet_treasury- ProposalBondMinimum
- [x] pallet_treasury- ProposalBondMaximum

pallet_assets: (especially if we decide to allow user callable extrinsics)
- [ ]  pallet_assets- AssetDeposit
- [ ] pallet_assets- AssetAccountDeposit
- [ ] pallet_assets- ApprovalDeposit
- [ ] pallet_assets- MetadataDepositBase
- [ ] pallet_assets- MetadataDepositPerByte;

ghzlatarev:
Increase `BaseDeposit` from 1KMA to 1k - 10k KMA . Karura is around ~6 KAR, Kintsugi is around ~6itbc, KILT is 1 KILT

Increase `ProposalBondMaximum` to 10k KMA (~100USD) The idea for the cap seems to be to not discourage people from proposing really big spends. Kusama's cap is 1k KSM, Karura's cap is 25 KAR, KILT and Interlay don't even set it.

Garandor:
Is thishas not been done in the meantime, we need to ensure these are correct for v4.0.1 @ghzlatarev

PR #436, opened 2022-03-08
Dolphin XCM Integration

### LINKED ISSUE #427: Integrate XCM into the Dolphin runtime
XCM design and tests are ready https://github.com/Manta-Network/Manta/pull/419
Now we need to integrate intol Dolphin
- [x] Integrate the relevant pallets into the Dolphin runtime.
- [x] Adjust `Barrier`
- [x] Test with 2 Dolphin networks.
- [x] Test with Moonbeam and Dolphin networks.
- [x] Test with Relay chain and Dolphin.
- [x] add some debug logs to all the `TooExpensive` errors in `buy_weight` 
- [ ] implement proper Barrier for tests and write must-fail xcm tests: https://github.com/Manta-Network/Manta/issues/442
- [x] equalize dolphin and mock xcm weight configs
- [ ] real weights and benchmarks for `pallet-asset-manager` https://github.com/Manta-Network/Manta/issues/440

ghzlatarev:

## Description

closes: #427

Initial XCM configuration for Dolphin runtime:

* Add `pallet-assets` and `pallet-asset-manager` to Dolphin. `pallet-assets` is not unfiltered and there's no need to unfilter because it only uses root calls.
* Add ORML `XTokens` for message composition. This is the only pallet that is unfiltered and callable by public users.
* There rest of the xcm pallets have almost the same configuration as our mock for tests, with a few differences:
   - Bump xcm weight configuration to more reasonable values:
     - `BaseXcmWeight` from 1 to  100_000_000 
     - `UnitWeightCost` from 1 to 1_000_000_000
  - Implement a Barrier with few additional implementors:
    - `TakeWeightCredit` is needed to allow local origin xcm messages which call `execute_xcm_in_credit` with non-zero `weight_credit` , hence passing the barrier.  
    - `AllowTopLevelPaidExecutionFrom<Everything>` is needed for messages that are handled by the xcm queues (so non-local origin).  which call `execute_xcm_in_credit` is called with 0 `weight_credit`. And which will fail the `TakeWeightCredit` barrier
    - `AllowUnpaidExecutionFrom<ParentOrParentsExecutivePlurality>` is needed to receive unpaid message with instructions like `Transact` from relay chain. For example for emergency runtime upgrades.
* Switched to `Nothing` for `XcmTeleportFilter` and `XcmReserveTransferFilter` to be on the safe side. We can allow these when we decide to unfilter `pallet_xcm`.

---

write a little note why.

- [ ] Wrote unit tests.
- [ ] Check if inheriting any upstream runtime storage migrations. If any, perform tests with `try-runtime`.


node/src/chain_specs/dolphin.rs
```diff
@@ -205,7 +208,7 @@ fn dolphin_dev_genesis(
 		aura_ext: Default::default(),
 		parachain_system: Default::default(),
 		polkadot_xcm: dolphin_runtime::PolkadotXcmConfig {
-			safe_xcm_version: Some(0),
+			safe_xcm_version: Some(SAFE_XCM_VERSION),
```
Dengjianping, line 211: Just mark it here: this version is `1` in kusama.

runtime/dolphin/src/lib.rs
```diff
@@ -241,6 +246,7 @@ impl Contains<Call> for BaseFilter {
 				| manta_collator_selection::Call::remove_collator{..}
 				| manta_collator_selection::Call::leave_intent{..})
 			| Call::Balances(_)
+			| Call::XTokens(_)
```
Dengjianping, line 249: Does `AssetManager` need to be allowed here?

Dengjianping:
I just realized that `asset-manager` is not benchmarked.

runtime/dolphin/src/lib.rs
ghzlatarev: Not necessarily. All of the calls are root calls, so they bypass the filter. But I can also unfilter them.

node/src/chain_specs/dolphin.rs
ghzlatarev: This is actually 2 and is imported from Polkadot.

ghzlatarev:
> I just realized that `asset-manager` is not benchmarked.

Yeah asset-manager does not have the benchmarking macros implemented yet, so that's why.
It will be done separately. https://github.com/Manta-Network/Manta/issues/440

node/src/chain_specs/dolphin.rs
Dengjianping: Sorry, I didn't explain it well here. I mean it's 1 on kusama. So if it's 2, maybe we should align it with kusama.

ghzlatarev: Ah ok, idk why they are still at 1, all parachains are already at 2.
Not sure some of the scenarios will work. I've only tested with 0 and 2.
For example, with version 0 the send-to-reserve scenarios didn't work.

ghzlatarev: Actually this is the safe xcm version not the version of the code used. It means that if the system doesn't know the destination's version it will try to encode the message with the version set in this storage item. 
The XCM code itself in Kusama is v2 already.  `use xcm::latest::prelude::*;`
Also V3 is about to be released https://github.com/paritytech/polkadot/pull/4097

runtime/dolphin/src/lib.rs
```diff
@@ -582,6 +588,93 @@ impl pallet_preimage::Config for Runtime {
 	type ByteDeposit = PreimageByteDeposit;
 }
 
+parameter_types! {
+	// Does not really matter as this will be only called by root
+	pub const AssetDeposit: Balance = 0;
```
Dengjianping, line 595: `Const<u128>`

runtime/dolphin/src/lib.rs
ghzlatarev: Added a comment to clear it up.
As per our discussion let's keep it unfiltered by choice, so any user callable extrinsics that we add to AssetManager will have to be manually unfiltered.

runtime/dolphin/src/lib.rs
```diff
@@ -736,7 +881,7 @@ impl pallet_xcm::Config for Runtime {
 	/// This means that no location will pass XcmExecuteFilter, so a dispatched `execute` message will be filtered.
 	/// This shouldn't be reachable since `LocalOriginToLocation = ();`, but let's be on the safe side.
```
ghzlatarev, line 882: Revisit this comment.

runtime/dolphin/src/lib.rs
```diff
@@ -736,7 +881,7 @@ impl pallet_xcm::Config for Runtime {
...
 	type XcmExecuteFilter = Nothing;
-	type XcmExecutor = XcmExecutor<XcmConfig>;
+	type XcmExecutor = XcmExecutor<XcmExecutorConfig>;
 	type XcmTeleportFilter = Everything;
 	type XcmReserveTransferFilter = Everything;
```
ghzlatarev: Let's be on the safe side and enable things 1 by 1 as we need them. That's why changing these 2 to Nothing

runtime/dolphin/src/lib.rs
```diff
@@ -746,23 +891,66 @@ impl pallet_xcm::Config for Runtime {
...
+				Ok(location) => Some(location),
+				Err(_) => None,
+			},
+		}
+	}
+}
+
+parameter_types! {
+	pub const BaseXcmWeight: Weight = 100_000_000;
+	pub const MaxAssetsForTransfer: usize = 2;
```
ghzlatarev: Not sure if there's a specific reason why everyone chose 2 for this configuration.
I haven't hit a scenario where I'd need to send 2 distinct assets so far and our tests all pass with this set to 1.

runtime/dolphin/src/lib.rs
```diff
@@ -866,11 +1054,16 @@ construct_runtime!(
...
 		DmpQueue: cumulus_pallet_dmp_queue::{Pallet, Call, Storage, Event<T>} = 33,
+		XTokens: orml_xtokens::{Pallet, Call, Event<T>, Storage} = 34,
 
 		// Handy utilities.
 		Utility: pallet_utility::{Pallet, Call, Event} = 40,
 		Multisig: pallet_multisig::{Pallet, Call, Storage, Event<T>} = 41,
 		Sudo: pallet_sudo::{Pallet, Call, Config<T>, Storage, Event<T>} = 42,
+
+		// Assets management
+		Assets: pallet_assets::{Pallet, Call, Storage, Event<T>} = 51,
```
ghzlatarev: Use same indices as manta-pay pr

runtime/dolphin/tests/xcm_tests.rs
```diff
@@ -553,18 +553,45 @@ fn send_para_a_native_asset_para_b_and_then_send_back() {
...
+			parachain::Origin::root(),
+			0,
+			0u128
+		));
+		assert_eq!(
+			Some(0),
+			parachain::AssetManager::location_asset_id(AssetLocation(VersionedMultiLocation::V1(
+				MultiLocation::new(1, X1(Parachain(2)),)
+			)))
+		);
```
ghzlatarev, line 582: maybe revert these tests and leave them for the tests PR.

runtime/dolphin/src/lib.rs
ghzlatarev: I don't want to change the Balance constants, as we will be assuming that Balance is configured to u128

node/src/chain_specs/dolphin.rs
```diff
@@ -28,6 +28,9 @@ const DOLPHIN_PROTOCOL_ID: &str = "dolphin"; // for p2p network configuration
 const KUSAMA_RELAYCHAIN_LOCAL_NET: &str = "kusama-local";
 const KUSAMA_RELAYCHAIN_DEV_NET: &str = "kusama-dev";
 
+/// The default XCM version to set in genesis config.
+const SAFE_XCM_VERSION: u32 = xcm::prelude::XCM_VERSION;
```
stechu: I think we should make the version number explicit, instead relying on some external defined version number.

runtime/dolphin/src/lib.rs
```diff
@@ -241,10 +246,13 @@ impl Contains<Call> for BaseFilter {
 				| manta_collator_selection::Call::remove_collator{..}
 				| manta_collator_selection::Call::leave_intent{..})
 			| Call::Balances(_)
+			| Call::XTokens(_)
 			| Call::Utility(_) => true,
+			// Filter XCM pallets, we only allow transfer with XTokens.
```
stechu, line 251: You also need to add `AssetManager` here.

runtime/dolphin/src/lib.rs
```diff
@@ -582,6 +590,92 @@ impl pallet_preimage::Config for Runtime {
 	type ByteDeposit = PreimageByteDeposit;
 }
 
+parameter_types! {
+	// Does not really matter as this will be only called by root
+	pub const AssetDeposit: Balance = 0;
+	pub const AssetAccountDeposit: Balance = 0;
+	pub const ApprovalDeposit: Balance = 0;
+	pub const MetadataDepositBase: Balance = 0;
+	pub const MetadataDepositPerByte: Balance = 0;
```
stechu, line 599: Can you add a FIXME here? I think these constants need to be revisited here.

runtime/dolphin/src/lib.rs
```diff
@@ -621,24 +716,21 @@ pub type LocationToAccountId = (
...
+	// Transacting native currency, i.e. MANTA, KMA, DOL
 	Balances,
-	// Use this currency when it is a fungible asset matching the given location or name:
-	IsConcrete<KsmLocation>,
-	// Do a simple punn to convert an AccountId32 MultiLocation into a native chain account ID:
+	IsNativeConcrete<SelfReserve>,
 	LocationToAccountId,
-	// Our chain's account ID type (we can't get away without mentioning it explicitly):
 	AccountId,
-	// We don't track any teleports.
```
stechu: Why removing the comment here?

runtime/dolphin/src/lib.rs
```diff
@@ -621,24 +716,21 @@ pub type LocationToAccountId = (
...
+/// Transactor for native currency, i.e. implements `fungible` trait
+pub type LocalAssetTransactor = XcmCurrencyAdapter<
+	// Transacting native currency, i.e. MANTA, KMA, DOL
 	Balances,
-	// Use this currency when it is a fungible asset matching the given location or name:
-	IsConcrete<KsmLocation>,
-	// Do a simple punn to convert an AccountId32 MultiLocation into a native chain account ID:
+	IsNativeConcrete<SelfReserve>,
 	LocationToAccountId,
-	// Our chain's account ID type (we can't get away without mentioning it explicitly):
```
stechu, line 632: please keep the comments.

runtime/dolphin/src/lib.rs
```diff
@@ -621,24 +716,21 @@ pub type LocationToAccountId = (
...
-pub type LocalAssetTransactor = CurrencyAdapter<
-	// Use this currency:
+/// Transactor for native currency, i.e. implements `fungible` trait
+pub type LocalAssetTransactor = XcmCurrencyAdapter<
+	// Transacting native currency, i.e. MANTA, KMA, DOL
 	Balances,
-	// Use this currency when it is a fungible asset matching the given location or name:
-	IsConcrete<KsmLocation>,
-	// Do a simple punn to convert an AccountId32 MultiLocation into a native chain account ID:
+	IsNativeConcrete<SelfReserve>,
```
stechu, line 724: Please add a comment here.

runtime/dolphin/src/lib.rs
stechu: yeah, 1 is probably fine, too.

stechu (changes requested):
LGTM in general. Some minor changes requested.

runtime/dolphin/src/lib.rs
ghzlatarev: I left it unfiltered on purpose.
`AssetManager`'s `T::ModifierOrigin` is configured to Root so all of the extrinsics bypass this filter right now.
But if we ever add public user callable extrinsic, or change the `ModifierOrigin`, I want to be forced to manually unfilter it. That seems like the safer way to go.
Anyway adding it is not a big deal if this idea is gimicky.

runtime/dolphin/src/lib.rs
ghzlatarev: I've added them to this issue https://github.com/Manta-Network/Manta/issues/386
Btw the comment here is still correct. These don't really matter because only Root can call the extrinsics.

runtime/dolphin/src/lib.rs
ghzlatarev: Mistake.

runtime/dolphin/src/lib.rs
stechu: oh, cool!

stechu (approved):

Garandor (approved):

PR #441, opened 2022-03-09
Dolphin benchmarking workflow

### LINKED ISSUE #376: CI workflows for Dolphin runtime

ghzlatarev:
## Description

closes: #376

* Benchmarking workflow for Dolphin
* Needs to be merged before being used.

---

write a little note why.

- [x] Wrote unit tests.

### DIFF
--- a/.github/workflows/generate_dolphin_weights_files.yml
+++ b/.github/workflows/generate_dolphin_weights_files.yml
@@ -1,0 +1,283 @@
+---
+
+# yamllint disable rule:line-length
+
+name: Benchmark Dolphin Runtime & Generate Weights Files
+
+# yamllint disable-line rule:truthy
+on:
+  workflow_dispatch:
+
+env:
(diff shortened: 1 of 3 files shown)


ghzlatarev:
We don't really need this actually.

.github/workflows/generate_dolphin_weights_files.yml
```diff
@@ -0,0 +1,283 @@
...
+          cargo build --profile production --verbose --features=runtime-benchmarks
+      -
+        name: stop sccache server
+        run: sccache --stop-server || true
+      -
+        name: upload
+        uses: actions/upload-artifact@v2
+        with:
+          name: manta
+          path: target/release/manta
```
Dengjianping: Should be `target/production/manta`

Dengjianping (approved):
LGTM!

PR #449, opened 2022-03-16
[manta]Add strip profile to production

### LINKED ISSUE #325: Remove strip operation in CI
We can remove strip command in CI due to this pr stablized: https://github.com/rust-lang/cargo/pull/10088
But not sure rust 1.58 will include this feature.

Dengjianping:

## Description

Remove `strip` from CI, and add `strip` profile to production.

https://doc.rust-lang.org/stable/cargo/reference/profiles.html#strip

closes: #325

---

write a little note why.

- [ ] Wrote unit tests.


Garandor:
Seems like srtool doesn't want the runtime to be built without strip:
https://github.com/Manta-Network/Manta/runs/5572932228?check_suite_focus=true#step:4:340

Dengjianping:
> Seems like srtool doesn't want the runtime to be built without strip: https://github.com/Manta-Network/Manta/runs/5572932228?check_suite_focus=true#step:4:340

Yeah, srtool is blocking us, I just created an issue in srtool, not sure when it will be fixed.
https://github.com/paritytech/srtool/issues/40

Garandor:
one of us might be misinterpreting the error
Caused by:
  ║ feature `strip` is required

The way I understand it It's not about not being stabilized (the last part of the error message is misleading here), it's about the CI build in your change trying to make a binary without `strip` enabled.
It'd be logical to me if srtool is preventing wasm runtime unstablenessbuild without `strip` as those should never end up onchain.

So it'd be about srtool allowing unstripped builds, not about unstableness

EDIT: nvm, just saw that `strip` was being invoked as a standalone command here https://github.com/Manta-Network/Manta/pull/449/files#diff-84486036e95f7e22168734603cbdd311736c85fb23a0d0e0660988a67d3605ecL163
so the misunderstanding was on my part

stechu:
@Garandor  can you give an explicit review of this PR (approve or request change)?

Cargo.toml
```diff
@@ -12,6 +12,7 @@ members = [
 inherits = "release"
 codegen-units = 1
 lto = true
+strip = "symbols"
```
Garandor, line 17: Do we use `--production` in the CI somewhere or is this manual exeuction only?
If so, is the release checklist adapted so we don't mistakenly publish a `--release` build

Dengjianping: Our CI has been switched to `production` profile.

Dengjianping:
> 

We have to wait, it seems parity will publish a release(0.9.20) for fixing this problem.

Garandor (approved):
LGTM for 3.2.1

Garandor (changes requested):
merge conflicts

Garandor (approved):
LGTM

ghzlatarev (approved):

PR #454, opened 2022-03-22
Fix potential expect() panic in runtime code, two xcm must-fail tests, and filter out unneeded XTokens extrinsics

### LINKED ISSUE #442: Improve XCM tests
Equalize Mock environment with Dolphin:
- [x] implement same Barrier and adjust tests
- [x] use same weights configurations
- [x] equalize relay configs with Kusama/Polkadot ---> https://github.com/Manta-Network/Manta/issues/648

Write must-fail xcm tests: 
- [x] `destWeight` ->  `Barrier` 
- [x] low `amount` -> `TooExpensive`
- [x] high `amount` -> `XcmExecutionFailed`
- [x] more self_insufficient tests

More:
- [x] fuzzer tests (AFL framework) ---> https://github.com/Manta-Network/Manta/issues/647
- [x] integration tests that xcm fees go to correct account - `Treasury` ---> https://github.com/Manta-Network/Manta/issues/649

ghzlatarev:
## Description

closes: #442

* The `take_revenue` function of our XCM code had an `expect` call which can potentially panic. Removing the expect and handling the error now.
* Added a couple of test scenarios where it that code would previously panic.
* Filter out all `XTokens` extrinsics except `transfer()` until they are practically needed.

---

write a little note why.

- [x] Wrote unit tests.


runtime/primitives/src/xcm.rs
```diff
@@ -278,7 +278,8 @@ impl<
 			Ok((asset_id, amount)) => {
 				if !amount.is_zero() {
 					Assets::mint_into(asset_id, &ReceiverAccount::get(), amount)
-						.expect("`mint_into` cannot generally fail; qed");
+						.map_err(|err| log::debug!("mint_into failed with {:?}", err))
```
Dengjianping: Good catch.
Just a suggestion, you can add a filter flag here, like
```rust
log::debug!(target: "manta-xcm", "mint_into failed with {:?}", err)
```

So now, we can filter manta xcm by this param `-lmanta-xcm=debug`

ghzlatarev: Aaa that's what this `target` is for. Yeah will do.

Dengjianping (approved):
LGTM!

stechu (approved):

PR #455, opened 2022-03-22
Calamari: Integrate new collator eviction from #358

### LINKED ISSUE #416: Integrate Calamari RT with blockcount-based eviction of manta_collator_selection
Currently, Calamari is set to an old revision of the manta_collator_selection pallet (TODO ADD LINK AFTER MERGE) in order to use the new eviction feature introduced in #358 on Dolphin only (for testing and stabilization).
This crutch is needed as both runtimes share the manta_collator_selection pallet code.

After the new code has been sufficiently tested, it needs to be integrated with and deployed to Calamari.

Necessary steps:
- [ ] Copy related RT changes from Dolphin to Calamari
- [ ] Write a migration to delete the now unused storage items
  - [ ] LastAuthoredBlock
  - [ ] KickThreshold
- [ ] Re-benchmark weights
- [ ] Deploy to and Upgrade the RT on-chain
- [ ] Enable eviction by having governance execute the `set_eviction_baseline(80)` and `set_eviction_tolerance(10)` extrinsics

Child of #335

Garandor:
## Description

Enable Calamari runtime to use the eviction mechanism from #358 originally merged to dolphin only

- [x] Write a migration for  KickThreshold
- [x] Write a migration for  LastAuthoredBlock
- [x] Test with try_runtime

closes: #416

---

write a little note why.

- [x] Wrote unit tests.


runtime/calamari/src/weights/manta_collator_selection.rs
```diff
@@ -46,187 +46,207 @@ pub trait WeightInfo {
...
 /// Weights for manta_collator_selection using the Substrate node and recommended hardware.
 pub struct SubstrateWeight<T>(PhantomData<T>);
 impl<T: frame_system::Config> manta_collator_selection::WeightInfo for SubstrateWeight<T> {
 	// Storage: CollatorSelection Invulnerables (r:0 w:1)
 	fn set_invulnerables(b: u32, ) -> Weight {
-		(10_564_000 as Weight)
-			// Standard Error: 15_000
-			.saturating_add((118_000 as Weight).saturating_mul(b as Weight))
+		(10_476_000 as Weight)
+			// Standard Error: 18_000
```
ghzlatarev: How did you generate this weights file ?

runtime/calamari/src/lib.rs
```diff
@@ -922,8 +923,24 @@ pub type Executive = frame_executive::Executive<
...
+	#[cfg(feature = "try-runtime")]
+	fn pre_upgrade() -> Result<(), &'static str> {
+		CollatorSelection::pre_migrate_to_v2()
+	}
+	#[cfg(feature = "try-runtime")]
+	fn post_upgrade() -> Result<(), &'static str> {
+		CollatorSelection::post_migrate_to_v2()
+	}
+}
+
```
ghzlatarev, line 943: You can also implement `OnRuntimeUpgrade` for some struct, feature-gated by `try-runtim`.
In the method implementations you can write to temporary storage and check invariants before and after any runtime upgrade.
Something simple to check is that collators still exist and are still the same accounts.
https://www.crowdcast.io/e/substrate-seminar/41 at 1hour 15min gives an example

runtime/calamari/src/lib.rs
```diff
@@ -116,7 +116,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("calamari"),
 	impl_name: create_runtime_str!("calamari"),
 	authoring_version: 1,
-	spec_version: 3141,
+	spec_version: 3142,
```
ghzlatarev: I guess you can bump this to trigger the CI runtime upgrade test, but let's revert it now.
We only bump these on release PRs.

ghzlatarev: Oh and there's a check in the CI that Calamari and Manta runtime versions are kept synchronized.

runtime/calamari/src/weights/manta_collator_selection.rs
Garandor: Originally, from the benchmark calamari github action right before #358 was merged. Back then I only applied the output file to Dolphin. I now copied it back from dolphin to Calamari.

runtime/calamari/src/lib.rs
Garandor: I tried having at least the pre/post_upgrade hooks gated by `try-runtime`, but then the runtime wouldn't compile.
I don't think the `on_runtime_upgrade` hook can't be featuregated as it's needed without try_runtime for the actual release. If the whole struct was featuregated, I think it would apply to the main fn as well

runtime/calamari/src/lib.rs
Garandor: Just saw it failed - neat :)

Garandor: reverted

runtime/calamari/src/weights/manta_collator_selection.rs
ghzlatarev: Can you just run the Calamari benchmark workflow on this branch

Garandor: Sure, I thought it won't be necessary since we'll rebenchmark on release building anyway, but i think it makes sense to see the exact impact of this changeset alone

runtime/calamari/src/lib.rs
ghzlatarev: I mean exactly what you did for `CollatorSelectionMigrationV2 ` but you can read the collators `pre_upgrade`, write them to temp storage and then read the temp storage in `post_upgrade` and compare to what is in the storage then, to make sure they are the same.

Garandor: updated

runtime/calamari/src/weights/manta_collator_selection.rs
Garandor: Updated. In short: Extrinsics lose weight, note_author (once per block) and new_session(once per session) get heavier

runtime/calamari/src/lib.rs
Garandor: I think the "upgrade checker" migrations are worthy of their own issue. I added one: #463
I also added a check that the expected migrations were executed

Garandor:
`cargo run --release --features=try-runtime try-runtime --chain calamari-dev on-runtime-upgrade live -p CollatorSelection -u wss://ws.calamari.systems:443` completes successfully.

```
2022-03-25 19:53:03 found matching spec name: "calamari"    
2022-03-25 19:53:03 spec version mismatch (local 3141 != remote 3140). This could cause some issues.    
2022-03-25 19:53:03 Precheck: KickThreshold does not exist    
2022-03-25 19:53:03  >>> Executing collator-selection V1->V2 migration!    
2022-03-25 19:53:03  >>> Cleaned 9 keys from LastAuthoredBlock    
2022-03-25 19:53:03  >>> Removed LastAuthoredBlock from storage    
2022-03-25 19:53:03  !!! Chain did not have KickThreshold in storage. This is uenexpected but is possible if the genesis config was never changed    
2022-03-25 19:53:03 ✅ no migration for System    
```

Ready to merge

ghzlatarev:
> 2022-03-25 19:53:03 spec version mismatch (local 3141 != remote 3140). This could cause some issues.

You wanna keep the local and remote chain spec versions the same. I haven't looked at the code of `try-runtime` as to exactly why that is though.
That's why I made the CI job manual, so the dev can just branch off, adjust the chain-spec and run it. Instead of having some logic to only trigger it when the versions are the same.

Garandor:
> You wanna keep the local and remote chain spec versions the same.

Re-ran it with matching spec versions just in case.

```
2022-03-27 12:59:27.198 TRACE main executor: Request for native execution succeeded (native: calamari-3140 (calamari-1.tx4.au1), chain: calamari-3140 (calamari-1.tx4.au1))    
2022-03-27 12:59:27.387  INFO main jsonrpsee_ws_client::transport: Connection established to target: Target { sockaddrs: [], host: "ws.calamari.systems", host_header: "ws.calamari.systems:443", mode: Tls, path_and_query: "/" } 
2022-03-27 12:59:27.448  INFO main try-runtime::cli: found matching spec name: "calamari"    
2022-03-27 12:59:27.448  INFO main try-runtime::cli: found matching spec version: 3140    
2022-03-27 12:59:27.450  WARN main manta_collator_selection::migrations: Precheck: KickThreshold does not exist    
2022-03-27 12:59:27.450  INFO main manta_collator_selection::migrations: Executing collator-selection V1->V2 migration!    
2022-03-27 12:59:27.450  INFO main manta_collator_selection::migrations:  >>> Cleaned 9 keys from LastAuthoredBlock    
2022-03-27 12:59:27.450  INFO main manta_collator_selection::migrations:  >>> Removed LastAuthoredBlock from storage    
2022-03-27 12:59:27.450  WARN main manta_collator_selection::migrations:  !!! Chain did not have KickThreshold in storage. This is uenexpected but is possible if the genesis config was never changed    
2022-03-27 12:59:27.450  INFO main runtime::frame-support: ✅ no migration for System    
```

pallets/collator-selection/src/lib.rs
```diff
@@ -115,6 +116,8 @@ pub mod pallet {
 		}
 	}
 
+	const STORAGE_VERSION: StorageVersion = StorageVersion::new(2);
```
Dengjianping: Why `2`? Currently, it's `0` on calamari mainnet.

pallets/collator-selection/src/migrations.rs
```diff
@@ -0,0 +1,110 @@
...
+};
+///  This migrates the pallet from the standard version by parity to our modified storage
+impl<T: Config> Pallet<T> {
+	pub fn migrate_to_v2() -> frame_support::weights::Weight {
+		use frame_support::migration::{
+			have_storage_value, remove_storage_prefix, storage_key_iter,
+		};
+		// Storage migrations should use storage versions for safety.
+		if Self::on_chain_storage_version() < 2 {
+			log::info!(" >>> Executing collator-selection V1->V2 migration!");
```
Dengjianping: Just a suggestion, you can add a filter flag here, like
```rust
log::info!(target: "collator-selection", ">>> Executing collator-selection V1->V2 migration!")
```
Then you can filter collator-selection with this flag
```
-lcollator-selection=info
```
You can see here how to handle logs.
https://substrate.recipes/runtime-printing.html#printing-from-the-runtime

Dengjianping:
~~Just last question here, @Garandor do you try to migrate the storage in local node?~~

Sorry I didn't see the above conversation.

Garandor:
> ~Just last question here, @Garandor do you try to migrate the storage in local node?~

It's a good one actually since I didn't test it with a local chain, I trusted `try-runtime` to be a better test than a manual upgrade on a `polkadot-launch` net.
I'll try it locally as well, if only to walk myself through the runtime upgrade process for the first time since tutorials

pallets/collator-selection/src/lib.rs
Garandor: Thanks for reviewing Jamie, I'll bump it down. I was also thinking of distinguishing it from parity's pallet_collator_selection which after migrating will likely have a StorageVersion of 1 (they haven't migrated yet and code doesn't have a STORAGE_VERSION annotation: https://github.com/paritytech/cumulus/blob/master/pallets/collator-selection/src/lib.rs)
But that's just an assumption and i think we should just use "our revision" in this field

pallets/collator-selection/src/migrations.rs
Garandor: @Dengjianping Wouldn't usefulness of that depend on us changing _all_ logs of the pallet to use the target flag?
Are parity pallets annotated like this already? We can change our code but not theirs

Dengjianping: It's just a suggestion, and I think it will be removed after storage migration. So it's fine if you don't use that way.

You can search keyword like `log::debug!` or `log::info!` in substrate, there're many examples.

Garandor: Only the upgrade struct in the calamari RT will be removed, the migration code here in the pallet can remain. I made the migrations module for that specific purpose.
- for future reference on migrations
- to keep a log of the migrations we applied to the chain

stechu (approved):
LGTM, please fix the existing conversation.

pallets/collator-selection/src/migrations.rs
Garandor: @Dengjianping From the doc you linked:
> If you do not specify the logger target, it will be set to the crate's name 

So by omitting the target, `-lmanta-collator-selection=debug` sets this code's loglevel, which I think is better than providing an arbitrary string

pallets/collator-selection/src/lib.rs
Garandor: addressed in recent commit

pallets/collator-selection/src/migrations.rs
Garandor: @Dengjianping please resolve if acceptable

Dengjianping (approved):
lgtm!

PR #458, opened 2022-03-24
Impl multilocation filter for asset-manager

Dengjianping:

## Description

closes: #XXXX

---

write a little note why.

- [x] Wrote unit tests.


pallets/asset-manager/src/tests.rs
```diff
@@ -189,3 +165,50 @@ fn update_asset() {
...
+
+#[test]
+fn filter_asset_location_should_work() {
+	let kusama_asset_metadata =
+		create_asset_metadata("Kusama", "KSM", 12, 1u128, None, false, false);
+	let kusama_location = AssetLocation(VersionedMultiLocation::V1(MultiLocation::parent()));
+
+	let manta_asset_metadata =
+		create_asset_metadata("Manta", "MANTA", 18, 1u128, None, false, false);
+	let mantan_location = AssetLocation(VersionedMultiLocation::V1(MultiLocation::new(
```
Garandor, line 177: just a typo -> mantan

pallets/asset-manager/src/tests.rs
```diff
@@ -189,3 +165,50 @@ fn update_asset() {
...
+			Some(mantan_location.clone())
+		);
+
+		let new_location = AssetLocation(VersionedMultiLocation::V1(MultiLocation::new(
+			1,
+			X2(Parachain(1999), GeneralKey(b"UNKNOWN".to_vec())),
+		)));
+
+		// new location should be filtered
+		assert!(!crate::Pallet::<Runtime>::contains(&new_location));
```
Garandor, line 210: I'm not very familiar with the XCM code yet, but is this actually testing any "filtering"?

The way i'm reading this logic flow `new_location` is just a local variable that hasn't been given to `AssetManager::register_asset` yet (which presumably is where any filtering of foreign assets would happen), so the assert above will always be satisfied unless `new_location` happens to be a clone of `kusama_location` or `mantan_location`

I think I'm mainly confused by the "filtering" naming of the PR. Maybe you could add some description text to it to elaborate. The way I understand the code, the above test doesn't check filtering, it's only a sanity check/unit test for the `contains` implementation.

Garandor (changes requested):
since you implemented contains on the pallet, maybe this
https://github.com/Manta-Network/Manta/blob/manta/pallets/asset-manager/src/lib.rs#L188 and
https://github.com/Manta-Network/Manta/blob/manta/pallets/asset-manager/src/lib.rs#L232
could be rewritten to Self.contains()

pallets/asset-manager/src/tests.rs
stechu: filter means white-list. 
I agree the test should be in an end to end manner.

stechu:
To test multilocation filter works, an end to end test to show that sending to location that not in the whitelist should return proper error.

Dengjianping:
> 

I have to wait this PR #437 to be merged, because `xtokens` has this feature in 0.9.17.

stechu:
> > 
> 
> I have to wait this PR #437 to be merged, because `xtokens` has this feature in 0.9.17.

Shall we move this out of 3.1.5 then?

Dengjianping:
This feature will be implemented by #488 . I'm going to close it.

PR #461, opened 2022-03-25
Fix AssetManager's update_asset_metadata

### LINKED ISSUE #462: Fix AssetManager's update_asset_metadata
* `AssetManager`'s `update_asset_metadata` is only updating an internal structure which is not being used anywhere
* Fix this as it should also update the underlying Assets data
* Add a test to check metadata items are actually updated

ghzlatarev:

## Description

closes: #462

* `AssetManager`'s `update_asset_metadata` was not actually updating the underlying `Assets` data, so now the extrinsic is calling the `T::AssetRegistrar::update_asset_metadata` in order to do so.
* Added a test case to make sure that metadata items, like symbol, name, decimals are being updated.
* `NextAssetId` storage should be set to `StartNonNativeAssetId` instead of `NativeAssetId` on genesis. Fix benchmarking, add asset-manager to benchmarking workflow.

---

write a little note why.

- [x] Wrote unit tests.


Dengjianping (approved):
LGTM!

stechu:
If you look at `AssetRegistrarMetadata`'s implementation, like:

https://github.com/Manta-Network/Manta/blob/manta/runtime/primitives/src/assets.rs#L30 
```rust
/// The metadata of a Manta Asset
#[derive(Clone, Default, Eq, Debug, PartialEq, Ord, PartialOrd, Encode, Decode, TypeInfo)]
pub struct AssetRegistarMetadata<Balance> {
	pub name: Vec<u8>,
	pub symbol: Vec<u8>,
	pub decimals: u8,
	pub evm_address: Option<H160>,
	pub is_frozen: bool,
	pub min_balance: Balance,
	/// `is_sufficient`: Whether a non-zero balance of this asset is deposit of sufficient
	/// value to account for the state bloat associated with its balance storage. If set to
	/// `true`, then non-zero balances may be stored without a `consumer` reference (and thus
	/// an ED in the Balances pallet or whatever else is used to control user-account state
	/// growth).
	/// For example, if is_sufficient set to `false`, a fresh account cannot receive XCM tokens.
	pub is_sufficient: bool,
}
```
This is more than what `Pallet-Asset` can store. That's why `AssetIdMetadata` is needed as a storage item.

pallets/asset-manager/src/lib.rs
```diff
@@ -320,7 +313,9 @@ pub mod pallet {
 				AssetIdLocation::<T>::contains_key(&asset_id),
 				Error::<T>::UpdateNonExistAsset
 			);
-			AssetIdMetadata::<T>::insert(&asset_id, &metadata);
+
+			T::AssetRegistrar::update_asset_metadata(asset_id, metadata.clone().into())?;
```
stechu: Despite we use `transactional`, it is still a good practice to put the code that could fail before the code should never fail.

stechu (changes requested):

pallets/asset-manager/src/lib.rs
```diff
@@ -256,18 +261,26 @@ pub mod pallet {
 			metadata: <T::AssetConfig as AssetConfig>::AssetRegistrarMetadata,
 		) -> DispatchResult {
 			T::ModifierOrigin::ensure_origin(origin)?;
+			ensure!(
+				!asset_id.is_one(),
```
stechu: The logic only works for a specific configuration.
Should check `AssetConfig<Config>::NativeAssetId` instead of `one`.

stechu (changes requested):
see the detailed comments.

stechu (approved):

PR #467, opened 2022-03-29
Fix ssl compilation issue in CI

Dengjianping:

## Description

closes: #XXXX

---

write a little note why.

- [ ] Wrote unit tests.

### DIFF
--- a/pallets/manta-pay/Cargo.toml
+++ b/pallets/manta-pay/Cargo.toml
@@ -37,11 +37,7 @@ std = [
 	"frame-system/std",
 	"sp-runtime/std",
 	"sp-std/std",
-	"manta-util/std",
-	"manta-sdk/download",
 	"manta-primitives/std",
-	"tempfile",
-	"rand",
 ]
 
 # Precompute Benchmark Transactions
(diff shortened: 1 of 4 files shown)


stechu:
Can you cherry pick the cargo.toml change I have in the other PR here as well?

Dengjianping:
> 

Let's clean these features in std by another pr.

Dengjianping:
> Can you cherry pick the cargo.toml change I have in the other PR here as well?

I will do it.

pallets/manta-pay/Cargo.toml
```diff
@@ -62,30 +58,33 @@ rand = { version = "0.8.4", default-features = false, optional = true }
...
+manta-accounting = { git = "https://github.com/manta-network/manta-rs.git", features = ["test"] }
+sp-core = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.16"}
+sp-io = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.16"}
 pallet-balances = { git = 'https://github.com/paritytech/substrate.git', branch = "polkadot-v0.9.16" }
 pallet-assets = { git = 'https://github.com/paritytech/substrate.git', branch = "polkadot-v0.9.16" }
 pallet-asset-manager = { path = "../asset-manager"}
-
+manta-util = { git = "https://github.com/manta-network/manta-rs.git"}
+manta-sdk = { git = "https://github.com/manta-network/sdk.git", features = ["download"] }
+tempfile = "3.3.0"
```
Garandor, line 89: i don't see which change needs tempfile and rand, could you explain?

pallets/manta-pay/Cargo.toml
```diff
@@ -62,30 +58,33 @@ rand = { version = "0.8.4", default-features = false, optional = true }
 tempfile = { version = "3.3.0", optional = true }
 
 # substrate dependencies
-frame-benchmarking = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.16", default-features = false, optional = true }
-frame-support = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.16", default-features = false }
-frame-system = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.16", default-features = false }
-sp-runtime = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.16", default-features = false }
+frame-benchmarking = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.16", default-features = false, optional = true }
```
Garandor, line 61: is making the `.git` explicit necessary? I'm thinking it's obviously a git link due to the `git = ` already

pallets/manta-pay/Cargo.toml
```diff
@@ -62,30 +58,33 @@ rand = { version = "0.8.4", default-features = false, optional = true }
 tempfile = { version = "3.3.0", optional = true }
```
Garandor, line 58: why do rand and tempfile have mentions in both deps and devdeps?

Garandor (changes requested):

pallets/manta-pay/Cargo.toml
Dengjianping: I'm not that familiar with manta-pay, but I suppose these test cases needs them.

pallets/manta-pay/Cargo.toml
Dengjianping: This is tricky. If I run `cargo b --release`, the `Cargo.lock` won't be updated. But while I run `cargo test` under `manta-pay`, it will be updated due to lacking of the tag `.git`.

pallets/manta-pay/Cargo.toml
Dengjianping: `tempfile` and `rand` are optional in `dev`. So they are not available while `cargo test` or `cargo bench`.

Garandor: Ok I see, tests and the bin target use both packages

pallets/manta-pay/Cargo.toml
Garandor: Ok i saw both being used in testcode
/resolve

pallets/manta-pay/Cargo.toml
Garandor: Don't really understand the different behavior since the short url should resolve to the .git one anyway, but there's no harm being explicit

Garandor (approved):
LGTM

Issue #459, opened 2022-03-24
Improve asset-manager genesis configurations


ghzlatarev:
Currently it is not possible to have asset-manager configurations at genesis, which is especially annoying during testing, as you have to manually register assets every time.
Have to wait for the below two PRs:
https://github.com/paritytech/polkadot/pull/5036
https://github.com/paritytech/polkadot/pull/4097

ghzlatarev:
We can add some checks to the current genesis configuration setup. Like whether the native asset id is less than the start of the non native assets. Or other sanity checks related to the AssetConfig trait.

PR #472, opened 2022-04-03
Improve asset manager

### LINKED ISSUE #440: Implement benchmarking macros in our pallet-asset-manager
Currently pallet-asset-manager is using 0 weights.
In order to be benchmarked, first the benchmarking framework needs to be implemented.
Benchmarking how-to:
https://www.notion.so/2d9269c2bb794ee0a1d05ab3ee3935ee?v=9a54b8caee9049e196c6d517776da318&p=eae0aab4425b4a04bd256b365d7a51fb
Afterwwards generate custom weight files for all runtimes.

stechu:
## Description

closes: #440 

- [x] Add asset mint function to asset-manager
- [x] Make settings of native asset id, dummy asset id, and start non-native asset id explicit
- [x] Add genesis config so that native asset will configured at genesis time
- [x] unit tests
- [x] Asset manager benchmarks
---

write a little note why.

- [x] Wrote unit tests.


pallets/asset-manager/src/lib.rs
```diff
@@ -90,30 +93,65 @@ pub mod pallet {
...
+		pub start_id: AssetId,
+		pub _marker: PhantomData<T>,
+	}
+
+	#[cfg(feature = "std")]
+	impl<T: Config> Default for GenesisConfig<T> {
+		fn default() -> Self {
+			Self {
+				start_id: 8u32,
+				_marker: PhantomData::<T>::default(),
```
bhgomes: You can just use `PhantomData` as the value here, or `Default::default()`

pallets/asset-manager/src/lib.rs
```diff
@@ -123,25 +161,27 @@ pub mod pallet {
 		LocationAlreadyExists,
 		/// Error creating asset, e.g. error returned from the implementation layer.
 		ErrorCreatingAsset,
-		/// Update a non-exist asset
+		/// Update a non-exist asset.
 		UpdateNonExistAsset,
 		/// Asset already registered.
 		AssetAlreadyRegistered,
+		/// Error on ming asset.
```
bhgomes: spelling

primitives/src/assets.rs
```diff
@@ -71,9 +72,28 @@ pub trait AssetRegistrar<T: AssetConfig> {
...
-	type AssetRegistrar: AssetRegistrar<Self>;
+pub trait AssetConfig<C>: 'static + Eq + Clone
+where
+	C: frame_system::Config,
+{
+	/// The AssetId that the non-native asset starts from.
+	/// A typical configuration is 8, so that asset 0 - 7 is reserved.
+	type StartNonNativeAssetId: Get<AssetId>;
+
+	/// Dummy Asset ID (used for transparent UTXO), a typical configuration is 0.
```
bhgomes: It's not necessarily used for transparent UTXO, it's a valuable feature on its own.

primitives/src/assets.rs
```diff
@@ -292,6 +317,9 @@ pub trait FungibleLedger<C>
 where
 	C: frame_system::Config,
 {
+	/// check if an asset id is valid
+	fn is_valid(self: Self, asset_id: AssetId) -> Result<(), FungibleLedgerConsequence>;
```
bhgomes: It should just be `self` not `self: Self`. Also, does it actually need to take `self` at all? And if so it should probably be by reference.

Since this method doesn't return a boolean I don't think it should be called `is_valid` instead something like `check_valid` or `check_validity`.

pallets/asset-manager/src/mock.rs
```diff
@@ -151,18 +167,141 @@ impl AssetRegistrar<MantaAssetConfig> for MantaAssetRegistrar {
 	}
 }
 
+pub struct MantaFungibleLedger;
+impl FungibleLedger<Runtime> for MantaFungibleLedger {
+	fn is_valid(self: Self, asset_id: AssetId) -> Result<(), FungibleLedgerConsequence> {
```
bhgomes: Just use `self` not `self: Self`.

runtime/dolphin/src/lib.rs
```diff
@@ -966,14 +966,9 @@ impl pallet_assets::Config for Runtime {
 	type WeightInfo = pallet_assets::weights::SubstrateWeight<Runtime>;
 }
 
-parameter_types! {
-	// Pallet account for record rewards and give rewards to collator.
-	pub const AssetManagerPalletId: PalletId = ASSET_MANAGER_PALLET_ID;
-}
-
 pub struct MantaAssetRegistrar;
 use frame_support::pallet_prelude::DispatchResult;
```
bhgomes, line 982: Lift import to top of file.

primitives/src/assets.rs
```diff
@@ -71,9 +72,28 @@ pub trait AssetRegistrar<T: AssetConfig> {
 	fn update_asset_metadata(asset_id: AssetId, metadata: T::StorageMetadata) -> DispatchResult;
 }
 
-pub trait AssetConfig: 'static + Eq + Clone {
-	/// The trait we use to register Assets
-	type AssetRegistrar: AssetRegistrar<Self>;
+pub trait AssetConfig<C>: 'static + Eq + Clone
```
bhgomes, line 80: Why does this require the implementer to be static and have a `Clone` and `Eq` implementation?

runtime/dolphin/src/lib.rs
```diff
@@ -1041,12 +1060,23 @@ impl pallet_asset_manager::Config for Runtime {
 
 pub struct MantaFungibleLedger;
 impl FungibleLedger<Runtime> for MantaFungibleLedger {
+	fn is_valid(self: Self, asset_id: AssetId) -> Result<(), FungibleLedgerConsequence> {
```
bhgomes: `self` parameter should not be required here

runtime/dolphin/src/lib.rs
```diff
@@ -1107,9 +1139,14 @@ impl FungibleLedger<Runtime> for MantaFungibleLedger {
 		beneficiary: &<Runtime as frame_system::Config>::AccountId,
 		amount: Balance,
 	) -> Result<(), FungibleLedgerConsequence> {
+		Self.is_valid(asset_id)?;
 		Self::can_deposit(asset_id, beneficiary, amount)?;
-		if asset_id == 0 {
-			let _ = <Balances as Currency<<Runtime as frame_system::Config>::AccountId>>::deposit_creating(beneficiary, amount);
+		if asset_id == <MantaAssetConfig as AssetConfig<Runtime>>::NativeAssetId::get() {
+			let _ =
```
bhgomes: Why ignore the error?

pallets/manta-pay/src/mock.rs
```diff
@@ -137,12 +142,23 @@ impl pallet_assets::Config for Test {
 
 pub struct MantaFungibleLedger;
 impl FungibleLedger<Test> for MantaFungibleLedger {
+	fn is_valid(self: Self, asset_id: AssetId) -> Result<(), FungibleLedgerConsequence> {
```
bhgomes: `self` should not be needed

bhgomes (changes requested):
Some minor design issues I have with the API's. Also, the three implementations of the asset manager look very similar to each other if not identical. We should refactor them into one implementation that can be called by the three.

pallets/asset-manager/src/lib.rs
```diff
@@ -292,6 +332,45 @@ pub mod pallet {
 			});
 			Ok(())
 		}
+
+		/// Mint asset by its asset id to a beneficiary.
+		///
+		/// * `origin`: Caller of this extrinsic, the acess control is specfied by `ForceOrigin`.
```
ghzlatarev: spelling - `access control is specified` 
in a few places.

pallets/asset-manager/src/lib.rs
```diff
@@ -292,6 +332,45 @@ pub mod pallet {
...
+		/// * `asset_id`: AssetId to be updated.
+		/// * `beneficiary`: Account to mint the asset.
+		/// * `amount`: Amount of asset being minted.
+		/// # <weight>
+		/// TODO: get actual weight
+		/// # </weight>
+		#[pallet::weight(50_000_000)]
+		#[transactional]
+		pub fn mint_asset(
```
ghzlatarev, line 344: It is not immediately clear to me why we need this as a callable extrinsic 
The PR description doesn't elaborate as well.

pallets/asset-manager/src/lib.rs
```diff
@@ -90,30 +93,65 @@ pub mod pallet {
...
+	pub struct GenesisConfig<T: Config> {
+		pub start_id: AssetId,
+		pub _marker: PhantomData<T>,
+	}
+
+	#[cfg(feature = "std")]
+	impl<T: Config> Default for GenesisConfig<T> {
+		fn default() -> Self {
+			Self {
+				start_id: 8u32,
```
ghzlatarev: Why 8 ?

pallets/asset-manager/src/mock.rs
```diff
@@ -151,18 +167,141 @@ impl AssetRegistrar<MantaAssetConfig> for MantaAssetRegistrar {
...
+	}
+
+	fn can_deposit(
+		asset_id: AssetId,
+		account: &<Runtime as frame_system::Config>::AccountId,
+		amount: Balance,
+	) -> Result<(), FungibleLedgerConsequence> {
+		Self.is_valid(asset_id)?;
+		if asset_id == <MantaAssetConfig as AssetConfig<Runtime>>::NativeAssetId::get() {
+			// we assume native asset with id 0
```
ghzlatarev: This comment is no longer accurate. Other places as well.

pallets/asset-manager/src/tests.rs
```diff
@@ -189,3 +212,44 @@ fn update_asset() {
...
+			min_balance: 1u128,
+			evm_address: None,
+			is_frozen: false,
+			is_sufficient: true,
+		};
+		let source_location = AssetLocation(VersionedMultiLocation::V1(MultiLocation::parent()));
+		assert_ok!(AssetManager::register_asset(
+			Origin::root(),
+			source_location.clone(),
+			asset_metadata.clone()
```
ghzlatarev: no need for these clones

primitives/src/assets.rs
ghzlatarev, line 80: you forgot to implement for xcm-mock so the tests are still failing

pallets/asset-manager/src/lib.rs
Dengjianping: Yeah, and just putting `PhantomData` here should compile as well, compiler will infer the type.

pallets/asset-manager/src/lib.rs
```diff
@@ -292,6 +332,45 @@ pub mod pallet {
...
+		/// Mint asset by its asset id to a beneficiary.
+		///
+		/// * `origin`: Caller of this extrinsic, the acess control is specfied by `ForceOrigin`.
+		/// * `asset_id`: AssetId to be updated.
+		/// * `beneficiary`: Account to mint the asset.
+		/// * `amount`: Amount of asset being minted.
+		/// # <weight>
+		/// TODO: get actual weight
+		/// # </weight>
+		#[pallet::weight(50_000_000)]
```
Dengjianping: Please add an associated type `WeightInfo` to `Config`, and then benchmark this pallet, finally update this weight.

ghzlatarev:
@stechu  If you're going to implement the asset manager benchmark in this PR, please link to https://github.com/Manta-Network/Manta/issues/440 in the description

stechu:
> Some minor design issues I have with the API's. Also, the three implementations of the asset manager look very similar to each other if not identical. We should refactor them into one implementation that can be called by the three.

Will abstract the code to a trait.

runtime/dolphin/src/lib.rs
stechu: The returned type is not really an error, instead it is an over-engineered `Imbalance` type (at least in my view). As the result, the error handling logic will be quite complicated.  Also, in our setting, this `mint` function will only be used by SUDO (a.k.a. governance). I will put a FIXME here tho.

pallets/asset-manager/src/lib.rs
stechu: Mostly for testing. The idea is we would like to do almost everything in asset-manager for asset related operations, regardless it is a native asset or non-native asset.

ghzlatarev:
This PR partially relates to this issue as well https://github.com/Manta-Network/Manta/issues/459

pallets/asset-manager/Cargo.toml
```diff
@@ -37,14 +38,18 @@ std = [
 	"sp-std/std",
 	"manta-primitives/std",
 	"frame-benchmarking/std",
+	"xcm/std",
+	"manta-primitives/std",
```
Dengjianping: Duplicated.

pallets/asset-manager/Cargo.toml
```diff
@@ -37,14 +38,18 @@ std = [
 	"sp-std/std",
 	"manta-primitives/std",
 	"frame-benchmarking/std",
+	"xcm/std",
```
Dengjianping: I think `xcm/std` and `frame-benchmarking/std` can be removed.
We don't need both in std feature.

runtime/dolphin/src/tests/xcm_mock/parachain.rs
```diff
@@ -537,14 +533,38 @@ impl AssetRegistrar<MantaAssetConfig> for MantaAssetRegistrar {
...
 	type AssetRegistrar = MantaAssetRegistrar;
+	type DummyAssetId = DummyAssetId;
+	type NativeAssetId = NativeAssetId;
+	type StartNonNativeAssetId = StartNonNativeAssetId;
+	type NativeAssetLocation = NativeAssetLocation;
+	type NativeAssetMetadata = NativeAssetMetadata;
+	type FungibleLedger = ConcreteFungibleLedger<Runtime, MantaAssetConfig, Balances, Assets>;
 }
 
 impl pallet_asset_manager::Config for Runtime {
```
ghzlatarev, line 570: @stechu  you need `type WeightInfo = ();` here to pass all tests.

Dengjianping (approved):
LGTM!

primitives/src/assets.rs
stechu: probably not, I think I simply copied type definition of `frame_system::Config`.

bhgomes (approved):

Issue #185, opened 2021-09-06
Use pallet id as treasury account to save xcm fee


Dengjianping:
We can take acala as an example.
```rust
pub struct ToTreasury;
impl TakeRevenue for ToTreasury {
	fn take_revenue(revenue: MultiAsset) {
		if let MultiAsset::ConcreteFungible { id, amount } = revenue {
			if let Some(currency_id) = CurrencyIdConvert::convert(id) {
				// ensure KaruraTreasuryAccount have ed for all of the cross-chain asset.
				// Ignore the result.
				let _ = Currencies::deposit(currency_id, &KaruraTreasuryAccount::get(), amount);
			}
		}
	}
}
```
https://github.com/AcalaNetwork/Acala/blob/master/runtime/karura/src/lib.rs#L1409

PR #473, opened 2022-04-04
Calamari XCM

### LINKED ISSUE #446: XCM integration testing with Moonbeam
- [x] Generate binary with xcm code
- [x] Generate genesis state/wasm
- [x] Setup parachain nodes on Terraform
- [x] Deploy and sync with relay chain
- [x] Pass state/wasm to Alberto
- [x] Create parachain sovereign account and give to Alberto to fund
- [x] Get HRMP chanel request encode data
- [x] Send xcm transact message to relay chian with the encoded data
- [x] Moonbase alpha needs to accept
- [x] Moonbase alpha does the same and we accept
- [x] Register each other's assets
- [x] Test transfers
- [x] Repeat with Calamari

ghzlatarev:
## Description

closes: #446

* Integrate XCM for Calamari
* Calamari mock-xcm tests

---

write a little note why.

- [x] Wrote unit tests.


runtime/calamari/src/lib.rs
```diff
@@ -696,22 +845,52 @@ pub type Barrier = (
 	AllowSubscriptionsFrom<ParentOrSiblings>,
 );
 
-pub struct XcmConfig;
-impl Config for XcmConfig {
+parameter_types! {
+	/// Xcm fees will go to the asset manager (we don't implement treasury yet)
+	pub XcmFeesAccount: AccountId = AssetManager::account_id();
```
Garandor, line 850: If fees accrue to the asset manager account (which if i see the code right is just converted from the pallet id?), who can access them in the future? It doesn't seem like asset-manager itself has a transfer extrinsic.

Is this going to be a governance force_transfer() type of tx to transfer to treasury some time later?

Garandor (commented):
I looked it over and LGTM, but i'm not knowledgeable enough in XCM code to approve this merge.

runtime/calamari/src/lib.rs
stechu: good catch. Since we have treasury, it should go to treasury actually.

ghzlatarev: Yeah the root through governance can force_transfer .
Otherwise I'm not sure if we have access to this account directly.

stechu (approved):

runtime/calamari/src/lib.rs
Garandor: Umm, @ghzlatarev I think you didn't change this to treasury before merging.
What's the reason you prefered it to stay the PalletId account?

I'm randomly looking at this again since I just found an issue by @Dengjianping describing this 
https://github.com/Manta-Network/Manta/issues/185

ghzlatarev: I think I missed that request, can you change it in your PR https://github.com/Manta-Network/Manta/pull/485

PR #485, opened 2022-04-13
Send XCM fees to Treasury & refactor RTs to use PalletIDs from constants

### LINKED ISSUE #185: Use pallet id as treasury account to save xcm fee

Garandor:
## Description
1. Point XCM Fees to Treasury account on KMA and DOL. Closes #185
2. MA KMA runtimes used hardcoded strings for PalletIDs instead of the constants from manta_primitives
Minor cleanup so no attached issue 
---

write a little note why.

- [ ] Wrote unit tests.

### DIFF
--- a/runtime/calamari/src/lib.rs
+++ b/runtime/calamari/src/lib.rs
@@ -61,7 +61,9 @@ use manta_primitives::{
 		AssetConfig, AssetIdLocationConvert, AssetLocation, AssetRegistrar, AssetRegistrarMetadata,
 		AssetStorageMetadata, ConcreteFungibleLedger,
 	},
-	constants::{time::*, ASSET_MANAGER_PALLET_ID, CALAMARI_DECIMAL},
+	constants::{
+		time::*, ASSET_MANAGER_PALLET_ID, CALAMARI_DECIMAL, STAKING_PALLET_ID, TREASURY_PALLET_ID,
+	},
 	prod_or_fast,
 	types::{AccountId, AssetId, AuraId, Balance, BlockNumber, Hash, Header, Index, Signature},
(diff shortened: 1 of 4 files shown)


ghzlatarev (commented):
can you also change `pub XcmFeesAccount:` configurations to the Treasury ID, please. 
I missed it in the Calamari XCM pr.
Can clone this issue https://github.com/Manta-Network/Manta/issues/185 @Garandor

ghzlatarev (approved):

runtime/calamari/src/lib.rs
```diff
@@ -846,8 +848,7 @@ pub type Barrier = (
 );
 
 parameter_types! {
-	/// Xcm fees will go to the asset manager (we don't implement treasury yet)
-	pub XcmFeesAccount: AccountId = AssetManager::account_id();
+	pub XcmFeesAccount: AccountId = Treasury::account_id();
```
Dengjianping, line 851: Do we need delete `AssetManager::account_id()` in storage migration?

Dengjianping (approved):
LGTM!

PR #509, opened 2022-04-19
`OnRuntimeUpgrade` hook for `AssetManager` to properly set `NativeAssetId` and `StartNonNativeAssetId`

### LINKED ISSUE #498: OnRuntimeUpgrade hook for Calamari AssetManager's `native-asset-id` and `start-non-native-asset-id`
Since we only just added the AssetManager pallet to Calamari the genesis configurations were not executed.
We need to set the native asset id and non-native asset id start to 1 and 8 respectively. 
As they are configured in the runtime configuration.

ghzlatarev:

## Description

closes: #498

* Use the genesis configuration code of `AssetManager` in a `OnRuntimeUpgrade` hook, as this pallet manager was not in Calamari's genesis and the code wasn't invoked.

---

write a little note why.

- [x] Wrote unit tests.

### DIFF
--- a/pallets/asset-manager/src/lib.rs
+++ b/pallets/asset-manager/src/lib.rs
@@ -214,6 +214,20 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type UnitsPerSecond<T: Config> = StorageMap<_, Blake2_128Concat, AssetId, u128>;
 
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_runtime_upgrade() -> Weight {
+			NextAssetId::<T>::set(<T::AssetConfig as AssetConfig<T>>::StartNonNativeAssetId::get());
+			let asset_id = <T::AssetConfig as AssetConfig<T>>::NativeAssetId::get();
+			let metadata = <T::AssetConfig as AssetConfig<T>>::NativeAssetMetadata::get();
+			let location = <T::AssetConfig as AssetConfig<T>>::NativeAssetLocation::get();
+			AssetIdLocation::<T>::insert(&asset_id, &location);
+			AssetIdMetadata::<T>::insert(&asset_id, &metadata);
+			LocationAssetId::<T>::insert(&location, &asset_id);
+			T::DbWeight::get().reads_writes(1, 4)
+		}
+	}
+
 	#[pallet::call]
 	impl<T: Config> Pallet<T> {
 		/// Register a new asset in the asset manager.
--- a/pallets/asset-manager/src/lib.rs
+++ b/pallets/asset-manager/src/lib.rs
@@ -400,7 +400,6 @@ pub mod pallet {
 		pub fn set_genesis_configuration() -> Weight {
 			let mut weight: Weight = T::DbWeight::get().reads(1);
 			if have_storage_value(Self::name().as_bytes(), b"NextAssetId", &[]) {
-				weight = weight.saturating_add(T::DbWeight::get().reads(1));
 				log::warn!(" !!! Aborting, NextAssetId was already set.");
 				return weight;
 			}
--- a/pallets/asset-manager/src/lib.rs
+++ b/pallets/asset-manager/src/lib.rs
@@ -216,82 +216,6 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type UnitsPerSecond<T: Config> = StorageMap<_, Blake2_128Concat, AssetId, u128>;
 
-	#[pallet::hooks]
-	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
-		fn on_runtime_upgrade() -> Weight {
-			let mut weight: Weight = T::DbWeight::get().reads(1);
-			if have_storage_value(Self::name().as_bytes(), b"NextAssetId", &[]) {
-				weight = weight.saturating_add(T::DbWeight::get().reads(1));
-				log::warn!(" !!! Aborting, NextAssetId was already set.");
-				return weight;
-			}
-
-			weight = weight.saturating_add(T::DbWeight::get().reads(4));
-			let asset_id = <T::AssetConfig as AssetConfig<T>>::NativeAssetId::get();
-			let metadata = <T::AssetConfig as AssetConfig<T>>::NativeAssetMetadata::get();
-			let location = <T::AssetConfig as AssetConfig<T>>::NativeAssetLocation::get();
-			if have_storage_value(
-				Self::name().as_bytes(),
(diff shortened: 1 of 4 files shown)


pallets/asset-manager/src/lib.rs
```diff
@@ -214,6 +214,20 @@ pub mod pallet {
...
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_runtime_upgrade() -> Weight {
+			NextAssetId::<T>::set(<T::AssetConfig as AssetConfig<T>>::StartNonNativeAssetId::get());
+			let asset_id = <T::AssetConfig as AssetConfig<T>>::NativeAssetId::get();
+			let metadata = <T::AssetConfig as AssetConfig<T>>::NativeAssetMetadata::get();
+			let location = <T::AssetConfig as AssetConfig<T>>::NativeAssetLocation::get();
+			AssetIdLocation::<T>::insert(&asset_id, &location);
+			AssetIdMetadata::<T>::insert(&asset_id, &metadata);
+			LocationAssetId::<T>::insert(&location, &asset_id);
+			T::DbWeight::get().reads_writes(1, 4)
```
Dengjianping: Why `(1, 4)`?
I can see there're 4 reads, 3 writes.

ghzlatarev: There are 3 inserts and 1 set, so 4 writes
There is 1 get for `NextAssetId`, the others gets are just for runtime constants, not actual storage reads.

ghzlatarev: Actually you're right there are 4 storage reads but 4 storage writes as well.

pallets/asset-manager/src/lib.rs
```diff
@@ -214,6 +214,20 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type UnitsPerSecond<T: Config> = StorageMap<_, Blake2_128Concat, AssetId, u128>;
 
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_runtime_upgrade() -> Weight {
+			NextAssetId::<T>::set(<T::AssetConfig as AssetConfig<T>>::StartNonNativeAssetId::get());
```
Garandor: as a best-practice, maybe check this is actually unset first. If we were to forget to remove this hook before the upgrade-after-next it'll reset this even if assets have been registered in the meantime.

The inserts are uncritical since they can implicitly update if the element already exists

Garandor: Also, please add a log::info() message indicating so we can observe it was executed

ghzlatarev: Ok, I made it so if any of the items is set we abort. Something has gone wrong in that case.

pallets/asset-manager/src/lib.rs
```diff
@@ -214,6 +216,75 @@ pub mod pallet {
...
+			let asset_id = <T::AssetConfig as AssetConfig<T>>::NativeAssetId::get();
+			let metadata = <T::AssetConfig as AssetConfig<T>>::NativeAssetMetadata::get();
+			let location = <T::AssetConfig as AssetConfig<T>>::NativeAssetLocation::get();
+			if have_storage_value(
+				Self::name().as_bytes(),
+				b"AssetIdLocation",
+				&Blake2_128Concat::hash(&asset_id.encode()),
+			) {
+				log::warn!(" !!! Aborting, AssetIdLocation was already set.");
+				return T::DbWeight::get().reads_writes(4, 0);
```
Garandor: you should be at 5 reads here ( and +1 at each one below )
Otherwise LGTM :+1:

Garandor:
Just for an FYI - Parity discourages using onruntimeupgrade in pallets, see this post: https://substrate.stackexchange.com/a/2030/504

It's not an issue here I, but worth keeping in mind for the future to implement it analogous to 
https://github.com/Manta-Network/Manta/blob/manta/runtime/calamari/src/lib.rs#L1159

pallets/asset-manager/src/lib.rs
Garandor: Check https://github.com/paritytech/cumulus/blob/master/pallets/parachain-system/src/migration.rs#L37
for how parity avoid messing up the counting

ghzlatarev: Should be resolved now.

ghzlatarev:
> Just for an FYI - Parity discourages using onruntimeupgrade in pallets, see this post: https://substrate.stackexchange.com/a/2030/504
> 
> It's not an issue here I, but worth keeping in mind for the future to implement it analogous to https://github.com/Manta-Network/Manta/blob/manta/runtime/calamari/src/lib.rs#L1159

Yeah thanks for pointing out this anti-pattern. I went ahead with the runtime level hook.

pallets/asset-manager/src/lib.rs
```diff
@@ -394,6 +396,79 @@ pub mod pallet {
 		pub fn account_id() -> T::AccountId {
 			T::PalletId::get().into_account()
 		}
+
+		pub fn set_genesis_configuration() -> Weight {
+			let mut weight: Weight = T::DbWeight::get().reads(1);
+			if have_storage_value(Self::name().as_bytes(), b"NextAssetId", &[]) {
+				weight = weight.saturating_add(T::DbWeight::get().reads(1));
```
Garandor: since you start with 1R in L401, I think this line shouldn't be here

ghzlatarev: indeed

Garandor (approved):
LGTM

pallets/asset-manager/src/lib.rs
```diff
@@ -394,6 +396,78 @@ pub mod pallet {
...
+			let asset_id = <T::AssetConfig as AssetConfig<T>>::NativeAssetId::get();
+			let metadata = <T::AssetConfig as AssetConfig<T>>::NativeAssetMetadata::get();
+			let location = <T::AssetConfig as AssetConfig<T>>::NativeAssetLocation::get();
+			if have_storage_value(
+				Self::name().as_bytes(),
+				b"AssetIdLocation",
+				&Blake2_128Concat::hash(&asset_id.encode()),
+			) {
+				log::warn!(" !!! Aborting, AssetIdLocation was already set.");
+				return weight;
```
Dengjianping, line 417: If this branch happens, the rest of code will not be executed. Is it expected?

ghzlatarev: Yes, we haven't done any writes so far so we can exit early.

Dengjianping (approved):
LGTM!

PR #511, opened 2022-04-21
SBP M3 review

stiiifff:


stechu:
@stiiifff  Thanks! We will bump in the next major version.

PR #529, opened 2022-05-03
feat: add ledger sync protocol

bhgomes:
## Description

This PR adds the `pull` method for use by wallets to request the diff from the latest state.

### Unit Test Spec

Use `sudo` to set the storage to random UTXO and void numbers: for each round add random number (1 to 2) of random `(UTXO, EncryptedNote)` pairs and a random number (0 to 2) of random `VoidNumber`s:
- `Shards`: Add random `(UTXO, EncryptedNote)` pairs with a random first key and increasing second key
- `InsertionOrderedVoidNumbers`: Add random `VoidNumber`s with increasing key

Checkpoint Description:
- `receiver_index: [usize; 256]`: Lengths of each first key in `Shards`
- `sender_index: usize`: Length of `InsertionOrderedVoidNumber`

Testing correctness of Normal Usage:
1. Add random elements to ledger
2. Select a random checkpoint that's below the current one
3. Pull the latest state and check that it returns the data items (in order!) up to the checkpoint returned in the diff
4. If `should_continue == true` then `pull` should be called again with the returned checkpoint (GOTO step 3)
5. If it passes GOTO step 1

Testing correctness of Abnormal Usage: Same as the previous test except try to get from a checkpoint _later_ than the current one. The `pull` should return nothing.

### TODO

- [x] Implement diffing algorithm
- [x] Add RPC interface
- [ ] Add Unit tests / Simulation testing

---

write a little note why.

- [ ] Wrote unit tests.


pallets/collator-selection/src/migrations.rs
```diff
@@ -87,7 +88,7 @@ impl<T: Config> Pallet<T> {
 
 	pub fn post_migrate_v0_to_v1() -> Result<(), &'static str> {
 		use frame_support::migration::{have_storage_value, storage_key_iter};
-		if !(Self::on_chain_storage_version() == 1) {
+		if Self::on_chain_storage_version() != 1 {
```
stechu: I think this migration code should be deleted. @Garandor

Garandor: This migration code has been applied to Calamari and is no longer in use. It is being preserved as 
- migration history
- Reference template for future migrations

This code shouldn't need to be touched, so unless you're trying to clear clippy warnings or the like, please just drop this change. If the change helps to keep our build logs cleaner, feel free to merge it as is.

I'm marking this as resolved for now. Feel free to reopen if it needs more discussion @bhgomes

bhgomes: Yeah, I was just clearing `clippy` warnings.

Dengjianping:
When runtime team integrate v0.9.22(jsonrpsee), we better mark storage method as `blocking`.
But not there's any rpc method like this in manta-pay.
https://github.com/paritytech/substrate/pull/11459

stechu:
> When runtime team integrate v0.9.22(jsonrpsee), we better mark storage method as `blocking`. But not there's any rpc method like this in manta-pay. [paritytech/substrate#11459](https://github.com/paritytech/substrate/pull/11459)

Indeed. Some context here: https://github.com/paritytech/substrate/issues/11401

pallets/manta-pay/src/lib.rs
```diff
@@ -480,19 +465,140 @@ pub mod pallet {
...
+		/// Maximum Size of Sender Data Update
+		const PULL_MAX_SENDER_UPDATE_SIZE: usize = 1024;
+
+		/// Pulls receiver data from the ledger starting at the `receiver_index`.
 		#[inline]
-		fn ledger() -> Ledger<T> {
-			Ledger(PhantomData)
+		fn pull_receivers(receiver_index: &mut [usize; 256]) -> (bool, ReceiverChunk) {
+			let mut more_receivers = false;
+			let mut receivers = Vec::new();
```
Dengjianping, line 477: Can we create it like this `Vec::with_capacity(Self::PULL_MAX_PER_SHARD_UPDATE_SIZE)`?
Then `receivers.shrink_to_fit()` if it doesn't reach the reserved length.

pallets/manta-pay/src/lib.rs
```diff
@@ -480,19 +465,140 @@ pub mod pallet {
...
+					_ => return false,
+				}
+			}
+			iter.next().is_some()
+		}
+
+		/// Pulls sender data from the ledger starting at the `sender_index`.
+		#[inline]
+		fn pull_senders(sender_index: &mut usize) -> (bool, SenderChunk) {
+			let mut senders = Vec::new();
```
Dengjianping, line 513: Same as `receivers`.

Dengjianping (commented):
LGTM!

pallets/manta-pay/src/lib.rs
stechu: This is not a good idea for the typical use case, when the actual receiver size is small.

pallets/collator-selection/src/lib.rs
```diff
@@ -614,13 +614,11 @@ pub mod pallet {
 			kick_candidates.iter().for_each(|(acc_id, my_blocks_this_session)| {
 				if *my_blocks_this_session < evict_below_blocks {
 					// If our validator is not also a candidate we're invulnerable or already kicked
-					if let Some(_) = candidates.iter().find(|&x|{ x.who == *acc_id })
-					{
+					if candidates.iter().any(|x| x.who == *acc_id) {
 						Self::try_remove_candidate(acc_id)
-							.and_then(|_| {
+							.map(|_| {
```
Garandor: nit: Instead of using `map` here, I prefer the semantically more clear `and_then`, although it's more verbose.
If it's a clippy warning, I'm fine yielding to our machine overlords.

primitives/src/assets.rs
```diff
@@ -444,7 +444,7 @@ where
 			<Native as Currency<C::AccountId>>::deposit_creating(beneficiary, amount);
 		} else {
 			<NonNative as Mutate<C::AccountId>>::mint_into(asset_id, beneficiary, amount)
-				.map_err(|e| FungibleLedgerError::InvalidMint(e))?;
+				.map_err(FungibleLedgerError::InvalidMint)?;
```
Garandor, line 447: out of curiosity, why is this random enum convertible to a FnOnce(e)?

primitives/Cargo.toml
```diff
@@ -17,7 +17,7 @@ smallvec = "1.8.0"
 log = "0.4.16"
 
 # manta-rs dependencies
-manta-accounting = { git = "https://github.com/manta-network/manta-rs.git", default-features = false }
+manta-accounting = { git = "https://github.com/manta-network/manta-rs.git", rev="d21b0af", default-features = false }
```
Garandor: Is rev stapling necessary over tagging?
All I can think of is development is too rapid to be worth spending time adding a tag each time, although it's just a two-liner ( plus adding protection on vX.Y.Z tags in github )
`
git tag v0.0.1
git push v0.0.1
`

finding and navigating tagged verison on github is much nicer than finding a state by rev hash

node/src/chain_specs/dolphin.rs
```diff
@@ -16,10 +16,7 @@
 
 use super::*;
 use crate::command::DOLPHIN_PARACHAIN_ID;
-
-use dolphin_runtime::{
-	AssetManagerConfig, CouncilConfig, DemocracyConfig, GenesisConfig, TechnicalCommitteeConfig,
-};
+use dolphin_runtime::{CouncilConfig, DemocracyConfig, TechnicalCommitteeConfig};
```
Garandor: instead of removing uses, I'd rather add them all here so that the `dolphin_runtime` qualifier isn't needed in actual spec code, making it identical over all runtimes ( makes it easier to diff for relevant changes between the specs )

Same for calamari and manta

pallets/manta-pay/src/lib.rs
```diff
@@ -480,19 +465,140 @@ pub mod pallet {
...
+				}
+			}
+			(iter.next().is_some(), senders)
 		}
 
-		/// The account ID of AssetManager
+		/// Returns the update required to be synchronized with the ledger starting from
+		/// `checkpoint`.
+		#[inline]
+		pub fn pull(mut checkpoint: Checkpoint) -> PullResponse {
```
Garandor: I had to look through the code to understand what pull does, it's not obvious from the signature.
Can we rename it to something more concrete like `request_ledger_diff_from` or `fetch_ledger_updates_since`

pallets/manta-pay/src/lib.rs
```diff
@@ -132,34 +142,33 @@ pub mod pallet {
...
-	pub(super) type VoidNumberSet<T: Config> = StorageMap<_, Identity, [u8; 32], (), ValueQuery>;
+	pub(super) type VoidNumberSet<T: Config> = StorageMap<_, Identity, VoidNumber, (), ValueQuery>;
 
-	/// Void number set insertion order
-	/// Each element of the key is an `u64` insertion order number of void number.
+	/// Void Number Ordered by Insertion
 	#[pallet::storage]
-	pub(super) type VoidNumberSetInsertionOrder<T: Config> =
-		StorageMap<_, Identity, u64, [u8; 32], ValueQuery>;
+	pub(super) type InsertionOrderedVoidNumbers<T: Config> =
```
Garandor: this name change will need a migration on dolphin.
Unless we restart the testnet with this release.

pallets/manta-pay/src/lib.rs
```diff
@@ -480,19 +465,140 @@ pub mod pallet {
...
+		/// Posts the transaction encoded in `post` to the ledger, using `sources` and `sinks` as
+		/// the public deposit and public withdraw accounts respectively.
+		#[inline]
+		fn post_transaction(
+			origin: Option<T::AccountId>,
+			sources: Vec<T::AccountId>,
+			sinks: Vec<T::AccountId>,
+			post: TransferPost,
+		) -> DispatchResultWithPostInfo {
+			Self::deposit_event(
```
Garandor, line 560: Did post_transaction go through some precheck before arriving here?
If not, we don't want to call deposit_event() in the failure case ~~, as spamming the node with invalid txs could be a DDoS vector, as every event carries overhead.~~

Normal modus operandi is to deposit an event **on success only**

pallets/manta-pay/Cargo.toml
```diff
@@ -67,13 +87,22 @@ sp-std = { git = 'https://github.com/paritytech/substrate.git', branch = "polkad
 scale-codec = { package = "parity-scale-codec", version = '3.0.0', default-features = false, features = ["derive", "max-encoded-len"] }
 scale-info = { version = "2.0.0", default-features = false, features = ["derive"] }
 
+# TODO: investigate whether these two are needed
+sp-api = { git = 'https://github.com/paritytech/substrate.git', branch = "polkadot-v0.9.18", optional = true, default-features = false }
+sp-blockchain = { git = 'https://github.com/paritytech/substrate.git', branch = "polkadot-v0.9.18", optional = true, default-features = false }
```
ghzlatarev, line 92: Let's log this TODO or resolve it.

pallets/manta-pay/src/lib.rs
```diff
@@ -480,19 +465,140 @@ pub mod pallet {
...
+
+		/// Pulls receiver data from the ledger starting at the `receiver_index`.
 		#[inline]
-		fn ledger() -> Ledger<T> {
-			Ledger(PhantomData)
+		fn pull_receivers(receiver_index: &mut [usize; 256]) -> (bool, ReceiverChunk) {
+			let mut more_receivers = false;
+			let mut receivers = Vec::new();
+			for (i, index) in receiver_index.iter_mut().enumerate() {
+				more_receivers |= Self::pull_receivers_for_shard(i as u8, index, &mut receivers);
```
Garandor, line 479: while this is compact and *clever* logic, I was wondering where `receivers` actually gets filled.
I'd prefer a simple multiline solution over this compact one

Can `pull_receivers_for_shard` not return an `Option<ReceiverChunk>`?

pallets/manta-pay/Cargo.toml
```diff
@@ -67,13 +87,22 @@ sp-std = { git = 'https://github.com/paritytech/substrate.git', branch = "polkad
 scale-codec = { package = "parity-scale-codec", version = '3.0.0', default-features = false, features = ["derive", "max-encoded-len"] }
 scale-info = { version = "2.0.0", default-features = false, features = ["derive"] }
 
+# TODO: investigate whether these two are needed
```
Garandor: is this a TODO for later?

pallets/manta-pay/src/lib.rs
```diff
@@ -132,34 +142,33 @@ pub mod pallet {
...
-	/// Each element of the key is an `u64` insertion order number of void number.
+	/// Void Number Ordered by Insertion
 	#[pallet::storage]
-	pub(super) type VoidNumberSetInsertionOrder<T: Config> =
-		StorageMap<_, Identity, u64, [u8; 32], ValueQuery>;
+	pub(super) type InsertionOrderedVoidNumbers<T: Config> =
+		StorageMap<_, Identity, u64, VoidNumber, ValueQuery>;
 
 	/// The size of Void Number Set
 	/// FIXME: this should be removed.
```
Garandor: VoidNumberSetSize is in use, why remove?

pallets/asset-manager/src/benchmarking.rs
```diff
@@ -47,21 +47,17 @@ benchmarks! {
...
-
 			Pallet::<T>::register_asset(RawOrigin::Root.into(), location.clone(), metadata.clone())?;
 			Pallet::<T>::set_units_per_second(RawOrigin::Root.into(), i, 0)?;
 		}
-
 		// does not really matter what we register, as long as it is different than the previous
 		let location = <T::AssetConfig as AssetConfig<T>>::AssetLocation::default();
 		let metadata = <T::AssetConfig as AssetConfig<T>>::AssetRegistrarMetadata::default();
 		let amount = 10;
-		Pallet::<T>::register_asset(RawOrigin::Root.into(), location.clone(), metadata.clone())?;
```
ghzlatarev, line 63: Btw all of these clones() in the setup and verification are not part the benchmark itself, so they didn't impact the performance.

Garandor (changes requested):
Partial review, I did not look at node and tests changes yet.

As a note for the future, there are a lot of unrelated clippy/style improvements (e.g. in collator-selection) interspersed with the feature changes

In a big PR like this, I'd recommend to separate these out into a separate "cleanup" PR to contain the scope of changes and in this case, make the actual rpc change easier to locate and reason about.

@Dengjianping : `manta-pay` looks like it's gonna need a rewrite for the 0. 9.22 jsonrpsee migration.

primitives/src/assets.rs
```diff
@@ -474,6 +474,6 @@ where
 			)
 			.map(|_| ())
 		}
-		.map_err(|e| FungibleLedgerError::InvalidTransfer(e))
```
ghzlatarev, line 477: Why not propagate the dispatch error ? It will have useful info.

Dengjianping:
> 

Correct, I just added a comment to #571.

primitives/src/assets.rs
Dengjianping: > Recall from the [“Enum values”](https://doc.rust-lang.org/stable/book/ch06-01-defining-an-enum.html#enum-values) section of Chapter 6 that the name of each enum variant that we define also becomes an initializer function. We can use these initializer functions as function pointers that implement the closure traits, which means we can specify the initializer functions as arguments for methods that take closures

https://doc.rust-lang.org/stable/book/ch19-05-advanced-functions-and-closures.html#function-pointers

This might be the answer.

Garandor: Huh, another cool quality-of-dev-life improvement over C.
Thanks Jamie and thanks Rust team :)

pallets/manta-pay/src/lib.rs
Dengjianping: If this storage has been used in dolphin, we better do storage migration, and tx version might be bumped as well.

node/src/rpc/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+mod dolphin;
+
+pub use common::Common;
+pub use dolphin::Dolphin;
+
+/// RPC Extension Type
+pub type RpcExtension = jsonrpc_core::IoHandler<sc_rpc::Metadata>;
+
+/// RPC Extension Builder
+pub struct Builder<C, P, M = common::Common> {
```
Garandor, line 32: is common a sensible default here? Shouldn't it be one of the concrete markers like `manta`? Not sure we'd ever want to build a common-only extension

Dengjianping:
@Garandor You just reminded me that we might do storage migration for manta-pay, but I didn't find the feature `try-runtime` in manta-pay. 

Can we add this feature to manta-pay?

primitives/Cargo.toml
stechu: I think this is a mistake. I forget to clean this one. thanks for pointing out.

pallets/collator-selection/src/lib.rs
stechu: agreed. Not my change tho :)
(reverted)

pallets/manta-pay/Cargo.toml
stechu: removed.

pallets/manta-pay/src/lib.rs
stechu: That was a legacy issue: I thought the storage map itself can track how many element it stored. It turned out not the case.

pallets/manta-pay/src/lib.rs
stechu: reverted.

Garandor:
Finished looking through the rest of the code.

The node looks great, thanks for the extensible design and all your refactoring of the messy code in service @bhgomes and/or @stechu 
We'll soon get rid of all the aura-cumulus code in there as well and then this file will have some semblance of readability.

Garandor:
> Can we add this feature to manta-pay?

Agree, it should be added along with the migration.
And migrating is mandatory, otherwise the new void numbers storage will be empty and ... all spent UTXOs become spendable again?

pallets/manta-pay/src/lib.rs
Garandor: Substrate has a `CountedStorageMap` for that

https://github.com/paritytech/substrate/blob/7a09b82772b0597cdae224f4b124d660dda9e82b/frame/support/src/storage/types/counted_map.rs#L49

node/src/chain_specs/dolphin.rs
stechu: `AssetManagerConfig` is not used at all. Adding back `GenesisConfig`.

pallets/manta-pay/src/lib.rs
stechu: `receivers` gets filled in `pull_receivers_for_shard`. This is a very common trick even in `C/C++`: pass a mutable reference used as return while returning a boolean type as a flag.

pallets/manta-pay/src/lib.rs
stechu: That is a valid point. Less about spamming or DDoS (since failed transaction will be on chain and be charged gas fee any way), more about the convention and things like subsquid support.

pallets/manta-pay/src/lib.rs
Garandor: yeah, in C++ out-references are common to the point of becoming an anti-pattern people are steered away from. e.g. https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f21-to-return-multiple-out-values-prefer-returning-a-struct-or-tuple

the idea is that a return value is simply *the right way* to express something that's clearly the output of a function as opposed to just a side-effect, where an out-reference hides that. This must also be possible for rust. Let me think about how I would rewrite this.

Garandor: ~~Also, isnt it just *wrong* to use `|=` here? If any of the calls returns `true`, `more_receivers` will always be `true` at the end of the call, even if a subsequent loop iteration returns `false`~~
nvm, i get it, it's sharded logic

stechu: > Also, isnt it just _wrong_ to use `|=` here? If any of the calls returns `true`, `more_receivers` will always be `true` at the end of the call, even if a subsequent loop iteration returns `false`

That is the intended behavior. To indicate the client/wallet that it needs to keep pulling since there are more diff to pull.

stechu: I didn't write this code, but the code is quite intuitive to me.

pallets/manta-pay/src/lib.rs
stechu: Reading the code more carefully. The error path will jump out so no event will be emitted. 
Notice the two `?`

primitives/src/assets.rs
stechu: The code has the same semantics :p
Nothing has been changed.

stechu: See previous conversation between @Garandor  and @Dengjianping

pallets/manta-pay/src/lib.rs
Garandor: the main gripe i have is that the fn returns the unimportant bool flag instead of its actual `receivers` vector.
i tried out a couple of tuple-based returns, and am unhappy with them too, so I'm ok keeping the current implementation

runtime/dolphin/src/lib.rs
```diff
@@ -869,6 +869,12 @@ impl_runtime_apis! {
 		}
 	}
 
+	impl pallet_manta_pay::runtime::PullRuntimeApi<Block> for Runtime {
+		fn pull(checkpoint: pallet_manta_pay::RawCheckpoint) -> pallet_manta_pay::PullResponse {
```
Garandor: at RT api level too

pallets/manta-pay/src/lib.rs
Garandor: Why resolve this without addressing it? Want to fix it later?f
Since a change to counted storage map would need migration again, i guess simply removing the FIXME would be the way to go

stechu: I think I meant to remove the FIXME comments line but forget to do so.

runtime/dolphin/src/lib.rs
stechu: I think `pull` here is proper. Since the RPC entrance is `mantaPay_pull`. The semantics is quite clear.

pallets/manta-pay/src/lib.rs
stechu: > the main gripe i have is that the fn returns the unimportant bool flag instead of its actual `receivers` vector. i tried out a couple of tuple-based returns, and am unhappy with them too, so I'm ok keeping the current implementation

Why the bool flag is unimportant? It is quite important in my view. 
More specifically, wallet/client relies on this flag to get a consistent ledger sync behavior.
We actually need to extensively test this behavior since it could easily go wrong and cause inconsistency.

runtime/dolphin/src/lib.rs
Garandor: if you look at the other APIs consensus seems to be to name the API like its callee e.g. `collect_collation_info`
But as you want, I'm ok without it

stechu: Open for better naming suggestion for sure! This is a change will cause front-end/wallet/signer change as well. I feel like we need to do a proper API doc.

pallets/asset-manager/src/lib.rs
```diff
@@ -564,7 +564,7 @@ pub mod pallet {
 					Self::get_para_id_from_multilocation(location.into().as_ref())
 				{
 					if para_id != 2084 {
-						Self::increase_count_of_associated_assets(para_id);
+						let _ = Self::increase_count_of_associated_assets(para_id);
```
Garandor, line 567: the call can fail in line 512. That error will be discarded. Why is that safe?
This behavior was the same before, but now it's explicitly tolerated and I don't see why.

Shouldn't we abort migration if the storage item can not be migrated cleanly?

Garandor (commented):
one more question, then LGTM

I logged a followup to re-align the chainspec files across all runtimes
https://github.com/Manta-Network/Manta/issues/581

runtime/dolphin/src/lib.rs
Garandor: i would just follow precedent and name the API `pull_ledger_diff`.
If that is too big a change i'm fine with reverting back to the old `pull` naming for now

pallets/asset-manager/src/lib.rs
stechu: I don't think there is going back for migration. (cc @Dengjianping )

Garandor:
One more question: The unit tests @bhgomes describes are not really implemented here. Is there a tracking issue to follow up?

runtime/dolphin/src/lib.rs
stechu: I agree `pull_ledger_diff` is a better name, let's change it across the board.

pallets/manta-pay/src/lib.rs
Garandor: the primary purpose of the fn is to return the `receivers`, right? the flag is of seconary importance to trigger a second round of reading. With the chosen signature, the bool flag is the return value, signifying higher importance than some out param.
For that reason i tried a (Vec<Chunks>,bool) return value, but it doesn't compose as nicely as the current impl

pallets/asset-manager/src/lib.rs
Garandor: correct, but we can test with `try_runtime` before deploying to the chain.

This test would currently pass

stechu:
> One more question: The unit tests @bhgomes describes are not really implemented here. Is there a tracking issue to follow up?

#579

Garandor (commented):
thanks for the new naming, it all LGTM except for the migration issue.

@Dengjianping actually alluded to that before that we should add the `try_runtime` feature to `manta-pay` so we can verify the runtime upgrade will succeed before actually issuing it to the chain.

I'll give it a shot to implement it so we can merge this PR today

Garandor (approved):
After discussion, migration of asset_manager will be discussed in #572, so this is good to go

ghzlatarev (approved):

PR #536, opened 2022-05-06
Multiply tx len and base fee component by max number of remarks per block, in the fees calculation of `congested_chain_simulation` test

### LINKED ISSUE #530: Adjust `congested_chain_simulation` test when we integrate account for the `length` component of the transaction fee
Currently the test disregards the `len` component as it was considered negligible for small txs.
With the change from https://github.com/paritytech/substrate/pull/10785 it might be worthwhile to start consider this component in the calculation of the test.

ghzlatarev:

## Description

closes: #530 

* Ad the tx len component to the fees calculation of `congested_chain_simulation`  because when https://github.com/paritytech/substrate/pull/10785 lands in 0.9.18 it might not be so negligible to ignore. It adds $500 to the result cost.
* Calculate a rough amount for maximum `remarks` in a TX and multiply the `len` and `base_weight` components by that amount. 
* The base component being multiplied by the number of extrinsics makes about 10x difference to the result cost, so we're still over our target cost of $250k. Currently costs ~$550k to congest the network on day 1

---

write a little note why.

- [x] Wrote unit tests.

### DIFF
--- a/runtime/calamari/src/fee.rs
+++ b/runtime/calamari/src/fee.rs
@@ -56,8 +56,12 @@ impl WeightToFeePolynomial for WeightToFee {
 
 #[cfg(test)]
 mod multiplier_tests {
-	use crate::{Runtime, RuntimeBlockWeights as BlockWeights, System, TransactionPayment, KMA};
+	use crate::{
+		Call, Runtime, RuntimeBlockWeights as BlockWeights, System, TransactionPayment, KMA,
+	};
+	use codec::Encode;
 	use frame_support::weights::{DispatchClass, Weight, WeightToFeePolynomial};
+	use frame_system::WeightInfo;
 	use pallet_transaction_payment::{Multiplier, TargetedFeeAdjustment};
 	use polkadot_runtime_common::{AdjustmentVariable, MinimumMultiplier, TargetBlockFullness};
 	use sp_runtime::{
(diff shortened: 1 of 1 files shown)


runtime/calamari/src/fee.rs
```diff
@@ -128,9 +132,21 @@ mod multiplier_tests {
 			.max_total
 			.unwrap() - 10;
 
-		let base_fee = <Runtime as pallet_transaction_payment::Config>::WeightToFee::calc(
-			&frame_support::weights::constants::ExtrinsicBaseWeight::get(),
-		);
+		let remark = Call::System(frame_system::Call::<Runtime>::remark_with_event {
```
Garandor, line 136: just for my understanding, you used the remark extrinsic arbitrarily to get a small, constant size TX, right? or is there more to this choice?
maybe add the answer as a comment line to the code for future reference

Garandor (approved):
LGTM!

runtime/calamari/src/fee.rs
ghzlatarev: Yeah that's it, let me comment it.

runtime/calamari/src/fee.rs
```diff
@@ -128,9 +132,22 @@ mod multiplier_tests {
...
+			remark: vec![1, 2, 3],
+		});
+		let len: u32 = remark.clone().encode().len() as u32;
+		let remark_weight: Weight =
+			<Runtime as frame_system::Config>::SystemWeightInfo::remark(len);
+		let max_number_of_remarks_per_block = (block_weight / remark_weight) as u128;
+		let per_byte = <Runtime as pallet_transaction_payment::Config>::TransactionByteFee::get();
+		// length fee. this is not adjusted.
+		let len_fee = max_number_of_remarks_per_block * per_byte.saturating_mul(len as u128);
```
stechu: why the first op use `*` and the second use `saturating_mul`?

stechu: why the first op use `*` and the second use `saturating_mul`?

Dengjianping:
Does it affect the congestion? Worse or better?

ghzlatarev:
> Does it affect the congestion? Worse or better?

The length component doesn't add much. Around $500 at the end of the day.
But the base fee being calculated for each extrinsic makes a big difference. Around 10x to the daily cost, which is ok because we're still over the target $250k. 
Right now the cost to congest the network is ~$550k

runtime/calamari/src/fee.rs
ghzlatarev: Just inconsistency, corrected it.

Dengjianping (approved):
LGTM!

PR #558, opened 2022-05-20
Fix try runtime and metadata diff ci workflows

### LINKED ISSUE #243: GHA for detecting whether transaction version bump is required
-The transaction version must be updated in the following circumstances: extrinsic parameters (number, order, or types) have been changed; extrinsics or pallets have been removed; or the pallet order in the construct_runtime! macro or extrinsic order in a pallet has been changed. If this number is updated, then the spec_version must also be updated.
-This is very hard to determine manually as we're depending on so many pallets from substrate.
-So we should create a GHA to automate the process. The best way to do this is to compare the metadata before and after the changes. Parity has already implemented this so use for reference --> https://github.com/paritytech/polkadot/pull/3631

### LINKED ISSUE #450: CI `try-runtime` test
We need a `try-runtime` test that uses Calamari mainnet to scrape its data.
Most likely will have to be a manually triggered job because it will take a while to scrape that much data.

ghzlatarev:
closes: #243 
closes: #450 

* Try-runtime workflow was broken due to missing npm and yarn on self hosted github runner. Normal non self hosted runners come with installed npm and yarn so switched to that.
* Metadata-diff workflow was broken due to missing relay `--chain` argument, so added `kusama` argument to fix it
* The `pull_request_template.md` was not working so reverted back to `PULL_REQUEST_TEMPLATE.md` directly in the `.github` folder to fix it.

---

write a little note why.

- [x] Wrote unit tests.


Dengjianping (dismissed):
LGTM!

.github/workflows/try-runtime-calamari-mainnet.yml
```diff
@@ -5,67 +5,28 @@
...
+        required: true
 
 env:
-  AWS_INSTANCE_SSH_PUBLIC_KEY: ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIPb24HEQ++aNFWaqVyMtIs6GotUB8R+q61XOoI2z6uMj
-  AWS_REGION: us-east-1
-  AWS_SUBNET_ID: subnet-08c26caf0a52b7c19
-  AWS_SECURITY_GROUP_ID: sg-0315bffea9042ac9b
-  AWS_INSTANCE_TYPE: r5dn.2xlarge
-  AWS_INSTANCE_ROOT_VOLUME_SIZE: 32
-  AWS_IMAGE_SEARCH_PATTERN: ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-*
```
Garandor, line 21: Seems a bit radical to revert self-hosted runners because the image doesn't have yarn.
Have you tried using another base image?
e.g. https://aws.amazon.com/marketplace/pp/prodview-3ypqfpp2gvv6a?sr=0-1&ref_=beagle&applicationId=AWSMPContessa

Alternatively, running https://github.com/nvm-sh/nvm#intro
```bash 
curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash
nvm install node
nvm use node
```
and maybe an `npm -i yarn` if the above doesn't bring it.

Instead of installing on every run, maybe consider making it a custom image
https://docs.aws.amazon.com/toolkit-for-visual-studio/latest/user-guide/tkv-create-ami-from-instance.html

ghzlatarev: This is the same image as the `publish-draft-release,yml` workflow, where it always worked, and I couldn't figure out what else is wrong.
Anyway now I tried it again and seems to work ... https://github.com/Manta-Network/Manta/actions/runs/2390139342

Garandor (approved):
LGTM

Dengjianping (approved):
LGTM!

PR #563, opened 2022-05-24
Reimplement `TransactAsset` with `FungibleLedger` and `AssetConfig` traits

### LINKED ISSUE #534: Simplify `AssetTransactor` logic with the `ConcreteFungibleLedger` trait
Right now the `AssetTransactor` uses a tuple with 2 implementations of the transact functionality, called one after the other during exection.
This can be simplified with the unified interface of fungible and fungibles in `ConcreteFungiblesLedger` trait.

ghzlatarev:
## Description

closes: #534

* Replace `FungiblesTransactor` and `LocalAssetTransactor` tuple with our own `AssetTransactor` implementor called `MultiAssetAdapter`
* `MultiAssetAdapter` uses the unified interface of `FungibleLedger` trait to implement only the `withdraw_asset` and `deposit_asset` functions from `TransactAsset`
* The rest of the `TransactAsset` interface like can_check_in, teleport, etc are not required. Even though they were implemented in our old configuration they were not used.
* An additional `burn` function and implementation had to be added to our `FungiblesLedger` trait. This implementation allows accounts to go down to 0, as that was the behavior of the old configuration.
* `ReducedToZero` return results in `can_withdraw` implementation were considered an error, when in fact they are a success case, so switched that.
* XCM tests were refactored to account for the new configuration
* Refactored 1 manta-pay test which was supposed to fail because of attempting to mint more private assets than the available public assets of an account, but was actually minting the exact available amount. However since `ReducedToZero` was considered an error, this case was also holding up.
---

write a little note why.

- [x] Wrote unit tests.


primitives/src/assets.rs
```diff
@@ -81,6 +81,15 @@ pub trait AssetConfig<C>: 'static + Eq + Clone
...
 {
+	// type AssetId: Member
+	// 	+ Parameter
+	// 	+ Default
+	// 	+ Copy
+	// 	//+ HasCompact
+	// 	//+ MaybeSerializeDeserialize
+	// 	//+ MaxEncodedLen
+	// 	+ TypeInfo;
+
```
ghzlatarev, line 92: clean up

primitives/src/assets.rs
```diff
@@ -471,4 +489,27 @@ where
...
+		amount: Balance,
+	) -> Result<(), FungibleLedgerError> {
+		Self::ensure_valid(asset_id)?;
+		Self::can_withdraw(asset_id, who, amount)?;
+		if asset_id == A::NativeAssetId::get() {
+			<Native as Currency<C::AccountId>>::withdraw(
+				who,
+				amount,
+				WithdrawReasons::TRANSFER,
+				ExistenceRequirement::AllowDeath,
```
ghzlatarev: `AllowDeath` because the `can_withdraw` impl uses `keep_alive = false` argument

primitives/src/assets.rs
```diff
@@ -323,10 +335,9 @@ impl FungibleLedgerError {
 			WithdrawConsequence::NoFunds => Self::NoFunds,
 			WithdrawConsequence::Overflow => Self::Overflow,
 			WithdrawConsequence::Underflow => Self::Underflow,
-			WithdrawConsequence::ReducedToZero(balance) => Self::ReducedToZero(balance),
 			WithdrawConsequence::UnknownAsset => Self::UnknownAsset,
 			WithdrawConsequence::WouldDie => Self::WouldDie,
-			WithdrawConsequence::Success => return Ok(()),
+			WithdrawConsequence::Success | WithdrawConsequence::ReducedToZero(_) => return Ok(()),
```
ghzlatarev: `ReducedToZero` is actually a success case, because it happens when the account doesn't have to be kept alive.

primitives/src/xcm.rs
```diff
@@ -312,3 +320,101 @@ where
...
+	}
+
+	fn withdraw_asset(
+		asset: &MultiAsset,
+		location: &MultiLocation,
+	) -> result::Result<xcm_executor::Assets, XcmError> {
+		let (asset_id, amount, who) = Self::match_asset_and_location(asset, location)?;
+
+		MultiAdapterFungibleLedger::burn(asset_id, &who, amount)
+			.map_err(|_| xcm::v2::Error::FailedToTransactAsset("Failed Burn"))?;
```
ghzlatarev, line 416: import xcm::v2::Error

runtime/calamari/src/assets_config.rs
```diff
@@ -123,6 +123,7 @@ pub type CalamariConcreteFungibleLedger =
 pub struct CalamariAssetConfig;
 
 impl AssetConfig<Runtime> for CalamariAssetConfig {
+	//type AssetId = AssetId;
```
ghzlatarev, line 126: remove commented code.

runtime/calamari/src/xcm_config.rs
```diff
@@ -144,21 +133,18 @@ parameter_types! {
 }
 
 /// Transactor for currency in pallet-assets, i.e. implements `fungibles` trait
-pub type FungiblesTransactor = FungiblesAdapter<
-	Assets,
+pub type MultiAssetTransactor = MultiAssetAdapter<
```
ghzlatarev, line 136: Documentation. Copy over to all other files.

runtime/common/tests/xcm_mock/mod.rs
```diff
@@ -96,6 +102,13 @@ pub fn para_ext(para_id: u32) -> sp_io::TestExternalities {
 	)
 	.unwrap();
 
+	pallet_asset_manager::GenesisConfig::<Runtime> {
+		start_id: <ParachainAssetConfig as AssetConfig<Runtime>>::StartNonNativeAssetId::get(),
+		_marker: PhantomData::<Runtime>::default(),
+	}
+	.assimilate_storage(&mut t)
+	.unwrap();
```
ghzlatarev, line 110: This is kind-of hidden detail of the tests but it's helpful because it sets the native asset and moves the next asset id along, so the `AssetConfig` impl can be used directly in the tests.

runtime/common/tests/xcm_mock/parachain.rs
```diff
@@ -188,32 +188,19 @@ parameter_types! {
...
+pub type MultiAssetTransactor = MultiAssetAdapter<
+	Runtime,
 	IsNativeConcrete<SelfReserve>,
 	LocationToAccountId,
-	AccountId,
-	(),
->;
-
-/// Transactor for currency in pallet-assets, i.e. implements `fungibles` trait
-pub type FungiblesTransactor = FungiblesAdapter<
```
ghzlatarev, line 202: Copy over documentation.

primitives/src/assets.rs
```diff
@@ -471,4 +489,27 @@ where
 		}
 		.map_err(|e| FungibleLedgerError::InvalidTransfer(e))
 	}
+
```
ghzlatarev: The `transfer()` impl actually tries to keep accounts alive, instead of allowing death. 
This is inconsistent with the behavior of `TransactAsset::withdraw` that we had so far and that I kept in this PR with `AllowDeath` in the `burn` impl

stechu:
My general feeling is that `Keep_Alive` and `Allow_Death` is a bit adhoc now. We probably need to write down the correct semantics for when and when not to `Keep_Alive` in the context of both XCM and MantaPay.

ghzlatarev:
> My general feeling is that `Keep_Alive` and `Allow_Death` is a bit adhoc now. We probably need to write down the correct semantics for when and when not to `Keep_Alive` in the context of both XCM and MantaPay.

In the context of XCM we should use `Allow_Death` for withdrawals/transfers imo. First, it's not any different than regular balance transfers, where this is allowed. Obviously if something relies on the account being alive, it will not be allowed to die anyway. I don't see any benefit in supporting both options besides UX> Second, the `can_decrease` impl that we rely on allows death, which means also the old transactors logic use `allow-death` by default. We can't make it a user input from xcm perspective because this is not an input to xcm instructions, or xtokens calls, etc.

In the context of MantaPay  we could make it a user input though. We'll have to slightly refactor the `FungibleLedger` trait and some of the manta-pay extrinsics, but I'm not sure if it's really needed.

@bhgomes

stechu:
Thanks for the PR.

primitives/src/xcm.rs
```diff
@@ -312,3 +320,101 @@ where
...
+	MultiAssetAdapter<
+		Runtime,
+		NativeMatcher,
+		AccountIdConverter,
+		AssetsMatcher,
+		MultiAdapterFungibleLedger,
+		MultiAdapterAssetConfig,
+	>
+{
+	fn match_asset_and_location(
```
stechu: Rust doc about the semantics of this function.

stechu:
A general comment on the design:
we need a clear documentation on the `allowDeath` behavior of all methods defined inside `FungibleLedger`, like 
`transfer`, `burn` etc. Also since `FungibleLedger` is used in both `mantaPay` and cross-chain token transfer. We need to examine in each use case, what is the `allowDeath` situation that makes sense.

Garandor:
merge conflicts to resolve

Garandor (commented):
LGTM, as it's an XCM PR I'd like to ask @Dengjianping and @stechu to review for approval here as well

pallets/asset-manager/src/lib.rs
```diff
@@ -522,6 +522,11 @@ pub mod pallet {
                 Ok(())
             }
         }
+
+        #[cfg(feature = "std")]
```
Dengjianping: Seems, it's just for testing,  but will be compiled into native binary.
Is it possible to add one more attribute here `#[cfg(test)]`?
But I'm not sure it would compile.

primitives/src/xcm.rs
```diff
@@ -312,3 +320,117 @@ where
...
+        Runtime,
+        AccountIdConverter,
+        NativeMatcher,
+        NonNativeMatcher,
+        MultiAdapterFungibleLedger,
+        MultiAdapterAssetConfig,
+    >
+{
+    fn deposit_asset(asset: &MultiAsset, location: &MultiLocation) -> XcmResult {
+        log::trace!(
```
Dengjianping: `trace` might print too much redundant logs, `debug` would be more friendly, imo.

primitives/src/xcm.rs
```diff
@@ -312,3 +320,117 @@ where
...
+        })?;
+
+        let (asset_id, amount) = match (
+            NativeMatcher::matches_fungible(asset),
+            NonNativeMatcher::matches_fungibles(asset),
+        ) {
+            // native asset
+            (Some(amount), _) => (MultiAdapterAssetConfig::NativeAssetId::get(), amount),
+            // assets asset
+            (_, result::Result::Ok((asset_id, amount))) => (asset_id, amount),
```
Dengjianping: I suppose `result::Result::` can be removed?

primitives/src/assets.rs
```diff
@@ -476,4 +481,27 @@ where
...
+    #[inline]
+    fn burn(
+        asset_id: AssetId,
+        who: &C::AccountId,
+        amount: Balance,
+    ) -> Result<(), FungibleLedgerError> {
+        Self::ensure_valid(asset_id)?;
+        Self::can_withdraw(asset_id, who, amount)?;
+        if asset_id == A::NativeAssetId::get() {
+            <Native as Currency<C::AccountId>>::withdraw(
```
Dengjianping, line 494: For native assets, it's transferred instead of burned, kind of confused to me. Perhaps leaving some comments here would make more sense.

runtime/calamari/Cargo.toml
```diff
@@ -80,6 +80,7 @@ xcm-executor = { git = 'https://github.com/paritytech/polkadot.git', default-fea
 
 # Third party (vendored) dependencies
 orml-traits = { git = "https://github.com/manta-network/open-runtime-module-library.git", default-features = false, rev = "7e2f985" }
+orml-xcm-support = { git = "https://github.com/manta-network/open-runtime-module-library.git", default-features = false, rev = "7e2f985" }
```
Dengjianping: Where is this crate used? I mean I didn't find any type or trait is import in calamari-runtime.

primitives/src/xcm.rs
```diff
@@ -312,3 +320,117 @@ where
...
+{
+    /// Matches the incoming `asset` to an `asset_id` and `amount` on this chain.
+    /// Matches the incoming `location` to a `receiver` account on this chain.
+    /// Uses the matcher implementation of both native and non-native assets.
+    /// Returns the `asset_id`, `amount` and `receiver` if all three were matched.
+    fn match_asset_and_location(
+        asset: &MultiAsset,
+        location: &MultiLocation,
+    ) -> result::Result<(AssetId, Balance, Runtime::AccountId), XcmError> {
+        let receiver = AccountIdConverter::convert_ref(location).map_err(|_| {
```
Dengjianping, line 418: Does it work for account type like `AccountKey20`?

Dengjianping (changes requested):

primitives/src/assets.rs
ghzlatarev: It is not transferred, the funciton is `withdraw`. The `Transfer` parameter are `reasons` which are used for `frozen()` checks

runtime/calamari/Cargo.toml
ghzlatarev: This is not used actually, will remove.

primitives/src/xcm.rs
ghzlatarev: Depends on the runtime configuration. In our particular case we don't handle `AccountKey20` locations. But remember that the expected sender origin is actually the parachain sovereign account, not a user account. That's why we don't need to handle this conversion.
User account origin logic is handled on the sender side. Once the message is passed to the backend transportation system, the origin is changed to the parachain id of the sender.

pallets/asset-manager/src/lib.rs
ghzlatarev: It doesn't compile, can't find the function when running tests.

ghzlatarev: This is for convenience while testing, I don't think there's any risk to it.
But the other option is to set the proper native and non native start at genesis.
What do you guys think @Dengjianping @stechu @Garandor

Dengjianping:
Please resolve clippy warnings.

Dengjianping (dismissed):
lgtm!
Great PR!

CHANGELOG.md
```diff
@@ -21,18 +22,8 @@
...
-- [\#507](https://github.com/Manta-Network/Manta/pull/507) Add issue template for Calamari xcm onboarding of other parachains.
-- [\#519](https://github.com/Manta-Network/Manta/pull/519) Concrete fungible ledger integration tests.
-- [\#523](https://github.com/Manta-Network/Manta/pull/523) Move xcm and assets related runtime configurations to own files.
-- [\#531](https://github.com/Manta-Network/Manta/pull/531) Clean up AssetManager migration code.
-- [\#541](https://github.com/Manta-Network/Manta/pull/541) Skip build on too tiny change.
-- [\#542](https://github.com/Manta-Network/Manta/pull/542) Update xcm integrations template issue.
-- [Calamari] [\#550](https://github.com/Manta-Network/Manta/pull/550) Remove sudo pallet from calamari runtime.
-- [\#560](https://github.com/Manta-Network/Manta/pull/560) Bump srtool to v0.4.0.
-- [Dolphin] [\#583](https://github.com/Manta-Network/Manta/pull/583) Remove checkpoint from RPC API when synchronizing with MantaPay.
+- [\#563](https://github.com/Manta-Network/Manta/pull/563) Re-implement the `TransactAsset` trait with the unified interface of `FungibleLedger` trait, and AssetConfig trait.
```
Garandor: I dont think you should be deleting v3.2.0 changelog entries here

Garandor (changes requested):
Forgot i never sent this review 🥇

CHANGELOG.md
ghzlatarev: oh my bad this must've been a bad merge conflict resolution

Dengjianping (approved):
lgtm!

Garandor (approved):
LGTM if you fix the changelog in #614 after merging this

Issue #452, opened 2022-03-21
Investigate (0, Here) for `SelfReserve`


ghzlatarev:
With this PR we can use (0, Here) for `SelfReserve` https://github.com/open-web3-stack/open-runtime-module-library/pull/708
But this might have other reprecuissions that need to be investigated.
The reason we chose (1, Parachain(id)) representations is because it will be consistent with how it is viewed on sibling chains. 
Otherwise we'd have to treat it specially, which is what needs to be explored in this issue.

ghzlatarev:
Needs orml release with https://github.com/open-web3-stack/open-runtime-module-library/pull/787 which is polkadot-v0.9.15

ghzlatarev:
https://github.com/Manta-Network/Manta/pull/761

PR #576, opened 2022-06-06
Unfilter `xtokens.transfer_multicurrencies` and bump `MaxAssetsForTransfer` to 2

### LINKED ISSUE #516: Integrate Statemine assets
- [x] Test locally
- [x] Test on Rococo
- [x] Send message to request to open Calamari -> Statemine channel
- [x] Governance motion on Kusama to command Statemine to accept our channel and open Statemine -> Calamari
- [ ] Send message to accept their channel
- [ ] Register RMRK and USDT assets
- [x] Enable `xtokens.transferMulticurrencies`

USDT: 
`parents: 1 interior: { X3: [ { Parachain: 1,000 } { PalletInstance: 50 } { GeneralIndex: 1,984 } ] }`
```
  deposit: 0
  name: Tether USD
  symbol: xcUSDT
  decimals: 6
  isFrozen: false
  minBalance: 1
  isSufficient: true
```
RMRK:
`parents: 1 interior: { X3: [ { Parachain: 1,000 } { PalletInstance: 50 } { GeneralIndex: 8 } ] }`
```
  deposit: 0
  name: xcRMRK
  symbol: xcRMRK
  decimals: 10
  isFrozen: false
  minBalance: 1
  isSufficient: true
```

Example governance motion ----> https://kusama.polkassembly.io/motion/417

ghzlatarev:

## Description

closes: #516 

* Statemine currently only allows to pay fees in KSM, so if we want to send assets to Statemine we need to also send KSM.
* To do so we have to unfilter `xtokens.transfer_multicurrencies` 
* Also have to bump `MaxAssetsForTransfer` to 2
* Added some must-fail and must pass tests to simulate different scenarios.

---

write a little note why.

- [x] Wrote unit tests.


runtime/common/Cargo.toml
```diff
@@ -13,6 +13,7 @@ repository = 'https://github.com/Manta-Network/Manta/'
 # 3rd dependencies
 codec = { package = "parity-scale-codec", version = "3.0" }
 scale-info = { version = "2.0", features = [ "derive" ] }
+smallvec = "1.8.0"
```
ghzlatarev, line 16: TODO: remove these.

runtime/common/tests/xcm_mock/parachain.rs
```diff
@@ -577,7 +577,7 @@ where
 
 parameter_types! {
 	pub const BaseXcmWeight: Weight = 100_000_000;
-	pub const MaxAssetsForTransfer: usize = 2;
+	pub const MaxAssetsForTransfer: usize = 3;
```
ghzlatarev: Need 3 for some of the must-fail scenarios.

runtime/calamari/src/lib.rs
```diff
@@ -239,9 +239,8 @@ impl Contains<Call> for BaseFilter {
 			// orml_xtokens::Call::transfer_with_fee {..}
 			// orml_xtokens::Call::transfer_multiasset {..}
 			// orml_xtokens::Call::transfer_multiasset_with_fee {..}
-			// orml_xtokens::Call::transfer_multicurrencies  {..}
 			// orml_xtokens::Call::transfer_multiassets {..}
-			| Call::XTokens(orml_xtokens::Call::transfer {..})
+			| Call::XTokens(orml_xtokens::Call::transfer {..} | orml_xtokens::Call::transfer_multicurrencies  {..})
```
Garandor: this'll need redoing after we merge #590 as the lines moved

Garandor:
merge conflicts to resolve

runtime/calamari/src/lib.rs
```diff
@@ -253,7 +252,7 @@ impl Contains<Call> for BaseFilter {
                 | manta_collator_selection::Call::leave_intent{..})
             | Call::Balances(_)
             | Call::Preimage(_)
-            | Call::XTokens(orml_xtokens::Call::transfer {..})
+            | Call::XTokens(orml_xtokens::Call::transfer {..} | orml_xtokens::Call::transfer_multicurrencies {..})
```
Garandor: minor: i'm surprised this passes `cargo fmt`, could you please add the second one on a new line like CollatorSelection above

runtime/dolphin/src/lib.rs
```diff
@@ -205,11 +205,9 @@ impl Contains<Call> for BaseFilter {
                                 orml_xtokens::Call::transfer_with_fee {..}
                                 | orml_xtokens::Call::transfer_multiasset {..}
                                 | orml_xtokens::Call::transfer_multiasset_with_fee {..}
-                                | orml_xtokens::Call::transfer_multicurrencies {..}
                                 | orml_xtokens::Call::transfer_multiassets {..})
             // Everything except transfer() is filtered out until it is practically needed:
             | Call::XcmpQueue(_) | Call::PolkadotXcm(_) | Call::DmpQueue(_) => false,
-
```
Garandor: if you prefer the condensed look without a newline, pls also adapt manta and calamari so we dont diverge

runtime/dolphin/src/lib.rs
```diff
@@ -250,11 +248,10 @@ impl Contains<Call> for BaseFilter {
                 | manta_collator_selection::Call::remove_collator{..}
                 | manta_collator_selection::Call::leave_intent{..})
             | Call::Balances(_)
-            | Call::XTokens(orml_xtokens::Call::transfer {..})
+            | Call::XTokens(orml_xtokens::Call::transfer {..} | orml_xtokens::Call::transfer_multicurrencies  {..})
```
Garandor: same

runtime/common/tests/xcm_tests.rs
```diff
@@ -2156,3 +2157,579 @@ fn less_than_min_xcm_fee_should_not_work() {
...
+    let b_asset_id_on_a = register_assets_on_parachain::<ParaA>(
+        &para_b_source_location,
+        &para_b_asset_metadata,
+        Some(units_per_sec),
+        None,
+    );
+
+    // register ParaB native asset on ParaB
+    let b_asset_id_on_b = register_assets_on_parachain::<ParaB>(
+        &para_b_source_location,
```
Garandor, line 2209: how come the locations are identical between paraA and B? I thought XCM locations are relative to the source?

runtime/common/tests/xcm_tests.rs
```diff
@@ -2156,3 +2157,579 @@ fn less_than_min_xcm_fee_should_not_work() {
...
+        ),
+    };
+
+    // Send some ParaB tokens from Alice on B to Alice on A
+    let amount_to_a = 10000000;
+    let weight = 40;
+    ParaB::execute_with(|| {
+        assert_ok!(parachain::XTokens::transfer(
+            Some(ALICE).into(),
+            parachain::CurrencyId::MantaCurrency(b_asset_id_on_b),
```
Garandor, line 2257: b was registered but never minted a balance. is the balance transferred here from the genesis endowment to alice on mock net?

runtime/common/tests/xcm_tests.rs
```diff
@@ -2156,3 +2157,579 @@ fn less_than_min_xcm_fee_should_not_work() {
...
+
+        assert_eq!(
+            parachain::Assets::balance(relay_asset_id_on_a, &ALICE),
+            relay_asset_amount_minted_on_a - fee_amount
+        );
+        assert_eq!(
+            // notice that total supply between the two chains is:
+            // (relay_asset_amount_minted_on_a - fee_amount) + (fee_amount - min_xcm_fee)
+            parachain::Assets::total_supply(relay_asset_id_on_a),
+            relay_asset_amount_minted_on_a - fee_amount
```
Garandor, line 2316: according to your comment above this should be `relay_asset_amount_minted_on_a - min_xcm_fee`

Which one is correct?

runtime/common/tests/xcm_tests.rs
```diff
@@ -2156,3 +2157,579 @@ fn less_than_min_xcm_fee_should_not_work() {
...
+    let _ = register_assets_on_parachain::<ParaC>(
+        &relay_source_location,
+        &relay_asset_metadata,
+        Some(units_per_sec),
+        None,
+    );
+
+    let amount = 1000;
+    let weight = 40;
+    assert_ok!(ParaA::execute_with(|| {
```
Garandor, line 2459: I'm not sure what you're testing here. Could you add a `///` doccomment fn header describing the test scenarios with a few words?

Garandor (commented):
minor format notes and questions.

Also, thanks for using clear variable names

Gonna leave approval to @Dengjianping and @stechu, not comfortable approving xcm code yet.

runtime/common/tests/xcm_tests.rs
ghzlatarev: Yes you're right, the correct way of representing a self location (native asset location) should be (0, Here). 
However representing in a "absolute" way (1, X1(self_para_id)) was a common mistake that many teams copied, including us. 
We worked around this mistake `impl<T, Balance> MatchesFungible<Balance> for IsNativeConcrete<T>` where we also check for `MultiLocation::is_here` 
Actually my next PR will be to actually start using the correct way of handling self locations and the code is ready, just need the other PRs to go in first. https://github.com/Manta-Network/Manta/issues/452

runtime/common/tests/xcm_tests.rs
ghzlatarev: Yeah this is in the mock setup:
```
pallet_balances::GenesisConfig::<Runtime> {
		balances: vec![(ALICE, INITIAL_BALANCE)],
	}
	.assimilate_storage(&mut t)
	.unwrap();
```
This is kind of hidden logic , which I'm usually against , so I didn't for example add `AssetManager` storage setup there. But this is not too unreasonable to assume, that's why I didn't rewrite it.
Not a big deal if you guys want to make it explicit, just let me know.

runtime/common/tests/xcm_tests.rs
ghzlatarev: The comment is not great. The total supply should be counted from both chains:
`relay_asset_amount_minted_on_a - fee_amount` is still on `ParaA`
`fee_amount - min_xcm_fee` is on ParaB.
And the thing to notice is that it comes out to `relay_asset_amount_minted_on_a - min_xcm_fee` meaning 1 `min_xcm_fee` is destroyed. This is a design choice by ORML people to make these kind of transfer. The end situation is that some of the relay chain token held as reserve on the relay chain, will become unwithdrawable.
Will rewrite it.

runtime/calamari/src/lib.rs
ghzlatarev: Oops forgot about this. I noticed that this wasn't getting formatted with ctrl-s, weird...

runtime/common/tests/xcm_tests.rs
ghzlatarev: First part are supposed to be only must fail cases on the sender side. covering all the error messages.
Added some comments on the asserts.
The second part are must fail case on the receiver part.

runtime/dolphin/src/lib.rs
ghzlatarev: No, accidentally removed this line. Reverted it now.

runtime/common/tests/xcm_tests.rs
ghzlatarev: Please don't make me write comments for all the tests though :D

runtime/common/tests/xcm_tests.rs
Garandor: So then the comment is correct and the logic in 2311 isn't, right?

P.S. since you checked in changes for the other points but not this one, am I right to assume you forgot to add this change?
You might want to resolve a comment only after you've pushed the commit that makes it outdated. I use that myself as kind of a checklist

runtime/common/tests/xcm_tests.rs
Garandor: Agree it'd be torture to comment each test case, my idea was to annotate the main fn:
https://github.com/Manta-Network/Manta/pull/576/files/17def4e433d575d75d82e3de7150e0871ce4ae3d#diff-c55c7706ced753270f03a35c10fdb301207231f2224a8e3f3951c5970f546643R2365

With a comment describing the structure you just put in the comment here.
So that when i'm reading the code I know what to expect from this long fn instead of having to rhyme things together line by line.

An alternative would be to split this up into multiple shorter fns (which I'd actually prefer, but might not be reasonable)

runtime/common/tests/xcm_tests.rs
ghzlatarev: Hmm something weird happened here, I wrote a big comment but it wasn't pushed.

ghzlatarev: https://github.com/Manta-Network/Manta/pull/576/commits/a140e5a016635b48e2abcf985d84d2e98461b4fc This is the commit with the change, I don't know why it's not reflected in this diff.
Also I remember resolving this conversation, but somehow it is not resolved...

ghzlatarev: Actually none of my changes from the last 3 commits are reflected

ghzlatarev: Please refresh the diff, should be showing up correctly now , but I think I understood your question about 2311:
```
                               (
					parachain::CurrencyId::MantaCurrency(relay_asset_id_on_a),
					fee_amount
				)
```
This is the fee being sent to ParaB, but the logic in ORML will subtract `min_xcm_fee` so the comment is `fee_amount - min_xcm_fee is on ParaB.` is correct. Let me write this explicitly.

runtime/common/tests/xcm_tests.rs
ghzlatarev: Ok, added the comment. Let me know if it doesn't make sense.

runtime/common/tests/xcm_tests.rs
Garandor: ok, got it thanks. i thought total_supply was calculated over both chains, not from the view of ParaA

Also, I unresolved this comment earlier to make sure you see it before merging

Garandor (dismissed):
LGTM after merging the log conflict

Garandor (dismissed):

Dengjianping (dismissed):
LTGM!
Thanks for the PR.
But I think this extrinsic `transfer_multicurrencies` is kind of more complicated than `transfer`, it would be better if there's a how-to tutorial.

ghzlatarev:
> LTGM! Thanks for the PR. But I think this extrinsic `transfer_multicurrencies` is kind of more complicated than `transfer`, it would be better if there's a how-to tutorial.

I think we decided not to advertise the polkadot-js front-end to users at all, otherwise I would've written docs for the `transfer` as well. We want to push users to our app, that was the reason.

Dengjianping (dismissed):
LGTM!

Garandor (dismissed):

Dengjianping (approved):
LGTM!

Garandor (approved):

PR #578, opened 2022-06-07
Runtime for new Kusama Parachain

### LINKED ISSUE #580: Runtime for new Kusama Parachain
- [x] Reserve new para id
- [x] Prepare runtime with new id and secrets. Can be anything as long as it can produce blocks and has sudo
- [x] Get genesis state and wasm blob of new runtime.
- [x] Launch nodes for new parachain
- [x] Win auction
- [x] Start producing blocks
- [x] Send pallet_xcm.send(Transact(regisrar.swap(new_id, 2084)) from the new parachain to Kusama, with sudo.
- [x] Send pallet_xcm.send(Transact(regisrar.swap(2084, new_id)) from Calamari to Kusama, with governance.

ghzlatarev:
## Description

closes: #580

---

write a little note why.

- [ ] Wrote unit tests.


node/src/command.rs
```diff
@@ -40,7 +40,7 @@ pub type Block = generic::Block<Header, OpaqueExtrinsic>;
 
 pub const MANTA_PARACHAIN_ID: u32 = 2015;
 pub const CALAMARI_PARACHAIN_ID: u32 = 2084;
-pub const DOLPHIN_PARACHAIN_ID: u32 = 2084;
+pub const DOLPHIN_PARACHAIN_ID: u32 = 2085;
```
ghzlatarev: TODO: change to actual reserved id.

node/src/chain_specs/dolphin.rs
```diff
@@ -216,10 +216,145 @@ fn dolphin_dev_genesis(
...
+			// account id: dmtwRyEeNyRW3KApnTxjHahWCjN5b9gDjdvxpizHt6E9zYkXj
+			hex!["0027131c176c0d19a2a5cc475ecc657f936085912b846839319249e700f37e79"].into(),
+			hex!["8ebf03bda1702d719f428bc0a4c7cfca010c44a48ef79752490818c901548d20"]
+				.unchecked_into(),
+		),
+	];
+
+	// root account: dmyBqgFxMPZs1wKz8vFjv7nD4RBu4HeYhZTsGxSDU1wXQV15R
+	let root_key: AccountId =
+		hex!["bc153ffd4c96de7496df009c6f4ecde6f95bf67b60e0c1025a7552d0b6926e04"].into();
```
ghzlatarev: TODO: change to whatever sudo account is going to be used.

Dengjianping:
1. ~~Can we use manta instead of dolphin for auction? I mean dolphin is kind of heavy to do such job.~~
2. 3 collators should be enough.
3. Launch nodes more early to catch blocks. Perhaps there's snapshot which can accelerate the sync..
4. I didn't see new new peer ids. Need to be created as well.
5. Related names like the name of runtime, token symbol, etc.

ghzlatarev:
> 1. ~Can we use manta instead of dolphin for auction? I mean dolphin is kind of heavy to do such job.~
> 2. 3 collators should be enough.
> 3. Launch nodes more early to catch blocks. Perhaps there's snapshot which can accelerate the sync..
> 4. I didn't see new new peer ids. Need to be created as well.
> 5. Related names like the name of runtime, token symbol, etc.

We don't need 5. Just need to use different ID.

Garandor (commented):
LGTdeploy (we're not merging this to manta, right?)

ghzlatarev:
> LGTdeploy (we're not merging this to manta, right?)

nope.

Issue #486, opened 2022-04-13
Tracking: Switch Chain Consensus to NPOS


Garandor:
In order to allow for permissionless staking and robust block production, we switch to a Nominated POS block production scheme for our Parachain as the current Aura-based method has technical limitations.

This is a tracking issue colllecting tasks undertaken with this goal in mind

Stages:

0. Nimbus keys enabled onchain #668 
1. Nimbus-Aura enabled #669
2. Nimbus-Aura with staking
    1. On Dolphin #690
    2. On Calamari #691
4. Nimbus-PseudoRandom or VRF-based BlockProducer
5. TBD NPoS
6. TBD Block-miss-resistant Block Production scheme

PR #582, opened 2022-06-08
Consensus migration stage 1: Enable Nimbus-Aura

### LINKED ISSUE #612: Metrics are not showing collated blocks
Metrics are not showing collated blocks while we are active collators.

When we do `curl localhost:9615/metrics | grep "substrate_proposer_block_constructed"`

the results is:

`substrate_proposer_block_constructed_count{chain="calamari"} 0`

Garandor:
This PR enables Nimbus-Aura consensus for Dolphin as stage 1 of #486

It creates a 3-step author filter pipeline:

1.  Check that a NimbusId attached to a block seal belongs to an AccountId and is registered with 
 SessionKeys
2. Check that the ValidatorId belonging to that account is part of Session::validators()
3. Select one collator from the set of eligibles per slot in round-robin fashion = aura

followup to #668 
addresses #669 
fixes #612

NOTE:
- Failure of `docker-image-test` is expected as this is a consensus-breaking change. The new client is incompatible with the currently deployed state of the calamari chain and will only work when the chain runtime is upgraded.
- Failure of dolphin- and calamari-integration-test is also expected and depends on an upstream fix in `polkadot-js/api`. The submitted PR is: https://github.com/polkadot-js/api/pull/5142

Todo:
- [x] Enable on calamari
- [x] runtime-benchmarks / try-runtime 
- ~~[ ] remove left-over mentions of aura~~
Can't do this as we still use the Aura session key until we migrate away
- [x] Merge `service_nimbus` back into `service`, leave old aura service used for manta as `service_aura`
- Upgrading  `manta` runtime to nimbus left for a followup issue
- [x] Localize/Fix avg block time degradation - this was a misconfiguration of local testing. block avg is 12s as normal
- [x] Fix benchmark failure
- [x] Fix polkadot-js API not getting block author correctly on Dolphin


runtime/dolphin/src/lib.rs
```diff
@@ -838,6 +876,23 @@ impl_runtime_apis! {
 		}
 	}
 
+    impl nimbus_primitives::NimbusApi<Block> for Runtime {
```
Garandor, line 947: I'm a bit at a loss here, I can't seem to implement the runtime APIs correctly.
@Dengjianping or @ghzlatarev could you please have a look what could be wrong with these impls?
Getting a unsatisfied trait bound for the two new impls, but it looks fine to me.
The corresponding `decl_runtime_apis!` macro is in `manta_primitives` which is listed in cargo.toml
```
error[E0277]: the trait bound `dolphin_runtime::RuntimeApi: nimbus_primitives::AuthorFilterAPI<sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>, nimbus_primitives::nimbus_crypto::Public>` is not satisfied
   --> node/src/command.rs:461:7
    |
461 |                         dolphin_runtime::RuntimeApi,
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `nimbus_primitives::AuthorFilterAPI<sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>, nimbus_primitives::nimbus_crypto::Public>` is not implemented for `dolphin_runtime::RuntimeApi`
    |
    = help: the trait `nimbus_primitives::AuthorFilterAPI<__SR_API_BLOCK__, nimbus_primitives::nimbus_crypto::Public>` is implemented for `dolphin_runtime::RuntimeApiImpl<__SR_API_BLOCK__, RuntimeApiImplCall>`
note: required by a bound in `service_nimbus::start_parachain_node`
   --> node/src/service_nimbus.rs:421:11
    |
408 | pub async fn start_parachain_node<RuntimeApi, Executor>(
    |              -------------------- required by a bound in this
...
421 |         + nimbus_primitives::AuthorFilterAPI<Block,NimbusId>
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `service_nimbus::start_parachain_node`

error[E0277]: the trait bound `dolphin_runtime::RuntimeApi: nimbus_primitives::NimbusApi<sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>>` is not satisfied
   --> node/src/command.rs:461:7
    |
461 |                         dolphin_runtime::RuntimeApi,
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `nimbus_primitives::NimbusApi<sp_runtime::generic::Block<sp_runtime::generic::Header<u32, BlakeTwo256>, OpaqueExtrinsic>>` is not implemented for `dolphin_runtime::RuntimeApi`
    |
    = help: the trait `nimbus_primitives::NimbusApi<__SR_API_BLOCK__>` is implemented for `dolphin_runtime::RuntimeApiImpl<__SR_API_BLOCK__, RuntimeApiImplCall>`
note: required by a bound in `service_nimbus::start_parachain_node`
   --> node/src/service_nimbus.rs:422:11
    |
408 | pub async fn start_parachain_node<RuntimeApi, Executor>(
    |              -------------------- required by a bound in this
...
422 |         + nimbus_primitives::NimbusApi<Block>
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `service_nimbus::start_parachain_node`
```

Garandor: **UPDATE: Nevermind, no need to c I seem to be missing a `sp_api::decl_runtime_apis`**

node/src/service_nimbus.rs
```diff
@@ -0,0 +1,657 @@
...
+		>,
+		(Option<Telemetry>, Option<TelemetryWorkerHandle>),
+	>,
+	sc_service::Error,
+>
+where
+	RuntimeApi: ConstructRuntimeApi<Block, TFullClient<Block, RuntimeApi, NativeElseWasmExecutor<Executor>>>
+		+ Send
+		+ Sync
+		+ 'static,
```
Garandor: I think I'm messing up the trait bounds here @Dengjianping

runtime/dolphin/src/lib.rs
ghzlatarev: https://github.com/Manta-Network/Manta/commit/72b3b2de6e0975a332b36feb98b144f6a215f4e2 seems the trait bounds were mixed up. Compiles with this change.

runtime/dolphin/src/lib.rs
```diff
@@ -707,11 +728,14 @@ construct_runtime!(
         TechnicalMembership: pallet_membership::<Instance2>::{Pallet, Call, Storage, Event<T>, Config<T>} = 18,
 
         // Collator support. the order of these 5 are important and shall not change.
+        // Session must be included *before* AuthorInherent
```
ghzlatarev: can you write explicitly why this is the case, please

runtime/dolphin/src/lib.rs
```diff
@@ -561,6 +562,24 @@ impl pallet_treasury::Config for Runtime {
...
+    /// for each slot in round-robin fashion
+    type PotentialAuthors = CollatorSelection;
+}
+
+impl pallet_author_inherent::Config for Runtime {
+    // We start a new slot each time we see a new relay block.
+    type SlotBeacon = cumulus_pallet_parachain_system::RelaychainBlockNumberProvider<Self>;
+    type AccountLookup = CollatorSelection;
+    type EventHandler = ();
+    type WeightInfo = (); // TODO: Add benchmarked weights
```
ghzlatarev: can you add this the benchmark workflows and maybe add a fake weights file, and we'll replace it with a real one in the release PR.

runtime/dolphin/src/lib.rs
```diff
@@ -707,11 +728,14 @@ construct_runtime!(
         TechnicalMembership: pallet_membership::<Instance2>::{Pallet, Call, Storage, Event<T>, Config<T>} = 18,
 
         // Collator support. the order of these 5 are important and shall not change.
+        // Session must be included *before* AuthorInherent
+        AuthorInherent: pallet_author_inherent::{Pallet, Call, Storage, Inherent} = 60,
+        AuraAuthorFilter: pallet_aura_style_filter::{Pallet, Storage} = 63,
```
ghzlatarev, line 720: Do we really need this pallet if it's just going to call back into `CollatorSelection` ?
`CollatorSelection` -> `AuraAuthorFilter` -> `CollatorSelection`

node/src/command.rs
```diff
@@ -457,6 +454,8 @@ pub fn run_with(cli: Cli) -> Result {
                 let state_version =
                     RelayChainCli::native_runtime_version(&config.chain_spec).state_version();
 
+                // TODO match chainspec
+                // let block: crate::service_nim::Block =
```
ghzlatarev: is this forgotten commented out code ?

runtime/dolphin/src/lib.rs
Garandor: stage 2 will replace `EligibleAuthors` to be pulled from `ParachainStaking` instead of `CollatorSelection`.
It's true though, this does not need to be its own pallet, it's nice to have in order to encapsulate the aura filtering functionality though. PureStake have also [mused on making it a type](https://github.com/PureStake/nimbus/blob/ec08bbeebfc07c790013434bb437376789107d35/pallets/aura-style-filter/src/lib.rs#L51)

I'm using their approach for now as this is simple to remove later since it doesn't have any storage

node/src/command.rs
Garandor: yes, forgot to remove. fixed, thx

runtime/dolphin/src/lib.rs
```diff
@@ -264,6 +265,7 @@ impl Contains<Call> for BaseFilter {
             | Call::TechnicalMembership(_)
             | Call::Scheduler(_)
             | Call::Session(_) // User must be able to set their session key when applying for a collator
+            | Call::AuthorInherent(pallet_author_inherent::Call::kick_off_authorship_validation {..}) // executes unsigned on every block
```
ghzlatarev, line 256: Are the inherents called before the on_initialize hooks ?

pallets/collator-selection/src/lib.rs
```diff
@@ -664,6 +670,72 @@ pub mod pallet {
...
+        fn can_author(account: &T::AccountId, slot: &u32) -> bool {
+            let validator_key = T::ValidatorIdOf::convert(account.clone());
+            if validator_key.is_none()
+                || !T::ValidatorRegistration::is_registered(
+                    &validator_key.expect("we checked against none before. qed"),
+                )
+            {
+                return false;
+            }
+            T::CanAuthor::can_author(account, slot) // filter passed, hand execution to the next pipeline step
```
ghzlatarev, line 739: Can this comment be more explicit, about the next and current step of the pipeline

pallets/collator-selection/src/lib.rs
```diff
@@ -162,6 +163,11 @@ pub mod pallet {
 
         /// The weight information of this pallet.
         type WeightInfo: WeightInfo;
+
+        /// The final word on whether the reported author can author at this height.
+        /// If the pallet that implements this trait depends on an inherent, that inherent **must**
+        /// be included before this one.
+        type CanAuthor: CanAuthor<Self::AccountId>;
```
ghzlatarev, line 170: So in our concrete case the implementor is `AuthorInherent` which depends on `kick_off_authorship_validation` inherent. So `AuthorInherent` must be included before `CollatorSelection` in the `construct_runtime!` macro, is that it?

runtime/dolphin/src/lib.rs
Garandor: Shawn's got the answer [here](https://substrate.stackexchange.com/a/3783/504)
Inherents are called before extrinsics and after `on_initialize` is relsolved.

pallets/collator-selection/src/lib.rs
Garandor: It's somewhat explained [here](https://github.com/PureStake/nimbus/blob/ffb03c0614af1d3649fe52423ba99d98035baf74/pallets/author-inherent/src/lib.rs#L110) with context [here](https://github.com/paritytech/substrate/pull/10128#issuecomment-988237700)

author_inherent injects an inherent that is designed to run *after* all other inherents ( at least the ones nimbus' CanAuthor execution pipeline depends on itself ) but *before* extrinsics.

Inherents are executed in pallet inclusion order, so author_inherent must be the last one in line - in our case since we have `AllPalletsWithSystemReversed`, it must be listed before other pallets providing inherents.

**However**, this only affects *inherents* and neither Session nor CollatorSelection have any 🤡 
They all do their stuff in `on_initialize` which resolves before inherents are built, so in our case, pallet inclusion order does **not** matter.

I'll remove the other comment about sequencing from the code. Thanks for making me think on this a bit more

P.S. Moonbeam needs it for the randomness pallet, which we don't use

runtime/dolphin/src/lib.rs
Garandor: I could if it were the case 🤡 
See comment here for details https://github.com/Manta-Network/Manta/pull/582#discussion_r938256087

i've removed this comment. Thanks for raising the question

pallets/collator-selection/src/lib.rs
Garandor: > the implementor is AuthorInherent 

Actually, *every* pallet implementing `CanAuthor` is meant with this, so `author-inherent`, `manta-collator-selection` and `aura-style-filter` all need proper sequencing with other inherent providing pallets they depend on ( in our case: none )

pallets/collator-selection/src/lib.rs
Garandor: it actually can't, because the pallet doesn't decide the sequencing, the runtime does.
The pallet just calls whatever next filter the runtime has placed in this pallet's `CanAuthor` trait.

The sequencing comments are all in the calamari/dolphin lib.rs

runtime/dolphin/src/lib.rs
ghzlatarev: Thanks, that clears it up. 
So it's `pallet 1 on_init` -> `pallet 1 intrinsic` -> `pallet 2 on_init` -> `pallet 2 intrinsic` etc...

Garandor: actually no, it's three distinct phases

1. all `on_init` hooks by hook order ( IIRC that depends on pallet index ), then
2. all inherents by pallet order until the block is full or no more inherents, then
3. extrinsics until block is full if there was space left after the inherents executed
(4.) `on_finalize` hooks by hook order

ghzlatarev: Ok thanks for clarifying again.
I really haven't found good documentation on this process, besides reading the code, which I don't have the best entry point for...

pallets/collator-selection/src/lib.rs
ghzlatarev: nit: can you mention that the order is decided by the runtime in the comment

runtime/dolphin/src/lib.rs
ghzlatarev: Gotcha, thanks.

.github/workflows/generate_calamari_weights_files.yml
```diff
@@ -90,6 +90,13 @@ jobs:
               id: frame_system
               name: frame_system
             iterations: 20
+          - extrinsic:
+              id: '*'
+              name: pallet_author_inherent
```
Dengjianping, line 95: Should we enable it for dolphin?

pallets/collator-selection/src/lib.rs
```diff
@@ -664,6 +670,88 @@ pub mod pallet {
...
+        {
+            use sp_runtime::traits::Convert;
+            pallet_session::Pallet::<T>::validators()
+                .into_iter()
+                .map(
+                    |session_validator_id: <T as pallet_session::Config>::ValidatorId| {
+                        <T as Config>::AccountIdOf::convert(session_validator_id.into())
+                    },
+                )
+                .collect::<Vec<T::AccountId>>()
```
Dengjianping, line 719: Can it be returned as iterator?

runtime/calamari/src/lib.rs
```diff
@@ -975,6 +1001,21 @@ impl_runtime_apis! {
...
+            System::initialize(&(parent_header.number + 1), &parent_header.hash(), &parent_header.digest);
+
+            // And now the actual prediction call
+            <AuthorInherent as nimbus_primitives::CanAuthor<_>>::can_author(&author, &relay_parent)		}
+    }
+
+    // We also implement the old AuthorFilterAPI to meet the trait bounds on the client side.
+    impl nimbus_primitives::AuthorFilterAPI<Block, NimbusId> for Runtime {
+        fn can_author(_: NimbusId, _: u32, _: &<Block as BlockT>::Header) -> bool {
+            panic!("AuthorFilterAPI is no longer supported. Please update your client.")
```
Dengjianping, line 962: Is anywhere else calling this API? I mean `panic` will cause runtime crash. Not sure it's safe.

runtime/dolphin/src/lib.rs
```diff
@@ -129,7 +130,7 @@ mod weights;
 pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("dolphin"),
     impl_name: create_runtime_str!("dolphin"),
-    authoring_version: 1,
+    authoring_version: 2,
```
Dengjianping, line 120: Not sure this filed should be bumped.

runtime/calamari/Cargo.toml
```diff
@@ -116,12 +117,14 @@ runtime-benchmarks = [
   'frame-system-benchmarking',
   'frame-support/runtime-benchmarks',
   'frame-system/runtime-benchmarks',
+  'manta-collator-selection/runtime-benchmarks',
+  'nimbus-primitives/runtime-benchmarks',
+  'pallet-author-inherent/runtime-benchmarks',
```
ghzlatarev, line 122: You need   'cumulus-pallet-parachain-system/runtime-benchmarks' for the benchmark to work.
[This impl](https://github.com/paritytech/cumulus/blob/3bb82e5126362f9cf3952c60b737c3b71b0a136b/pallets/parachain-system/src/lib.rs#L1113-L1126) of `BlockNumberProvider` is never used since the feature is not enabled.

node/Cargo.toml
```diff
@@ -81,6 +81,11 @@ cumulus-relay-chain-inprocess-interface = { git = 'https://github.com/paritytech
 cumulus-relay-chain-interface = { git = 'https://github.com/paritytech/cumulus.git', branch = "polkadot-v0.9.26" }
 cumulus-relay-chain-rpc-interface = { git = 'https://github.com/paritytech/cumulus.git', branch = "polkadot-v0.9.26" }
 
+# Nimbus dependencies
+nimbus-consensus = { git = "https://github.com/manta-network/nimbus.git", branch = "garandor/polkadot-v0.9.26", default-features = false }
+nimbus-primitives = { git = "https://github.com/manta-network/nimbus.git", branch = "garandor/polkadot-v0.9.26", default-features = false }
+pallet-author-inherent = { git = "https://github.com/manta-network/nimbus.git", branch = "garandor/polkadot-v0.9.26", default-features = false }
```
ghzlatarev: Don't think you need the `default-features = false` here.

.github/workflows/generate_calamari_weights_files.yml
Garandor: good catch, thanks! added

pallets/collator-selection/src/lib.rs
Garandor: Unfortunately, the return type is given by the nimbus trait which we don't control :(

runtime/calamari/Cargo.toml
Garandor: Thanks for pointing to the fix!

runtime/calamari/src/lib.rs
Garandor: Yes, it's meant to crash the runtime because that old API is no longer supported.
We have to provide it because upstream nimbus still has it for backwards compatibility, but doesn't use it anywhere anymore.
Our client also never used this, so we don't care.

runtime/dolphin/src/lib.rs
Garandor: It must, because this release is incompatible with the current block production mechanism.

We normally bump it in the release PR instead, but I think it doesn't hurt to add it here

node/Cargo.toml
Garandor: Thanks, this was bad copypasta

runtime/dolphin/src/lib.rs
Garandor: done

ghzlatarev (dismissed):

node/src/service_aura.rs
```diff
@@ -0,0 +1,748 @@
+// Copyright 2020-2022 Manta Network.
+// This file is part of Manta.
```
Dengjianping, line 2: Can you create a folder like rpc/chain_specs, then move service code to it?
Of cause, we can do it in next release.

runtime/calamari/src/lib.rs
```diff
@@ -921,6 +947,21 @@ impl_runtime_apis! {
         }
     }
 
+    impl nimbus_primitives::NimbusApi<Block> for Runtime {
+        fn can_author(author: NimbusId, relay_parent: u32, parent_header: &<Block as BlockT>::Header) -> bool{
+            System::initialize(&(parent_header.number + 1), &parent_header.hash(), &parent_header.digest);
+
+            // And now the actual prediction call
+            <AuthorInherent as nimbus_primitives::CanAuthor<_>>::can_author(&author, &relay_parent)		}
```
Dengjianping: Can you reformat this part of code? `}` looks weird.

node/src/service_aura.rs
Garandor: This `service_aura` file is temporary and we'll be getting rid of it. In a cleanup we'll put the manta runtime on nimbus too and then this'll be obsolete

runtime/calamari/src/lib.rs
Garandor: Right, dunno how that happened. Fixed it

ghzlatarev (approved):

Dengjianping (approved):

PR #583, opened 2022-06-09
feat: remove checkpoint from PullResponse RPC API

bhgomes:

This PR removes the response checkpoint in `PullResponse` which is no longer used in the RPC API. This does not drop any meaningful information since the checkpoint can be computed from the diff directly.

Requires https://github.com/Manta-Network/manta-rs/pull/86.

---

write a little note why.

- [ ] Wrote unit tests.

### DIFF
--- a/pallets/manta-pay/Cargo.toml
+++ b/pallets/manta-pay/Cargo.toml
@@ -96,18 +96,18 @@ jsonrpc-core-client = { version = "18.0.0", optional = true, default-features = 
 jsonrpc-derive = { version = "18.0.0", optional = true, default-features = false }
 
 # manta dependencies
-manta-accounting = { git = "https://github.com/manta-network/manta-rs.git", branch = "feat/simplify-sync", default-features = false }
-manta-crypto = { git = "https://github.com/manta-network/manta-rs.git", branch = "feat/simplify-sync", default-features = false }
-manta-pay = { git = "https://github.com/manta-network/manta-rs.git", branch = "feat/simplify-sync", default-features = false, features = ["groth16", "scale"] }
+manta-accounting = { git = "https://github.com/manta-network/manta-rs.git", tag = "v0.5.0", default-features = false }
+manta-crypto = { git = "https://github.com/manta-network/manta-rs.git", tag = "v0.5.0", default-features = false }
+manta-pay = { git = "https://github.com/manta-network/manta-rs.git", tag = "v0.5.0", default-features = false, features = ["groth16", "scale"] }
 manta-sdk = { git = "https://github.com/manta-network/sdk.git", default-features = false }
-manta-util = { git = "https://github.com/manta-network/manta-rs.git", branch = "feat/simplify-sync", default-features = false }
+manta-util = { git = "https://github.com/manta-network/manta-rs.git", tag = "v0.5.0", default-features = false }
 manta-primitives = { path = "../../primitives", default-features = false }
 
 [dev-dependencies]
 bencher = "0.1.5"
 criterion = "0.3.4"
 lazy_static = "1.4.0"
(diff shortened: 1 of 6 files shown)


stechu (approved):

primitives/Cargo.toml
```diff
@@ -17,7 +17,7 @@ smallvec = "1.8.0"
 log = "0.4.16"
 
 # manta-rs dependencies
-manta-accounting = { git = "https://github.com/manta-network/manta-rs.git", default-features = false }
+manta-accounting = { git = "https://github.com/manta-network/manta-rs.git", tag = "v0.5.0", default-features = false }
```
Garandor, line 20: do we need to have the `manta-accounting` <=> `manta-rs` difference?
I'd rather have `manta-rs` on both sides

pallets/manta-pay/Cargo.toml
```diff
@@ -96,18 +96,18 @@ jsonrpc-core-client = { version = "18.0.0", optional = true, default-features =
 jsonrpc-derive = { version = "18.0.0", optional = true, default-features = false }
 
 # manta dependencies
-manta-accounting = { git = "https://github.com/manta-network/manta-rs.git", default-features = false }
-manta-crypto = { git = "https://github.com/manta-network/manta-rs.git", default-features = false }
-manta-pay = { git = "https://github.com/manta-network/manta-rs.git", default-features = false, features = ["groth16", "scale"] }
+manta-accounting = { git = "https://github.com/manta-network/manta-rs.git", tag = "v0.5.0", default-features = false }
```
Garandor, line 99: same here

I suppose these name differences are there for historic reasons, aka. let's unify them.
If not now, please create an issue

Garandor (approved):
LGTM

pallets/manta-pay/src/lib.rs
```diff
@@ -486,21 +486,18 @@ pub mod pallet {
...
+				let raw_key = Shards::<T>::hashed_key_for(shard_index, receiver_index as u64 - 1);
 				Shards::<T>::iter_prefix_from(shard_index, raw_key)
 			};
 			for _ in 0..Self::PULL_MAX_PER_SHARD_UPDATE_SIZE {
 				match iter.next() {
-					Some((_, (utxo, encrypted_note))) => {
-						*receiver_index += 1;
-						receivers.push((utxo, encrypted_note));
-					}
+					Some((_, next)) => receivers.push(next),
```
Dengjianping, line 500: Is it possible that hackers use it to attack us? I mean the loop and allocation happen to runtime,  and anyone can send rpc request, if there're too many requests sent, like 10K or even more, runtime might be exploded. Of cause we could set limits on rpc request like only 100/sec, and only enable on full node.

primitives/Cargo.toml
stechu: `manta-rs` is the repo name, and `manta-accounting` is the crate name.
`manta-rs` has many crates.

pallets/manta-pay/Cargo.toml
stechu: see above. It doesn't make sense to put all things in `manta-rs` in the same crates.

pallets/manta-pay/src/lib.rs
stechu: The is outside of the consensus. For example, collators should have RPC disabled.

stechu: We do need to do a proper DDoS tests tho.

Dengjianping: > We do need to do a proper DDoS tests tho.

Definitely.

stechu: https://github.com/Manta-Network/Manta/issues/587

PR #572, opened 2022-06-06
Release v3.2.0

Dengjianping:
Signed-off-by: Dengjianping <djptux@gmail.com>

## Description

closes: #553

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [x] Linked to Github issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [ ] Wrote unit tests.
- [x] Updated relevant documentation in the code.
- [x] Added **one** line describing your change in `<branch>/CHANGELOG.md`
- [x] Re-reviewed `Files changed` in the Github PR explorer.
- [x] If runtime changes, need to update the version numbers properly:
   * `authoring_version`: The version of the authorship interface. An authoring node will not attempt to author blocks unless this is equal to its native runtime.
   * `spec_version`: The version of the runtime specification. A full node will not attempt to use its native runtime in substitute for the on-chain Wasm runtime unless all of spec_name, spec_version, and authoring_version are the same between Wasm and native.
   * `impl_version`: The version of the implementation of the specification. Nodes are free to ignore this; it serves only as an indication that the code is different; as long as the other two versions are the same then while the actual code may be different, it is nonetheless required to do the same thing. Non-consensus-breaking optimizations are about the only changes that could be made which would result in only the impl_version changing.
   * `transaction_version`: The version of the extrinsics interface. This number must be updated in the following circumstances: extrinsic parameters (number, order, or types) have been changed; extrinsics or pallets have been removed; or the pallet order in the construct_runtime! macro or extrinsic order in a pallet has been changed. You can run the `metadata_diff.yml` workflow for help. If this number is updated, then the `spec_version` must also be updated 
- [x] Verify benchmarks & weights have been updated for any modified runtime logics
- [x] If importing a new pallet, choose a proper module index for it, and allow it in `BaseFilter`. Ensure **every** extrinsic works from front-end. If there's corresponding tool, ensure both work for each other.
- [x] If needed, update our Javascript/Typescript APIs. These APIs are officially used by exchanges or community developers.
- [x] If modifying existing runtime storage items, make sure to implement storage migrations for the runtime and test them with `try-runtime`. This includes migrations inherited from upstream changes, and you can search the diffs for modifications of `#[pallet::storage]` items to check for any.

Issue #514, opened 2022-04-23
Ways to shorten dev build time


Dengjianping:
- [ ] Take a look at this [PR](https://github.com/paritytech/polkadot/pull/4807), seems pretty impressive.
- [ ] Try build tool [`fleet`](https://github.com/dimensionhq/fleet).
- [ ] Try [`cargo-nextest`](https://nexte.st/#cargo-nextest).
- [ ] Try [`cargo-hakari`](https://docs.rs/cargo-hakari/latest/cargo_hakari/).

PR #590, opened 2022-06-11
Auto-enforce code formatting

### LINKED ISSUE #307: Clean up clippy warnings
For example in https://github.com/Manta-Network/Manta/runs/4608072266?check_suite_focus=true
There are warnings like:
`warning: casting integer literal to u64 is unnecessary`

Can use `#![allow(clippy::unnecessary_cast)]` for the custom weight files warnings.

### LINKED ISSUE #310: Clippy/fmt warnings should be failures in CI.
For example in https://github.com/Manta-Network/Manta/runs/4608072266?check_suite_focus=true
There are warnings like:
`warning: casting integer literal to u64 is unnecessary`
That should fail the job.

### LINKED ISSUE #584: Achieve automatic linting and consistent formatting
I keep noticing reformats in successive PRs by different people on the same code. two likely reasons
- no auto-linting means first guy simply forgot to fix clippy errors, or format before comitting
- their rust version or `cargo fmt` settings differ from the second guy

We want to automate this, so I'd like to trial run a new tool called https://trunk.io/products/check on our node.
If it works well it will be useful for the other teams too. 
It's a multilinter/multiformatter, supports a lot of stuff from Rust, over docker to credential leaks and spellcheck.

Steps:
1. ~every dev runs `curl https://get.trunk.io -fsSL | bash` *from their manta repository root*~
2. (if you run windows vscode, you can get the extension, linux is not yet supported)
3. You add a pre-commit hook to git on your machine (instructions to follow)
> NOTE
> By default trunk check runs only on the files you've changed.

So this should be reasonably fast
4. there is no 4, whenever you commit, your stuff gets autoformatted and linted (using the same version and config of the linter across everyone, from devs to CI). 
If you want to, you can manually run `trunk check`

Later we can add a `trunk check` to github actions to ensure subsequent commits don't introduce new problems.

on my first trial run i got
`✖ 224 blocking issues` across all files in the repo, including Dockerfiles.

### LINKED ISSUE #586: Align code base with autoformatting rules
We want to have a "big bang" integration of all formatting changes, so that from that point on, introduced changes will be real code changes, not formatting only changes.

Adherence to formatting rules will be enforced by the CI.

Followup to #585

Garandor:
Merge after #572

Only semi-significant change besides the formatting is in the manta/calamari/dolphin RT where i reordered the CallFilter to lead the match with explicitly disallowed extrinsics for safety on refactor.
=> https://github.com/Manta-Network/Manta/pull/590/files#diff-3c21538b15b471a2e041d55e54f471aca028763b85b77a8c30ed31409dd49a27R185

Use https://github.com/Manta-Network/Manta/pull/590/files?diff=unified&w=1 to see only the functional changes

## Description

closes #584 
closes #586 
closes #310 
closes #307 

---

write a little note why.

- [ ] Wrote unit tests.


stechu:
It appears that it reformat all the 2 spaces indent for 4 spaces indent. I think for the sake of not introducing unnecessary diff. I would say keep the 4 spaces indent. (unless a overwhelming amount of runtime devs want to do 2 spaces indent).

Garandor:
Seems that is the rustfmt default setting.
Interestingly, we do have an inconsistency here since taplo formats cargo.toml files with 2 spaces instead.
This makes it somewhat more difficult to write correct-from-the-start code since `.editorconfig` is set to 4 spaces, so default formatting of cargo.toml files will be incorrect.
We can add a section for [*.toml] files to indent by 2 instead, so as long as you have a compliant IDE it'll produce correct formatting.

~~If we wanted to make it consistent, I'd vote for changing cargo.toml's to 4 spaces.~~
I'm voting no-change in indentation and adding a `.editorconfig` change instead.

@ghzlatarev @Dengjianping need your votes :)

Garandor:
This PR is also a step towards #514  as it looks like the `check_build` step is quite a bit faster by simply
- grouping compiling and non-compiling steps together
- skipping WASM build on compilation

In the everything-cached case it runs for 17 mins, otherwise it seems to take 30-40min where previously it took ~1hr

.github/workflows/check_build.yml
```diff
@@ -83,25 +83,26 @@ jobs:
         run: |
```
ghzlatarev, line 83: what about the secrets leak linter ?

runtime/manta/Cargo.toml
```diff
@@ -5,7 +5,7 @@ homepage = 'https://manta.network'
 license = 'GPL-3.0'
 name = 'manta-runtime'
 repository = 'https://github.com/Manta-Network/Manta/'
-version = '3.1.5'
+version = '3.2.0'
```
ghzlatarev, line 8: why is the node version getting bumped?

ghzlatarev (dismissed):

ghzlatarev:
> Seems that is the rustfmt default setting. Interestingly, we do have an inconsistency here since taplo formats cargo.toml files with 2 spaces instead. This makes it somewhat more difficult to write correct-from-the-start code since `.editorconfig` is set to 4 spaces, so default formatting of cargo.toml files will be incorrect. We can add a section for [*.toml] files to indent by 2 instead, so as long as you have a compliant IDE it'll produce correct formatting.
> 
> ~If we wanted to make it consistent, I'd vote for changing cargo.toml's to 4 spaces.~ I'm voting no-change in indentation and adding a `.editorconfig` change instead.
> 
> @ghzlatarev @Dengjianping need your votes :)

I don't have a strong opinion either way. As long as `ctrl-s` still formats for me.

.github/workflows/check_build.yml
Garandor: Won't spend time now to find out how to manually integrate it and leave that as a followup if/when we reintroduce trunk.io

Tbh. having secret leaks detected by the CI is *too late* anyway, as the commit is already public.
This needs to run on the dev PC

Garandor:
The PR is finished, but please don't review yet as the changes from #572 are still listed as changes here.
Marking the PR as draft until it is merged.

When you DO review, use "hide whitespace" mode -> https://github.com/Manta-Network/Manta/pull/590/files?diff=unified&w=1

Garandor (commented):
I looked it over, the changeset looks good to me.
Some of the automatic formatting is a bit wonky where it inserted new-lines fairly pointlessly, but that's autoformatting i guess ¯\_(ツ)_/¯

runtime/calamari/src/lib.rs
```diff
@@ -131,70 +131,91 @@ pub const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(70);
...
+
+        if pallet_tx_pause::PausedTransactionFilter::<Runtime>::contains(call) {
+            // no paused call
+            return false;
+        }
+
+        #[allow(clippy::match_like_matches_macro)]
+        // keep CallFilter with explicit true/false for documentation
+        match call {
+            // Explicitly DISALLOWED calls
```
ghzlatarev, line 189: Some of the comments here don't seem correctly formatted.

ghzlatarev, line 189: What does the explicit disallow protect me from

Garandor: Previously we had disallowed calls listed in the `=> false` branch below the allowed ones.
If one adds a call into that branch while it's still present in the `=> true` branch, the `true` branch will take precedence and we'll have in-fact allowed calls we think are disallowed.

The false -> true fallthrough order makes sure that if we filter a call explicitly, it'll never be overwritten by for example a blanket allow of its pallet.

ghzlatarev: Can't we just have the false arm cover all cases not in the true arm:
```
| _ => false
```
Not sure why there were specifically mentioned calls in the false arm anyway. The original idea is to have a specific whitelist and the rest is not allowed.

Garandor: what do you mean specifically? They look ok to me in github and vscode

Garandor: that default fallthrough is still there at the bottom.

Garandor: > Can't we just have the false arm cover all cases not in the true arm:
> 
> ```
> | _ => false
> ```
> 
> Not sure why there were specifically mentioned calls in the false arm anyway. The original idea is to have a specific whitelist and the rest is not allowed.

this way is more flexible, e.g. for collator_selection where we're blacklisting a single extrinsic, we can blacklist it and allow the full remaining pallet instead of explicitly allowing every extrinsic except the filtered one

I didn't want to make those changes in this PR as they need some thought whether it's really safe to upgrade the explicit calls to a blanket pallet allow

Garandor: ok i concede that they do look wrong in github. they're not indented as far as they should be.
vscode looks normal though. not sure which display source is wrong here

Garandor: it looks like rustfmt did *NOT* automatically convert all leading tabs to spaces.
I ran a simple replace of tabs -> 4 spaces to normalize this.
Strange thing is that it didnt even fail the `rustfmt --check --all`

pallets/asset-manager/src/weights.rs
```diff
@@ -39,124 +39,124 @@
 #![allow(unused_parens)]
 
 use frame_support::{
-	traits::Get,
-	weights::{constants::RocksDbWeight, Weight},
+    traits::Get,
+    weights::{constants::RocksDbWeight, Weight},
```
Dengjianping, line 43: About formatting weights file, you better change the weight template, or we have to format it while every time we re-benchmark runtime. 

https://github.com/Manta-Network/Manta/blob/manta/.github/resources/frame-weight-template.hbs#L32

pallets/asset-manager/src/weights.rs
```diff
@@ -1,16 +1,16 @@
 // Copyright 2020-2022 Manta Network.
 // This file is part of Manta.
-
+//
 // Manta is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
 // the Free Software Foundation, either version 3 of the License, or
 // (at your option) any later version.
-
+//
```
Dengjianping, line 8: And add `//` to empty lines in https://github.com/Manta-Network/Manta/blob/manta/.github/resources/frame-weight-template.hbs

Dengjianping (changes requested):

pallets/asset-manager/src/weights.rs
Garandor: adapted it, thanks!

pallets/asset-manager/src/weights.rs
Garandor: adapted it, thanks!

runtime/calamari/src/lib.rs
Garandor: I tried adding a new change with leading tab instead of spaces. `cargo fmt --all -- --check` fails as expected. Should be good for CI use then.
We're guaranteed nobody introduces new tabs into the codebase.
cargo fmt checks .rs files and taplo checks .tomls

node/src/command.rs
```diff
@@ -122,9 +122,9 @@ impl SubstrateCli for Cli {
     fn description() -> String {
         format!(
             "Manta/Calamari/Dolphin Collator\n\nThe command-line arguments provided first will be \
-		passed to the parachain node, while the arguments provided after -- will be passed \
-		to the relaychain node.\n\n\
-		{} [parachain-args] -- [relaychain-args]",
+        passed to the parachain node, while the arguments provided after -- will be passed \
```
Garandor, line 125: not sure reformatting these multiline strings is a good idea, the default tab width in bash is 8, not 4 so this will visually move the CLI

node/src/command.rs
```diff
@@ -349,7 +349,7 @@ pub fn run_with(cli: Cli) -> Result<()> {
                 runner.sync_run(|config| cmd.run::<Block, DolphinRuntimeExecutor>(config))
             } else {
                 Err("Benchmarking wasn't enabled when building the node. \
-				You can enable it with `--features runtime-benchmarks`."
+                You can enable it with `--features runtime-benchmarks`."
```
Garandor, line 352: Also what about these. Is rust automatically discarding leading whitespace in strings?

Garandor: answering myself: yes it does

node/src/command.rs
Garandor: nevermind, leading space is discarded from nonliteral strings

Garandor (commented):
gone through again, space indented files look sane

Dengjianping (approved):
LGTM!

runtime/calamari/src/lib.rs
ghzlatarev: I think your initial example of having a call in both the true and false arm cannot happen if we just remove the arm with the 3 specifically disallowed extrinsics (which was needed in the first place), and leave the default case as the only false arm.
In that example I can still make a mistake and put calls in multiple arms, just the consequence will be different. With only an allowed/true and default/false arms you won't be able to duplicate and by default new calls be filtered, so you're safe if you forget about the filter.

Garandor: e.g. in this case:
```
// Disallowed
// Allowed
| Call::CollatorSelection(_) => true

// Disallowed
- => false
```
How do you filter ONLY `register_as_candidate` from that pallet.
You'd have to explicitly list all but the filtered extrinsic

Conversely:
```
// Disallowed
// Allowed
| Call::CollatorSelection(
				manta_collator_selection::Call::set_desired_candidates{..}
				| manta_collator_selection::Call::set_candidacy_bond{..}
				| manta_collator_selection::Call::set_eviction_baseline{..}
				=> true
// Disallowed
				| - => false
```
if someone wants to ban remove_collator from a potentially long list of explicitly allowed extrinsics and forgets to remove it from above, allow takes precedence - silently.
I get that simply *removing* the extrinsic from allowed does the job, but that brings us back to case a - what if you want the full pallet allowed except one filtered extrinsic

ghzlatarev: > How do you filter ONLY register_as_candidate from that pallet.

That's the case in manta branch, just explicitly whitelist all except `register_as_candidate`

> if someone wants to ban remove_collator from a potentially long list of explicitly allowed extrinsics and forgets to remove it from above, allow takes precedence - silently.

We don't need an explicit list of disallowed that's my point, they should all just fall into the default case `| _ => false` So in your example you would just remove it from the whitelist, no need to add it to the blacklist explicitly, so you can't make the mistake.

My other point is that even if we have a false, true, false arms I can still forget to move from one place to another, so there will be a silent failure (from what I'm expecting)

Garandor: I see two ways out of this:

a) we prioritize security against pallet upgrades adding new extrinsics and accidentally breaking stuff:
=> We list out _every single extrinsic_ we want to allow, no `Call:Scehduler(_)`, at which point we should move the filter into its own file

b) we prioritize compactness of expressing allowed/banned extrinsics
=> We keep my new approach.

I'm torn myself, we have the CI release script that detects metadata/extrinsic changes to hopefully prevent a), but us being invariant against parity pallets adding new extrinsics with some upgrade is enticing as well

Garandor: > My other point is that even if we have a false, true, false arms I can still forget to move from one place to another, so there will be a silent failure (from what I'm expecting)

My point is that with a prioritized blacklist, this silent failure will be on the **safe** side, having the extrinsic be not callable instead of potentially exposing some sudo-only functionality to users

Garandor: > > How do you filter ONLY register_as_candidate from that pallet.
> 
> That's the case in manta branch, just explicitly whitelist all except `register_as_candidate`

Ok, so either we roll this out across all runtimes **and all pallets**, which is gonna get large, (solution a) from above) or

> We don't need an explicit list of disallowed that's my point, they should all just fall into the default case `| _ => false` So in your example you would just remove it from the whitelist, no need to add it to the blacklist explicitly, so you can't make the mistake.

But you still need to list all other extrinsics, not possible to collapse the case to `Call::CollatorSelection(_)` in that case, so it's *verbose*.
So using b), we can express the same filter much more concisely =>
```
// Disallowed
| Call::CollatorSelection(register_as_candidate) => false
// Allowed
| Call::CollatorSelection(_) => true
// Disallowed
| _ => false
```

ghzlatarev: IMO the most secure is to have everything disallowed by default with the `| _ => false` and having to explicitly add "safe" calls to the allowed list.
a) what's wrong with `Call::Scheduler(_)` when I want all of them? At a point where we want to ban one or more of them we'd have to explicitly list the rest instead of `Call::Schedule(_)`.

Garandor: > a) what's wrong with Call::Scheduler(_) when I want all of them? At a point where we want to ban one or more of them we'd have to explicitly list the rest instead of Call::Scheduler(_).

it's fragile against upgrades of `pallet_scheduler`. If parity add a new extrinsic and we miss that, we'll have a new extrinsic allowed that we didn't know existed and might be exposing functionality that should not be exposed.
*If* we're willing to tolerate that by saying we'll always catch this across all of our dependencies (which is reasonable, since we have automated tooling), approach b) has no downsides over your proposal and adds compact expression of the filter

ghzlatarev: Right I see your point about this, didn't consider it before, better to explicitly write them out instead of `Call::Schedule(_)`

Garandor: Ok cool, let's switch it to an explicit extrinsic list then.

Are you ok with deferring the switch to a followup PR?
If I update it now it'll invalidate Jamie's approval

ghzlatarev: Sure, let's log it and move this along. 
My vote would be for a) with the assumption that we will catch all new extrinsics, so we can keep using (_). At some point you just can't protect from dev mistakes : D

ghzlatarev (approved):

PR #601, opened 2022-06-13
Autoassign new issues to the Development Github project

### LINKED ISSUE #599: Tracking: Improve Dev Tooling
This is a long-running/permanent tracking issue to link other tooling issues and small tool changes that don't need their own PRs against

Garandor:
As it says on the tin
fix #599 

Signed-off-by: Adam Reif <garandor@manta.network>

## Description

write a little note why.

**CI workflow changes only, checklist not required**


.github/workflows/assign_new_issue_to_project
```diff
@@ -0,0 +1,16 @@
+name: Auto Assign Issues to Projects
+
+on: issues
```
ghzlatarev: We should filter out the PRs, if it doesn't already. I think they are considered issues by github.

.github/workflows/assign_new_issue_to_project
```diff
@@ -0,0 +1,16 @@
...
+jobs:
+  assign_to_project:
+    runs-on: ubuntu-latest
+    name: Assign topic to Project
+    steps:
+    - name: Assign issues with new-topic label to GitHub Project
+      uses: srggrs/assign-one-project-github-action@1.2.0
+      with:
+        project: 'https://github.com/orgs/Manta-Network/projects/8'
+        column_name: 'No Status'
```
ghzlatarev: Probably better to just rename the `No Status` to `Backlog` instead of having them both.

Garandor: There's a difference.
No status means "needs decision on when"
Backlog means "declared not important for now"

I'd like to keep new issues we haven't talked about yet in a separate category

.github/workflows/assign_new_issue_to_project
Garandor: Can't find reference to that in the docs
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issues

Let's just try and see if it works, we can always push another fix PR

bhgomes:
@Garandor Does this need to be a `.yml` file?

Garandor:
> @Garandor Does this need to be a `.yml` file?

Thanks, totally didn't notice I forgot the extension. Not sure tbh if it strictly *needs* the extension, but consistency ifself is reason enough to add it.

.github/workflows/assign_new_issue_to_project.yml
```diff
@@ -0,0 +1,19 @@
+name: Auto Assign Issues to Projects
```
ghzlatarev, line 3: can you fix the warnings

ghzlatarev (dismissed):

.github/workflows/assign_new_issue_to_project.yml
Garandor: thanks, done

ghzlatarev (approved):

.github/workflows/assign_new_issue_to_project.yml
```diff
@@ -0,0 +1,22 @@
+---
+
+name: Auto Assign Issues to Projects
+
+# yamllint disable-line rule:truthy
+on:
+  issues:
+    types: opened
```
Dengjianping, line 8: So it doesn't work for PRs, right?

Garandor: From what i can read in the docs, it doesn't. See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issues
there's a separate `pull_request` event.

If it turns out it does, we'll push a followup PR

Dengjianping (approved):
LGTM!

Garandor:
There's a difference.
No status means "needs decision on when"
Backlog means "declared not important for now"

PR #614, opened 2022-06-20
Remove OnRuntimeUpgrade hooks

Dengjianping:

## Description

Remove `OnRuntimeUpgrade` hooks after v3.2.0 has been released.

closes: paritytech/substrate#613 

---

write a little note why.

- [ ] Wrote unit tests.


Garandor:
is this a draft or ready for review?

Garandor:
The idea of using the StorageVersion guards is supposed to enable keeping the migration code in the pallet after migration. 
See https://github.com/paritytech/polkadot-sdk/issues/343

I think we should adopt this best practice and keep executed migrations in the pallet as a sort of "backlog"

ghzlatarev:
> The idea of using the StorageVersion guards is supposed to enable keeping the migration code in the pallet after migration. See [paritytech/polkadot-sdk#343 (comment)](https://github.com/paritytech/polkadot-sdk/issues/343)
> 
> I think we should adopt this best practice and keep executed migrations in the pallet as a sort of "backlog"

Yeah. When we need to use the pallet hook again we can just copy paste the old migration to some separate file, or we can do it now.

Garandor (changes requested):
pls re-request when ready

Dengjianping:
> 

Now it's ready.

Garandor (changes requested):
Could you pls move the migration from the `asset-manager` src to a `migrations` submodule?
We should keep those separate from the active code.
Check `manta-collator-selection` for reference

ghzlatarev (dismissed):

ghzlatarev (dismissed):

Dengjianping:
> Could you pls move the migration from the `asset-manager` src to a `migrations` submodule? We should keep those separate from the active code. Check `manta-collator-selection` for reference

Done.

Garandor (dismissed):
LGTM!

ghzlatarev:
@Dengjianping Merge conflicts again :(
@Garandor we'll need to re-review this when resolved...

ghzlatarev (dismissed):

CHANGELOG.md
```diff
@@ -8,10 +8,11 @@
 ### Features
 
 ### Improvements
-- [\#607](https://github.com/Manta-Network/Manta/pull/607) Turn node client code into library for CLI project
 - [\#449](https://github.com/Manta-Network/Manta/pull/449) Remove strip from CI, and add strip profile to production.
 - [\#576](https://github.com/Manta-Network/Manta/pull/576) Unfilter xtokens.transfer_multicurrencies and bump MaxAssetsForTransfer to 2.
-- [\#609](https://github.com/Manta-Network/Manta/pull/609) Update parameter path from `sdk` to `manta-parameters`
```
Garandor: Found a mistake with the log. Seems i cant add commits to this branch for some reason so i have to leave it here like this. The 609 entry should be 

- [\#622](https://github.com/Manta-Network/Manta/pull/622) Update parameter path from `sdk` to `manta-parameters`.

Garandor (changes requested):

CHANGELOG.md
Dengjianping: Updated

ghzlatarev (approved):

Garandor (approved):

PR #616, opened 2022-06-20
Storage benchmark

stechu:
## Description

This implemented a storage benchmark that is intended to provide base cost measurement of storage operations, especially under large ledger size. The storage cost to be measured:

- [x] point read/write on `StorageDoubleMap` with first key and second key hashed in `Identity`
- [x] point read/write on `StorageDoubleMap` with first key hashed using `TwoxConcat` and second key hashed using `Identity`
- [x] point read/write on `StorageDoubleMap` with first key and second key hashed in `TwoxCancat`

Note: this is a manual test only pallet, it is not designed for CI nor should we generate weight to this. 

Benchmark result:
https://docs.google.com/spreadsheets/d/10BA-gE-NPjoNeCpMB_FmpQ7KkZkjfJVJKaVWCbKDqJg/edit#gid=35004538

closes: #XXXX

---

write a little note why.

- [x] Wrote unit tests.


Garandor:
if this is benchmark only, why the hotfix release milestone?

stechu:
> if this is benchmark only, why the hotfix release milestone?

The hotfix release requires the result from this benchmark.

pallets/storage-bench/src/lib.rs
```diff
@@ -0,0 +1,390 @@
...
+
+/// Ciphertext Type
+pub type Ciphertext = [u8; CIPHER_TEXT_LENGTH];
+
+/// Transfer Proof Type
+pub type Proof = [u8; PROOF_LENGTH];
+
+/// Encrypted Note
+#[cfg_attr(
+    feature = "rpc",
```
Dengjianping, line 60: I didn't find this feature in the toml file, so the rest fo derives should fail when try run `cargo b --all-features`.

pallets/storage-bench/src/lib.rs
```diff
@@ -0,0 +1,390 @@
...
+    #[pallet::pallet]
+    #[pallet::without_storage_info]
+    pub struct Pallet<T>(_);
+
+    #[pallet::hooks]
+    impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {}
+
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {
+        #[pallet::weight(500_000)]
```
Dengjianping, line 167: I see `benchmarking.rs` file, perhaps you should associate weight with `WeightInfo`.

pallets/storage-bench/src/lib.rs
```diff
@@ -0,0 +1,390 @@
...
+        #[transactional]
+        pub fn insert_shard_element_identity_identity(
+            origin: OriginFor<T>,
+            shard_idx: u8,
```
Dengjianping, line 171: `#[pallet::compact]`?
And there're some numbers as params in the rest of extrinsics.

pallets/storage-bench/src/lib.rs
```diff
@@ -0,0 +1,390 @@
...
+            amount: u64,
+        ) -> DispatchResult {
+            T::UpdateOrigin::ensure_origin(origin)?;
+            let mut receivers = Vec::new();
+            let mut iter = if utxo_idx == 0 {
+                ShardsTwoxIdentity::<T>::iter_prefix(shard_idx)
+            } else {
+                let raw_key = ShardsTwoxIdentity::<T>::hashed_key_for(
+                    shard_idx,
+                    Self::encode(utxo_idx as u64 - 1),
```
Dengjianping, line 314: What about inputting `0`?
But if this pallet is just for benchmark, it's fine.

pallets/storage-bench/src/lib.rs
```diff
@@ -0,0 +1,390 @@
...
+            Ok(())
+        }
+
+        pub(crate) fn do_insert_shard_element_twox_twox(
+            shard_idx: u8,
+            utxo: Utxo,
+            encrypted_note: EncryptedNote,
+        ) -> DispatchResult {
+            let utxo_idx = ShardsTwoxTwoxIndices::<T>::get(shard_idx);
+            ShardsTwoxTwox::<T>::insert(shard_idx, utxo_idx, (utxo, encrypted_note.clone()));
```
Dengjianping, line 384: `clone()` might be removable, use reference.

pallets/storage-bench/src/lib.rs
```diff
@@ -0,0 +1,390 @@
...
+                    _ => return Err(Error::<T>::NullStorageItem.into()),
+                }
+            }
+            Self::deposit_event(Event::ShardElementBatchRead(amount));
+            Ok(())
+        }
+    }
+
+    impl<T: Config> Pallet<T> {
+        fn encode(n: u64) -> u64 {
```
Dengjianping, line 330: This is a small function and called in several places, `#[inline]` would be better.

pallets/storage-bench/src/lib.rs
```diff
@@ -0,0 +1,390 @@
...
+                (utxo, encrypted_note),
+            );
+            ShardsIdentityIdentityIndices::<T>::insert(shard_idx, utxo_idx + 1);
+            Ok(())
+        }
+
+        pub(crate) fn do_insert_shard_element_twox_identity(
+            shard_idx: u8,
+            utxo: Utxo,
+            encrypted_note: EncryptedNote,
```
Dengjianping, line 360: Use reference, I see this param is using `clone()` in several places.

runtime/dolphin/src/lib.rs
```diff
@@ -708,6 +713,9 @@ construct_runtime!(
         Assets: pallet_assets::{Pallet, Call, Storage, Event<T>} = 45,
         AssetManager: pallet_asset_manager::{Pallet, Call, Storage, Config<T>, Event<T>} = 46,
         MantaPay: pallet_manta_pay::{Pallet, Call, Storage, Event<T>} = 47,
+
+        // Benchmark
+        StorageBench: pallet_storage_benchmark::{Pallet, Call, Storage, Event<T>} = 48,
```
Garandor, line 718: We should make testnet pallets have a high index by convention, maybe something like `201`.
That way we 
- can write a CI that checks calamari/manta don't have any of these
- don't run into issues with dol having a number we'd want to use for a grouped pallet (e.g. in the consensus or democracy group of ten) occupied by a private pallet

pallets/storage-bench/src/lib.rs
Garandor: this one definitely will need to be fixed to get the PR mergeable

pallets/storage-bench/src/lib.rs
```diff
@@ -0,0 +1,390 @@
...
+        }
+
+        pub(crate) fn do_insert_shard_element_twox_twox(
+            shard_idx: u8,
+            utxo: Utxo,
+            encrypted_note: EncryptedNote,
+        ) -> DispatchResult {
+            let utxo_idx = ShardsTwoxTwoxIndices::<T>::get(shard_idx);
+            ShardsTwoxTwox::<T>::insert(shard_idx, utxo_idx, (utxo, encrypted_note.clone()));
+            ShardsTwoxIdentityIndices::<T>::insert(shard_idx, utxo_idx + 1);
```
Garandor, line 385: why do you insert twoxIdentity in the twoxtwox insert fn?
could this be left over from copy-pasting?
~~seems this would slow down the twoxtwox results needlessly~~
i feel this should be `ShardsTwoxTwoxIndices` instead.
Looks like you're not reading these anywhere in the benchmark, so it might not matter

Garandor (changes requested):
found one potential problem with the benchmark in the comment above + as discussed please refactor into a submodule

ghzlatarev:
@stechu we can close this, right ?

PR #619, opened 2022-06-20
Dolphin CI runtime upgrade test

### LINKED ISSUE #618: Dolphin CI runtime upgrade test

ghzlatarev:
## Description

closes: #618 

* Adds runtime upgrade test for Dolphin in our CI
* Before we would upgrade from an old version, which was not necessarily the previous release version. So now we first update to the previous release version and then to the newest release in order to have this scenario closer to the production environment.

---

write a little note why.

- [x] Wrote unit tests.


bhgomes:
There's no `CHANGELOG.md` update but it was marked as done?

.github/workflows/publish_draft_releases.yml
```diff
@@ -318,6 +321,11 @@ jobs:
             --no-autorestart \
             -- \
             --address=ws://127.0.0.1:9921
+      -
+        if: ${{ needs.check-for-runtime-upgrade.outputs.do-versions-match == 'false' }}
+        name: fetch previous ${{ matrix.runtime.name }}_runtime.compact.compressed.wasm
+        run: |
+          wget -cO - https://github.com/Manta-Network/Manta/releases/download/v3.1.5-1/${{ matrix.runtime.name }}-runtime-v3151.compact.compressed.wasm > ${{ github.workspace }}/${{ matrix.runtime.name }}_runtime_previous.compact.compressed.wasm
```
Garandor, line 324: is hardcoded version number here a good idea?
Whatever this does will need manual change every release, no?

ghzlatarev:
> There's no `CHANGELOG.md` update but it was marked as done?

Oops, corrected.

.github/workflows/publish_draft_releases.yml
ghzlatarev: No. but there's no other way. Either hardcoded release to download or hard-coded branch to build it. Unless I'm missing something better. 
We already hardcode the polkadot binary here, so that was another reason for me to brush that concern off.

Garandor: if updating these is part of release QA i'm okay with keeping it that way.
If not, please add it to the docs ( and then send them to me :) )

You might want to annotate this with a comment in the file (with a link to QA doc) otherwise this question will keep popping up in the future

ghzlatarev: Added a mention to updating these files in the release template

Garandor (dismissed):
LGTM!

node/src/command.rs
```diff
@@ -89,6 +89,7 @@ fn load_spec(id: &str) -> std::result::Result<Box<dyn sc_service::ChainSpec>, St
         "dolphin-dev" => Ok(Box::new(chain_specs::dolphin_development_config())),
         "dolphin-local" => Ok(Box::new(chain_specs::dolphin_local_config())),
         "dolphin-testnet" => Ok(Box::new(chain_specs::dolphin_testnet_config()?)),
+        "dolphin-testnet-ci" => Ok(Box::new(chain_specs::dolphin_testnet_ci_config()?)),
```
Dengjianping, line 92: Not sure you create one more branch here.
what are differences between `genesis/dolphin-testnet-ci-genesis.json` and `genesis/dolphin-testnet-genesis.json`?
I kind of forgot it.

ghzlatarev: So I can just paste the collator mnemonics in the yml file directly for injection, they are different to the normal `-testnet` secrets, which I don't want to leak.

Dengjianping (approved):
LGTM!

Garandor (approved):

PR #628, opened 2022-06-23
New MantaPay Storage Layout and RPC

### LINKED ISSUE #471: Add RPC End-to-End Tests
During our Testnet V2 hacking we were running into issues during RPC calls to the node that were not catchable during the regular testing pipeline. We should add some testing of the actual RPC endpoints, for commands like `submitExtrinsics`.

### LINKED ISSUE #579: Typescript Unittest For Ledger RPC
Add a typescript unittest for ledger RPC.

### Unit Test Spec

Use `sudo` to set the storage to random UTXO and void numbers: for each round add random number (1 to 2) of random `(UTXO, EncryptedNote)` pairs and a random number (0 to 2) of random `VoidNumber`s:
- `Shards`: Add random `(UTXO, EncryptedNote)` pairs with a random first key and increasing second key
- `InsertionOrderedVoidNumbers`: Add random `VoidNumber`s with increasing key

Checkpoint Description:
- `receiver_index: [usize; 256]`: Lengths of each first key in `Shards`
- `sender_index: usize`: Length of `InsertionOrderedVoidNumber`

Testing correctness of Normal Usage:
1. Add random elements to ledger
2. Select a random checkpoint that's below the current one
3. Pull the latest state and check that it returns the data items (in order!) up to the checkpoint returned in the diff
4. If `should_continue == true` then `pull` should be called again with the returned checkpoint (GOTO step 3)
5. If it passes GOTO step 1

Testing correctness of Abnormal Usage: Same as the previous test except try to get from a checkpoint _later_ than the current one. The `pull` should return nothing.

stechu:
## Description

closes: #603 #566 

This PR implements a more optimized MantaPay storage layout from the benchmark result (#616). It contains the following major changes:

- [x] changes the `Identity` hash function in `MantaPay` to `TwoxConcat`
- [x] optimize the ledger RPC implementation by set an optimal parameter and change the pulling algorithm to a greedy algorithm
- [x] update typescript end to end test
- [x] unit test for RPC

Benchmark data: https://docs.google.com/spreadsheets/d/10BA-gE-NPjoNeCpMB_FmpQ7KkZkjfJVJKaVWCbKDqJg/edit#gid=0
---

write a little note why.

- [x] Wrote unit tests.


pallets/manta-pay/src/lib.rs
```diff
@@ -487,42 +506,57 @@ pub mod pallet {
...
                 }
             }
-            (iter.next().is_some(), senders)
+            (
+                VoidNumberSetInsertionOrder::<T>::contains_key(max_sender_index as u64),
+                senders,
+            )
         }
 
         /// Returns the diff of ledger state since the given `checkpoint`.
```
Garandor: pls extend doccomment for the new params

pallets/manta-pay/src/lib.rs
```diff
@@ -145,30 +143,31 @@ pub mod pallet {
     /// UTXOs and Encrypted Notes Grouped by Shard
     #[pallet::storage]
     pub(super) type Shards<T: Config> =
-        StorageDoubleMap<_, Identity, u8, Identity, u64, (Utxo, EncryptedNote), ValueQuery>;
+        StorageDoubleMap<_, Twox64Concat, u8, Twox64Concat, u64, (Utxo, EncryptedNote), ValueQuery>;
```
Garandor, line 146: Are we resetting testnet for this? if not this'll need a migration

pallets/manta-pay/src/lib.rs
```diff
@@ -465,18 +464,38 @@ pub mod pallet {
         T: Config,
     {
         /// Maximum Number of Updates per Shard
-        const PULL_MAX_PER_SHARD_UPDATE_SIZE: usize = 128;
+        const PULL_MAX_RECEIVER_UPDATE_SIZE: u64 = 32768;
```
Garandor, line 467: How about `i16::MAX`?
is this number arbitrary?

same below

pallets/manta-pay/src/lib.rs
```diff
@@ -487,42 +506,57 @@ pub mod pallet {
...
         /// Pulls sender data from the ledger starting at the `sender_index`.
         #[inline]
-        fn pull_senders(sender_index: usize) -> (bool, SenderChunk) {
+        fn pull_senders(sender_index: usize, max_update_request: u64) -> (bool, SenderChunk) {
             let mut senders = Vec::new();
-            let mut iter = VoidNumberSetInsertionOrder::<T>::iter().skip(sender_index);
-            for _ in 0..Self::PULL_MAX_SENDER_UPDATE_SIZE {
-                match iter.next() {
-                    Some((_, next)) => senders.push(next),
+            let max_sender_index = if max_update_request > Self::PULL_MAX_SENDER_UPDATE_SIZE {
```
Garandor, line 535: I think we should either
- just fail the request (and return the node's max update size in the error) instead of silently returning fewer results than the caller expected
or 
- clearly document in the ledger_pull rpc doccomment that this call can return fewer than the requested amount.

tests/README.md
```diff
@@ -3,5 +3,11 @@ Functional Tests for Manta
...
-```
\ No newline at end of file
+yarn ts-node manta_pay_rpc.ts --address=some_address
+```
+The `address` is optional, with default `ws://127.0.0.1:9801`
+
+Current tests:
+
+1. `check_single_rpc_performance` starting from 0 indices, which should not take more than ~200ms
+1. `check_full_sync_order_and_performance` which should be ordered correctly
```
Kevingislason: nit: 2

tests/manta_pay_rpc.ts
```diff
@@ -0,0 +1,303 @@
...
+const SHARD_NUMBER: number = 256;
+const PULL_MAX_SENDER_UPDATE_SIZE: number = 1024;
+const PULL_MAX_PER_SHARD_UPDATE_SIZE: number = 128;
+
+// generate utxo deterministically using shard_idx and utxo_idx
+function generate_utxo(shard_idx: number, utxo_idx: number): Uint8Array {
+    const utxo = new Uint8Array(32+32+68);
+    let mod = (shard_idx + utxo_idx) % 256;
+    // we need to avoid fill all 0s, since all 0s means empty
+    var element = ( mod === 0 ? 42 : mod );
```
Kevingislason: nit: `let` is preferred over `var` and can always be used instead of it

Kevingislason (dismissed):
LGTM

pallets/manta-pay/src/lib.rs
```diff
@@ -487,42 +506,57 @@ pub mod pallet {
         fn pull_receivers_for_shard(
             shard_index: u8,
             receiver_index: usize,
+            max_update: u64,
             receivers: &mut ReceiverChunk,
+            receiver_pulled: &mut u64,
```
Garandor:
```suggestion
            receivers_pulled_count: &mut u64,
```
along with its uses

pallets/manta-pay/src/rpc.rs
```diff
@@ -32,7 +32,12 @@ pub trait PullApi {
     /// Returns the update required to be synchronized with the ledger starting from
     /// `checkpoint`.
```
Garandor: actually this doccomment is probably more important to annotate with the new params

Garandor (changes requested):
Partial review of non-test code

Garandor:
just noticed the benchmark results only have 2 datasets. Shouldnt it be 3?
- identityidentity
- twoxidentity
- twoxtwox

pallets/manta-pay/src/lib.rs
stechu: This is a proper number after testing. It cannot be bigger.

pallets/manta-pay/src/lib.rs
stechu: The fact that it is a count  is implied in the type. Will change to plural form tho.

pallets/manta-pay/src/lib.rs
stechu: The name clearly said it is `max_update_request`, it is a "max" value and it is a "request".  Also, please take a look at how RPC works, the returned result contains a boolean field to indicate whether further pull is needed. So failing the request is a bad idea. 

And yes, we need a proper RPC documentation.

pallets/manta-pay/src/lib.rs
stechu: Yes, storage migration is needed. But since it is a testnet, we will do it mannually.

pallets/manta-pay/src/lib.rs
Garandor: Could you please put that reasoning in a comment?

pallets/manta-pay/src/lib.rs
Garandor: Failing malformed requests sounds very reasonable to me in Server/Client communication.
Wasn't the main point of this comment tho, I'm happy as long as there's an APIdoc showing correct usage of the call so client implementors know what outputs to handle

pallets/manta-pay/src/lib.rs
Garandor: disagreeing on the technicality that numeric type doesn't imply count, especially not in the singular variable name form - when i read this I thought it meant a receiver index.
i personally favor the explicit naming, but simple plural works too

tests/README.md
ghzlatarev: I don't think the full sync test is ported yet from the other PR

pallets/manta-pay/src/lib.rs
```diff
@@ -465,18 +464,38 @@ pub mod pallet {
...
+                Self::PULL_MAX_RECEIVER_UPDATE_SIZE
+            } else {
+                max_update_request
+            };
+
             for (i, index) in receiver_index.into_iter().enumerate() {
-                more_receivers |= Self::pull_receivers_for_shard(i as u8, index, &mut receivers);
+                more_receivers |= Self::pull_receivers_for_shard(
+                    i as u8,
+                    index,
```
ghzlatarev: nit: can we rename `i` and `index` to something explicit like `shard_index` + `utxo/receiver_index`

pallets/manta-pay/src/lib.rs
```diff
@@ -465,18 +464,38 @@ pub mod pallet {
         T: Config,
     {
         /// Maximum Number of Updates per Shard
-        const PULL_MAX_PER_SHARD_UPDATE_SIZE: usize = 128;
+        const PULL_MAX_RECEIVER_UPDATE_SIZE: u64 = 32768;
 
         /// Maximum Size of Sender Data Update
-        const PULL_MAX_SENDER_UPDATE_SIZE: usize = 1024;
+        const PULL_MAX_SENDER_UPDATE_SIZE: u64 = 32768;
```
ghzlatarev, line 470: nit: comment how these values were chosen

pallets/manta-pay/src/lib.rs
```diff
@@ -465,18 +464,38 @@ pub mod pallet {
...
 
         /// Pulls receiver data from the ledger starting at the `receiver_index`.
         #[inline]
-        fn pull_receivers(receiver_index: [usize; 256]) -> (bool, ReceiverChunk) {
+        fn pull_receivers(
```
ghzlatarev, line 476: `receiver_index` should probably be `receiver_indices` 
Also I'd mention in a comment or in the description that the new logic pulls all `utxo`s in a `shard` before moving on to the next `shard`, instead of pulling data from all shards at a time.

ghzlatarev (approved):

Garandor (approved):
Approved after discussion

PR #636, opened 2022-06-30
Equalize Barrier impl between Calamari/Dolphin and mock parachain for XCM tests

### LINKED ISSUE #442: Improve XCM tests

ghzlatarev:
## Description

closes: #442 

* Equalize xcm `Barrier` configurations between production and test environment, to get the test cases closer to reality.
* Remove unnecessary Barrier which was supposed to be letting through messages from the relay's executive plurality. Even though the message is let through we don't have a convereter for `Plurality { id: BodyId::Executive, .. }` multilocation, so it's not going to be interpreted.
* Adjust tests to account for Barriers

---

write a little note why.

- [x] Wrote unit tests.


runtime/common/tests/xcm_mock/parachain.rs
```diff
@@ -209,8 +210,35 @@ pub type MultiAssetTransactor = MultiAssetAdapter<
...
+        MultiLocation { parents: 1, interior: Here }
+    };
+}
+match_type! {
+    pub type ParentOrSiblings: impl Contains<MultiLocation> = {
+        MultiLocation { parents: 1, interior: Here } |
+        MultiLocation { parents: 1, interior: X1(_) }
+    };
+}
+pub type Barrier = (
```
Garandor, line 226: could you type alias to a runtime's barrier instead of recreating the type here?
If the code is meant to be identical, there's no need to duplicate it, right?

Garandor (dismissed):
LGTM

runtime/common/tests/xcm_mock/parachain.rs
ghzlatarev: Yeah we could directly import and use either the Dolphin or Calamari barrier. 
I'm going to log this and think about this a bit more, because we could import a lot of other configurations as well, or just move things into the Dolphin or Calamari folders. But not the ones which are generic over the Runtime, so it will get really mixed.

Garandor (dismissed):

runtime/common/tests/xcm_mock/parachain.rs
```diff
@@ -18,7 +18,7 @@
 
 use codec::{Decode, Encode};
 use frame_support::{
-    assert_ok, construct_runtime,
+    assert_ok, construct_runtime, match_type,
```
Dengjianping: This macro is going to be deprecated, please `match_types`.

ghzlatarev: This branch is still on 0.9.18

runtime/common/tests/xcm_mock/parachain.rs
```diff
@@ -209,8 +210,35 @@ pub type MultiAssetTransactor = MultiAssetAdapter<
...
+    };
+}
+pub type Barrier = (
+    // Allows local origin messages which call weight_credit >= weight_limit.
+    TakeWeightCredit,
+    // Allows non-local origin messages, for example from from the xcmp queue,
+    // which have the ability to deposit assets and pay for their own execution.
+    AllowTopLevelPaidExecutionFrom<Everything>,
+    // Parent root gets free execution
+    AllowUnpaidExecutionFrom<ParentLocation>,
```
Dengjianping, line 233: Does this line mean relaychain can send xcm to us without paid?

runtime/common/tests/xcm_tests.rs
```diff
@@ -39,12 +39,51 @@ use crate::xcm_mock::parachain::{
...
+            weight_limit: Limited(3999999999),
+        },
+        DepositAsset {
+            assets: Wild(All),
+            max_assets: 1,
+            beneficiary: MultiLocation {
+                parents: 0,
+                interior: X1(AccountId32 {
+                    network: Any,
+                    id: [
```
Dengjianping, line 75: Who is it?

runtime/common/tests/xcm_tests.rs
```diff
@@ -132,9 +200,11 @@ fn xcmp() {
         ));
     });
 
+    // The `AllowUnpaidExecutionFrom<ParentLocation>` barrier implementation
```
Dengjianping, line 203: So now, we only execute the xcm which contains `Transact` from relaychain?

runtime/common/tests/xcm_mock/parachain.rs
ghzlatarev: Yes, this has been the case so far. This basically means from the root because only the root can send messages without `ClearOrigin`

runtime/common/tests/xcm_tests.rs
ghzlatarev: Just a random account, it doesn't matter.

runtime/common/tests/xcm_tests.rs
ghzlatarev: No, this is just one of the barriers.

Dengjianping (dismissed):
LGTM!

Dengjianping (approved):
lgtm!

Garandor (approved):

PR #638, opened 2022-06-30
`existence_requirement` argument for `FungibleLedger` trait functions

### LINKED ISSUE #637: `keep_alive` argument for `FungibleLedger` trait functions
Need this argument for `burn`, `can_withdraw` and `transfer`

Argument usage should be consistent between MantaPay and XCM cases.

ghzlatarev:
## Description

closes: #637 

* Add `existence_requirement` enum argument to three of the functions in `FungibleLedger` trait: `burn`, `can_withdraw`, `transfer`
* For MantaPay cases `KeepAlive` is now used everywhere, whereas before some calls were with `KeepAlive` and others were `AllowDeath`
* For XCM cases `AllowDeath` is used, as was prior to this PR.
* Renamed `can_withdraw` to `can_reduce_by_amount` and removed all related error codes. There's only 1 error code now `CannotReduceMoreThan(Balance)`
* The reason for renaming was that it no longer uses fungible/s `can_withdraw` but `reducible_amount` because it accepts a `keep_alive` argument.
* In our `burn` function, which uses fungble/s `burn_from` the `existence_requirement` is checked in the `can_reduce_by_amount` prior to calling `burn_from`, which doesn't accept the `existence_requirement` argument, but uses `AllowDeath` internally. In the case of a user selecting `KeepAlive` the function can exit early.
* Added more integration tests to cover new cases.

---

write a little note why.

- [x] Wrote unit tests.


primitives/src/assets.rs
```diff
@@ -424,17 +399,32 @@ where
...
     ) -> Result<(), FungibleLedgerError> {
         Self::ensure_valid(asset_id)?;
-        FungibleLedgerError::from_withdraw(if asset_id == A::NativeAssetId::get() {
-            <Native as FungibleInspect<C::AccountId>>::can_withdraw(account, amount)
+
+        let keep_alive = match existence_requirement {
+            ExistenceRequirement::KeepAlive => true,
+            ExistenceRequirement::AllowDeath => false,
+        };
+        let reducible_amount = if asset_id == A::NativeAssetId::get() {
```
Dengjianping, line 414: This means available amount from users, right? Doesn't include like locked or reserved tokens.

pallets/manta-pay/src/lib.rs
```diff
@@ -806,13 +808,18 @@ where
...
+                    &account_id,
+                    withdraw.0,
+                    ExistenceRequirement::KeepAlive,
+                )
+                .map(|_| WrapPair(account_id.clone(), withdraw))
+                .map_err(|_| InvalidSourceAccount {
+                    account_id,
+                    asset_id,
+                    withdraw,
+                })
```
Dengjianping, line 822: It doesn't relate to this PR, I just kind don't like returning `Vec` directly. Perhaps iterator would be better. Because iterator behaves laziness, but `collect` will allocate memory on head directly.

primitives/src/assets.rs
ghzlatarev: Locked or reserved tokens cannot be withdrawn so they are not part of the reducible amount,

pallets/manta-pay/src/lib.rs
ghzlatarev: @bhgomes

pallets/manta-pay/src/lib.rs
```diff
@@ -806,13 +808,18 @@ where
...
-                    .map(|_| WrapPair(account_id.clone(), withdraw))
-                    .map_err(|_| InvalidSourceAccount {
-                        account_id,
-                        asset_id,
-                        withdraw,
-                    })
+                FungibleLedger::<T>::can_reduce_by_amount(
+                    asset_id.0,
+                    &account_id,
+                    withdraw.0,
```
Garandor, line 814: For me this tuple access is confusing to read especially on withdraw, but also on asset_id.
Any chance we can make these named accessors  in the future?

Garandor (dismissed):
A question, but the logic LGTM

pallets/manta-pay/src/lib.rs
ghzlatarev: This change is actually not from this PR, it's already in master and manta-rs. They probably can be named, what do you think @bhgomes

bhgomes: This is because `AssetId` and `AssetValue` are wrapped types in `manta-rs`, so we just unwrap them by one level. It's a 1-tuple.

pallets/manta-pay/src/lib.rs
bhgomes: It's quite complicated to return iterators on trait methods. Also, there is no advantage to laziness here this is a very small data structure. The memory overhead is tiny compared to the real costs here (ZKP verification). Memory allocation is the least of our concerns in the extrinsics right now.

Garandor (dismissed):

Dengjianping (dismissed):
LGTM!

Garandor (approved):

Dengjianping (approved):
LGTM!

Issue #661, opened 2022-07-08
Automate Changelog Generation


bhgomes:
**This is part of an effort to remove pain points in the merging process, especially as we are growing the team and have a very long CI pipeline that needs to run as smoothly as possible.**

To reduce the number of merge conflicts (and so the number of CI-reruns) that occur while merging PRs we should automate the entire changelog generation process. We add a script and a GitHub action to generate/verify the changelog:

## Constraints

1. The script may not need commit privileges to the manta repo ( read only access token )
2. We will in the future enforce `release-xX.Y.Z-something`

## Ideas

- One binary, with create and verify mode - alternative:
- One binary, **create mode only**
  - when creating, you run the script and commit the change to branch
  - when verifying, CI runs the script => new changelog is generated => `diff  output.md Changelog.md` => if `diff` returns empty, return success

## Generate Changelog Diff

1. Find the previous tag.
3. Find all merged PRs since that tag (PR number, PR title, labels)
4. Bucket by `L-` label, if no label exit
5. New changelog section for the current tag
6. Generate changelog entry per (number, title) pair in its bucket
7. Generate a new changelog footer.

## Steps

0. Come up with a sane and "future-proof" changelog layout and labelling system for PRs
  1. e.g. `https://keepachangelog.com/en/1.0.0`
1. Create locally executable tool that creates the modified changelog for us ( may have dependencies to `git` and `gh` )
  1. know what the next release is going to be - parse the version from the `release-vX.Y.Z-something` branchname. If the branch name doesn't parse, return failure
  2. (Optional) Check that the version number is strictly increasing
  4. skip changelog generation and verification if the release PR has a `L-nochangelog` ????
  5. how to get the commits since last release
  6. how to fetch the labels
  7. how to write those commits to the new Changelog
2. Create verification utility
  1. Have a github workflow - may be an action, doesn't have to - that executes the binary from step 1 => store output.md
  2. as part of that same workflow - compare `output.md` with `changelog.md`, return success if identical
3. Make sure this workflow triggers on release branches only
4. ( Maybe adapt for extra sections in the release notes on GH )

FUTURE:

1. Support withdrawing releases:
   2. check old sections of the changelog
   3. pull their release PR
   4. check if it has `L-withdrawn` or something like that
   5. rewrite the old section to add [ YANKED ] 

## Verify Changelog Diff

1. If the branch name is `release-vX.Y.Z` continue otherwise skip with return value `0`
6. Run the generation script and compare equal to the current PR

[...]

Issue #600, opened 2022-06-13
Tracking: Improve documentation


Garandor:
This is a long-running/permanent tracking issue to link other docs issues and small tool changes that don't need their own PRs against

Issue #669, opened 2022-07-11
Nimbus Consensus Migration Stage 1


Garandor:
This is stage 1 of the phased Nimbus consensus rollout described in https://github.com/Manta-Network/Manta/issues/486

We aim to enable the `nimbus-aura` consensus mechanism on dolphin

It will consist of 2 distinct runtime upgrades implemented in succession:
- [x]  Switch to nimbus
    - [x]  Merge code that will enable nimbus-aura & disable cumulus-aura #582 
    - [x] Merge code that will enable to new client to sync from genesis and verify Aura blocks correctly #763
    - [x]  client upgrade 2 manta-run nodes
    - [x]  runtime upgrade
    - [x]  Upgrade the rest of the nodes
- [ ] #740
    - [ ]  Merge a migration that will call `upgrade_keys` to remove aura keys
    - [ ]  optional: add/enable `parachain-staking` ( this would merge stage 2 into this )
    - [ ]  runtime_upgrade

Tests:
- [x]  Deploy 3.2.1 testnet, runtime upgrade to 3.3.0
- [x]  Add collator
- [x]  Remove collator

PR #678, opened 2022-07-12
Minor: Update PULL_REQUEST_TEMPLATE.md

Garandor:
Signed-off-by: Adam Reif <garandor@manta.network>

## Description

I would like everyone to start using the labelling system on our issues and PRs.
Later (#661 ) this'll enable us to auto-generate the Changelog.
This is an RFC, pls discuss if we should any more mandatory steps @ghzlatarev @Dengjianping

Two goals:
- Clear scope of Milestones and resulting releases (changes only after *discussion*)
- Easy filtering and resurfacing of related issues

relates to: #600  

---

Before we can merge this PR, make sure that __all__ the following items have been
checked off.

- [x] Added *exactly* one label from the `L-` group against this PR
- [x] Added *at least* one label from the `A-` and one label from the `C-` group against either this PR or its associated issue
- [x] Either this PR or its associated issue is targeted against the *current* Milestone ( otherwise discuss if it can be added )

Situational Notes:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -4,12 +4,11 @@ relates to OR closes: #XXXX
...
 
-Before we can merge this PR, please make sure that all the following items have been
-checked off. If any of the checklist items are not applicable, please leave them but
-write a little note why.
-
+Before we can approve this PR for merge, please make sure that **all** the following items have been checked off:
 - [ ] Linked to Github issue with discussion and accepted design OR have an explanation in the PR that describes this work.
-- [ ] Added **one** line describing your change in `<branch>/CHANGELOG.md`
+- [ ] Added *exactly* one label from the `L-` group against this PR
+- [ ] Added *at least* one label from the `A-` and one label from the `C-` group against either this PR or its associated issue
```
Garandor: This `either this PR or its associated issue` stuff is clumsy and prone to error.

I think we need to figure out a way to resolve it so all the info ( all labels and the milestone ) is always in the PR
( except for the problem description, that may remain in the linked issue)

Looking for ideas @ghzlatarev @Dengjianping @Apokalip @flame4

Dengjianping: How can PR be categorized? I suppose some PRs are bug fix, some are new features, and some are improvements, some are for dolphin only.

Dengjianping: Maybe we need to introduce more labels like:

`Added`: for new features.
`Changed`:  for changes in existing functionality.
`Deprecated`: for soon-to-be removed features.
`Removed`: for now removed features.
`Fixed`: for any bug fixes.
`Security`: fix a security issue.
`Dolphin`
`Calamari`
`Manta`

Garandor: > Maybe we need to introduce more labels like:
> 
> `Added`: for new features. `Changed`: for changes in existing functionality. `Deprecated`: for soon-to-be removed features. `Removed`: for now removed features. `Fixed`: for any bug fixes. `Security`: fix a security issue. `Dolphin` `Calamari` `Manta`

I think @Apokalip is doing that in his #661 PR
The labels he added all begin with `L-` and are already available ( i also referenced them in the change above )

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -4,12 +4,11 @@ relates to OR closes: #XXXX
...
+- [ ] Either this PR or its associated issue is targeted against the *current* Milestone ( otherwise discuss if it can be added )
```
ghzlatarev: I don't think putting the milestone on the PR will show it on the milestone board, unless it also has a Project. So maybe mention that as well. But these special cases would just make things more confusing imo.

I'm not quite sure what you're optimizing for, but if it's visibility I'd still prefer to only have issues on the milestone board with linked PRs. That way everything is shown, without duplication, in 1 view (except the long standing issues, which are a separate view)

Garandor: Yeah, i agree - I'm looking for ideas to design this in a way that

a) avoids duplication of info
b) contains all the needed info for the CHANGELOG in the PR
c) handles the special cases where a PR does not close an issue
d) has no ambiguity, no `either or`s or `if`s

It's likely that all these at the same time are not possible, in which case i'd be willing to drop a) if we can keep d)

Garandor: i just saw we can filter the milestone list with `is:issue` and `is:PR` for planning and merge-date sync respectively.

Given that, I think we should be adding both PRs and Issues to the project
image

Dengjianping: Currently, we don't distinguish which PR is for calamari or which PR is for dolphin, which is for client or runtime.
Like polkadot, there're prefixes like `P` means `Polkadot`, `S` means `Substrate`

Garandor: with the PRs we have the `A-calamari` `A-manta` and `A-dolphin`

we could apply them to issues as well

Dengjianping (approved):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -4,14 +4,12 @@ relates to OR closes: #XXXX
...
-Before we can merge this PR, please make sure that all the following items have been
-checked off. If any of the checklist items are not applicable, please leave them but
-write a little note why.
-
+Before we can approve this PR for merge, please make sure that **all** the following items have been checked off:
 - [ ] Linked to Github issue with discussion and accepted design OR have an explanation in the PR that describes this work.
 - [ ] Added **one** label out of the `L-` group to this PR
 - [ ] Added **one or more** labels from the `A-` and `C-` groups to this PR
 - [ ] Explicitly labelled `A-calamari`, `A-dolphin` and/or `A-manta` if your changes are meant for/impact either of these (CI depends on it)
+- [ ] This PR is targeted against the *current*  Milestone ( otherwise discuss if it can be added in time)
```
ghzlatarev, line 12: this pr or the linked issue*

Garandor: i think we should sharpen that rule to *this PR*
Every PR that gets merged should have a milestone, i don't want to have to search through the issues to find why someone merged it in version x

We can still add the milestone to the issue, but this file only concerns itself with whether a PR is ready for merge

ghzlatarev: Then the board will be filled with duplicates, that was the reason we don't add it to PRs.
Where are you looking for RPs anyway ?

Garandor: Check the board, we have it split between issues and PRs now!

Garandor: > Where are you looking for RPs anyway ?

e.g. Clicking into the merge commit from a `git blame`

ghzlatarev: Right I know there are 2 but the issues one will still have duplicates since PRs are also considered Issues.

ghzlatarev: Actually nvm, that's not the case, so it's fine to tag both the issue and the linked PR in a milestone.

ghzlatarev: But still I would rather not have these personal preferences how we track work becoming a rule.
eg: right now i'm also tagging issues where the milestones are missing and linking PRs as best as I can can so I can see everything in the issues board.

Garandor: This is an attempt at standardization.

We've seen tons of inconsistencies with how we do these and asking every PR to be milestoned is a simple rule that will need no exceptions or deep thought to stick to and produces a consistent look on the Projects board.

The fact that you have to spend effort to only come up with a "as best as I can" result is evidence that this system needs to be changed.

If you can articulate a case where adopting this rule would make the current situation *worse*, I'm willing leave it to *personal preference*.

ghzlatarev: Having many boards to track is not convenient.
If you're looking for a team wide thing, I'd prefer to return to our previous standard way of tracking issues with linked prs in a single board.

Garandor: why are you up lol

> If you're looking for a team wide thing, I'd prefer to return to our previous standard way of tracking issues with linked prs in a single board.

The problem with that as you know is that there's no way of linking PRs to issues without closing them, making Issues that have multiple PRs like https://github.com/Manta-Network/Manta/issues/669 impossible to use

flame4 (approved):

.github/PULL_REQUEST_TEMPLATE.md
ghzlatarev: It's not impossible , reopening issues is definitely less convenient than having to always track two boards. Also having the long standing issues totally separated is also inconvenient, that's just another board to track.
Anyway let's just merge this since everyone is ok with it and I don't mind it as well. Will just make a new PR for my preference as well.

Garandor: i think a way to have long-running issues work fine is by just rolling them from milestone to milestone
That way they'll always show up in the current issues board

I agree impossible was too strongly worded 😅

PR #681, opened 2022-07-14
CI Ledger RPC Tests

### LINKED ISSUE #644: Ledger RPC CI
- [ ] Integrate the new ledger RPC type script tests into Manta's CI.

- [ ] Full sync test still missing

ghzlatarev:
## Description

closes: #644 

* Made the ledger-rpc-performance test a mocha test like the other ledger-rpc-correctness test
* Added both tests to the CI, which expects the success return code 0
* Both tests have a setup phase injecting items into storage and testing phase checking correctness or performance, In the performance test, both stages have a different timeout in order to make the test more robust.
* Split the integration test runner in publish-draft-release workflow to 2 runners. One for calamari and one for dolphin. There are tests like the these ledger-rpc-tests that only apply to one of the runtimes.
* Always runt he Dolphin integration test, but only run the Calamari inegration test on release tags or pushes to `manta`, in order to reduce unnecessary CI workload

---

write a little note why.

Situational Notes:


Garandor:
`relates to OR closes: https://github.com/Manta-Network/Manta/issues/644`
pick one

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 - [\#652](https://github.com/Manta-Network/Manta/pull/652) Reduce CI failure rate by switching AWS CI runners from AMD to Intel
 - [\#653](https://github.com/Manta-Network/Manta/pull/653) Add concurrency groups for pull request CI builds to reduce CI costs
 - [\#657](https://github.com/Manta-Network/Manta/pull/657) retire manta-pc-launch with polkadot-launch.
+- [\#657](https://github.com/Manta-Network/Manta/pull/681) CI Ledger RPC Tests.
```
Garandor: 681

Garandor (commented):
Does it make sense to have both dolphin and KMA integration tests run on every commit to every PR?

I recommend to split this workflow and do KMA checks only
- on release branches
- on master ( as the final verification that a merged PR actually works on KMA as well )

Splitting checks in two, with fast checks being done on all PR commits and the comprehensive ones limited to commits on/merges to master is a common mode of operation and since we don't deploy from master immediately, we can afford a merge breaking master CI.

If we need to rework, we can just re-open the merged PR

I recommend:
- This workflow shouldnt run at all on draft PRs
- Only the dolphin integration should be checked on PRs
- KMA checks run on master AND non-draft release-xyz branches

CC @bhgomes

ghzlatarev:
>I recommend to split this workflow and do KMA checks only
>* on release branches
>* on master ( as the final verification that a merged PR actually works on KMA as well )

Thanks, that's reasonable I think, added the change in this PR.

tests/rpc_performance_test.ts
```diff
@@ -0,0 +1,64 @@
+import { ApiPromise, WsProvider } from '@polkadot/api';
+import { Keyring } from '@polkadot/keyring';
+import { manta_pay_types, rpc_api } from './types';
+import { StoragePrepareConfig, setup_storage, manta_pay_config} from './manta_pay';
+import minimist, { ParsedArgs } from 'minimist';
+import { performance } from 'perf_hooks';
+import { expect } from 'chai';
+
+const test_config = {
+    ws_address: "ws://127.0.0.1:9800",
```
Dengjianping, line 10: Can the configuration be put in a toml file?

ghzlatarev: Maybe yeah. @Apokalip can look into while implementing the framework.

Apokalip: @Dengjianping Do you mean like a config toml containing all test configurations or smth like that? If it's that it should be possible with typescript I think, but I do need to look into it. Looking at moonbeam they have their config in typescript hard coded (tests/utils/accounts.ts) for example

Dengjianping: Sometimes, I start a node with port like `9988`, so I have to modify the port in related source code.

Apokalip: Well if we put it in typescript we can have extra arguments to modify the config

.github/workflows/publish_draft_releases.yml
```diff
@@ -764,7 +1008,44 @@ jobs:
...
           runner-label: ${{ needs.start-node-builder-current.outputs.runner-label }}
           aws-instance-id: ${{ needs.start-node-builder-current.outputs.aws-instance-id }}
-  start-integration-tester:
+  start-calamari-integration-tester:
+    runs-on: ubuntu-20.04
+    outputs:
+      runner-label: ${{ steps.start-self-hosted-runner.outputs.runner-label }}
+      aws-region: ${{ steps.start-self-hosted-runner.outputs.aws-region }}
+      aws-instance-id: ${{ steps.start-self-hosted-runner.outputs.aws-instance-id }}
+    if: ${{ startsWith(github.ref, 'refs/tags') || github.ref == 'refs/heads/manta' }}
```
Garandor: we'll probably want this running on branches leading to a release merge instead of the tag. i'm thinking a filter like 
`startsWith(github.ref, 'refs/heads/release-` and then we force everyone to make release branches with the `release-v1.2.3` name

The tag case is btw already covered by the run from `heads/manta`, since we're always tagging `manta` commits

tests/rpc_performance_test.ts
```diff
@@ -0,0 +1,64 @@
...
+        vn_batch_number: 2,
+        vn_batch_size: 4096,
+    },
+    timeout: 200000,
+    expected_sync_time: 600
+}
+
+async function single_rpc_performance(api:ApiPromise) {
+    const before_rpc = performance.now();
+    const receiver_checkpoint = new Array<number>(manta_pay_config.shard_number);
```
Garandor: why is this call part of the measured rpc execution time?

ghzlatarev: The array fill? Moved it out.

tests/rpc_performance_test.ts
Garandor: Typescript tests are meant to be part of CI, not of nodes running locally on a dev PC.
So IMHO, just hardcode it. If you *need* to run the CI tests locally, just use the default ports

Garandor (dismissed):
LGTM!

Dengjianping:
Can you append `/tests/node_modules/` to `.gitignore`?
If not, it will blow out my git changes in vscode when doing these tests.

ghzlatarev:
> /tests/node_modules/

Yep

Dengjianping (dismissed):

CHANGELOG.md
```diff
@@ -18,9 +28,9 @@
 - [Dolphin] [\#622](https://github.com/Manta-Network/Manta/pull/622) Update parameter path from `sdk` to `manta-parameters`.
 - [\#636](https://github.com/Manta-Network/Manta/pull/636) Equalize Barrier impl between Calamari/Dolphin production and XCM tests mock parachain.
 - [\#638](https://github.com/Manta-Network/Manta/pull/638) `existence_requirement` argument for `FungibleLedger` trait functions.
-- [\#652](https://github.com/Manta-Network/Manta/pull/652) Reduce CI failure rate by switching AWS CI runners from AMD to Intel.
```
Garandor: don't change the already released sections pls

ghzlatarev: Oof my bad, will revert.

Dengjianping (dismissed):

Dengjianping (dismissed):

Garandor (dismissed):

.github/workflows/publish_draft_releases.yml
```diff
@@ -571,6 +572,249 @@ jobs:
...
+        run: |
+          # TODO: implement moonbeam-like js test suite triggers here
+          sleep 120
+          cd ${{ github.workspace }}/Manta/tests
+          yarn install
+          yarn
+          yarn correctness_test --address=ws://127.0.0.1:9921 --exit
+          echo $?
+          if [ $? != 0 ]; then echo "Ledger RPC correctness test failed!"; exit 1; fi
+          yarn performance_test --address=ws://127.0.0.1:9921 --exit
```
stechu, line 734: I don't see a lot of reasons to put the performance test in CI. Since the result of the performance test is a score rather than pass/not pass.

stechu:
@ghzlatarev  can you elaborate why performance test need to be in CI?

.github/workflows/publish_draft_releases.yml
ghzlatarev: I changed it to to check that it is below an expected sync time https://github.com/Manta-Network/Manta/pull/681/files#diff-19d51960f59786f9c40532e38e3a902c64be0e21e3b84a46702bcb0718a20605R20

ghzlatarev: Useful as a sanity check mainly so we're sure it hasn't regressed for whatever reason.

ghzlatarev:
> @ghzlatarev can you elaborate why performance test need to be in CI?

Sanity check for regressions. The sync time can regress for a variety of reasons, including upstream code - storage, rpc, etc. 
So it has an expected sync time now which is an average of 50 tries, https://github.com/Manta-Network/Manta/pull/681/files#diff-19d51960f59786f9c40532e38e3a902c64be0e21e3b84a46702bcb0718a20605R20
Usually it comes out to 520ms on the CI machines so the gave it a ~20% tolerance to 650

Dengjianping (dismissed):

Apokalip (dismissed):

tests/rpc_performance_test.ts
```diff
@@ -16,23 +16,30 @@ const test_config = {
         vn_batch_number: 2,
         vn_batch_size: 4096,
     },
-    timeout: 200000,
-    expected_sync_time: 600
+    timeout: 1000000,
```
Garandor: looks like you can calculate the timeout from `expected_average_sync_time ` * `sync_iterations` * (`max_deviation_percentage`+1) instead of hardcoding it

Garandor (changes requested):

tests/rpc_performance_test.ts
ghzlatarev: I can't have it in this const struct if I want to make it a variable. The current timeout should give enough time for everything to finish so the prints can be logged.

Garandor: exactly my point. It can be calculated => it shouldnt be an explicitly hardcoded constant

>  The current timeout should give enough time for everything to finish so the prints can be logged.

Until you decided to run 50 more/less sync loops for some reason and forget to adjust the timeout

Garandor (changes requested):
pls calculate timeout as per comment

tests/rpc_performance_test.ts
ghzlatarev: sure

ghzlatarev: Actually you can't calculate it like this, because there's storage setup phase.

tests/rpc_performance_test.ts
ghzlatarev: All the CI tests just pass --address argument to overwrite this.

tests/rpc_performance_test.ts
Garandor: From what I can see this is included in `expected_average_sync_time`, no?
before_rpc and after_rpc include the setup and it should be constant no matter the iterations

ghzlatarev: The expected_average_sync_time is total_sync_time / total_iterations, different than the timeout for the whole test.

Garandor: different how and/or why?

ghzlatarev: because the test has to setup the storage first

ghzlatarev: Let me include this in the description. I kind of skipped over these details because the tests were already merged from a dfifferent PR.

tests/rpc_performance_test.ts
```diff
@@ -16,23 +16,30 @@ const test_config = {
...
 }
 
 async function single_rpc_performance(api:ApiPromise) {
-    const receiver_checkpoint = new Array<number>(manta_pay_config.shard_number);
-    receiver_checkpoint.fill(0);
-    const before_rpc = performance.now();
-    const data = await (api.rpc as any).mantaPay.pull_ledger_diff(
-        {receiver_index: new Array<number>(manta_pay_config.shard_number).fill(0), sender_index: 0},
-        BigInt(16384), BigInt(16384));
-    const after_rpc = performance.now();
```
Garandor: Not knowing the test framework, i assume the setup happens in that line

https://github.com/Manta-Network/Manta/pull/681/files/c8261bf4b29f8a1ef09386b96221c15a481d6ce5..43b1de6efd1c84c391b306359db3aa197c3fb0ed#diff-19d51960f59786f9c40532e38e3a902c64be0e21e3b84a46702bcb0718a20605R29

If that's not the case and setup is a one-time invocation that's constant, can't you just add that constant to the calculated timeout?

tests/rpc_performance_test.ts
Garandor: Not knowing the test framework, i assume the setup happens in that line

https://github.com/Manta-Network/Manta/pull/681/files/c8261bf4b29f8a1ef09386b96221c15a481d6ce5..43b1de6efd1c84c391b306359db3aa197c3fb0ed#diff-19d51960f59786f9c40532e38e3a902c64be0e21e3b84a46702bcb0718a20605R29

If that's not the case and setup is a one-time invocation that's constant, can't you just add that constant to the calculated timeout?

ghzlatarev: It's not constant.

Anyway, there can be two timeouts, one that is eye-balling the storage injection plus expected total sync time.

tests/rpc_performance_test.ts
```diff
@@ -0,0 +1,71 @@
...
+            types: manta_pay_types,
+            rpc: rpc_api});
+    
+        const keyring = new Keyring({ type: 'sr25519' });
+        const sudo_key_pair = keyring.addFromMnemonic(test_config.mnemonic);
+        await setup_storage(api, sudo_key_pair, 0, test_config.storage_prepare_config);
+        await single_rpc_performance(api);
+
+        api.disconnect();
+    }).timeout(test_config.storage_setup_timeout + (test_config.sync_iterations * test_config.expected_average_sync_time * 1.5));
```
Garandor: Last request would be to make `1.5` a named constant variable, but looks good.

Also maybe a smaller buffer like 1.2 would be more appropriate, leaving 5min over when it's expected to take  e.g. 10 is fairly long

Garandor (dismissed):

Garandor (approved):

stechu (approved):

PR #694, opened 2022-07-19
Use u128::MAX in fungible ledger transfer test

### LINKED ISSUE #685: Switch to u128::MAX in fungible ledger mint integration tests
// TODO:: Switch to u128::MAX when we start using https://github.com/paritytech/substrate/pull/11241
https://github.com/Manta-Network/Manta/blob/8ab34fcb1022d4891d519bdcbb6ae22d97852ed4/runtime/calamari/tests/integration_tests.rs#L1140

ghzlatarev:

## Description

closes: #685 

* Use u128::MAX in fungible ledger transfer test for initial balance. This case used to fail before because the `can_deposit` checks would try to increase total supply and overflow with large numbers. Fixed with https://github.com/paritytech/substrate/pull/11241
* Get rid of division by 2 of the supply in MantaPay tests which was what originally caught the bug.

---

write a little note why.

Situational Notes:


pallets/manta-pay/src/test/payment.rs
```diff
@@ -160,9 +160,8 @@ where
         Some(id) => id,
         None => rng.gen(),
     };
-    // FIXME: get rid of the division after parity fixes the pallet-asset bug
-    let double_balance: u128 = rng.gen();
-    let total_free_balance = AssetValue(double_balance / 2);
+    let balance: u128 = rng.gen();
```
Dengjianping: Can both line of code be a function? I mean it's being used several places.
Not sure whether it's possible.

runtime/calamari/tests/integration_tests.rs
```diff
@@ -1108,8 +1108,7 @@ fn concrete_fungible_ledger_transfers_work() {
             ),);
 
             // Register and mint for testing.
-            // TODO:: Switch to u128::MAX when we start using https://github.com/paritytech/substrate/pull/11241
-            let amount = INITIAL_BALANCE;
+            let amount = u128::MAX;
```
Garandor: Wouldn't it be cleaner to access u128 through `manta_primitives::types::Balance::MAX`?

Garandor: Actually, this is testcode, so nvm

ghzlatarev: I can use the manta-primitives Balance, so it's more of an integration test.

pallets/manta-pay/src/test/payment.rs
ghzlatarev: Just made it a 1 liner.

Garandor (dismissed):
LGTM

Dengjianping (dismissed):
LGTM!

Dengjianping:
I'm afraid you have to separate a `Unreleased` section for new changelogs.

Garandor (approved):
LGTM

Dengjianping (approved):

PR #695, opened 2022-07-19
Refactor fungible ledger mint/burn

### LINKED ISSUE #687: Rename fungible ledger mint/burn
mint/burn are not very appropriate names because the implementations can sometimes not mint or burn, but simply deposit/withdraw.
Something like `deposit_can_mint` and `withdraw_can_burn` would be btter.
can_deposit should not be called within mint, caller should call it separately.

ghzlatarev:
## Description

closes: #687 

* Changed the names of mint/burn to signify that in some cases they can simply deposit/withdraw without increasing/decreasing the total supply of the assets.
* Added some more documentation about the assumptions of the code.
* Move `can_deposit` call out of `deposit_can_mint` and leave it for the caller of the trait to use.

---

write a little note why.

Situational Notes:


pallets/asset-manager/src/lib.rs
```diff
@@ -426,8 +426,20 @@ pub mod pallet {
                 AssetIdLocation::<T>::contains_key(&asset_id),
                 Error::<T>::UpdateNonExistAsset
             );
+
+            ensure!(
```
Dengjianping: Perhaps question mark `?` is more elegant.

primitives/manta/src/xcm.rs
```diff
@@ -365,8 +365,15 @@ impl<
 
         let (asset_id, amount, who) = Self::match_asset_and_location(asset, location)?;
 
-        MultiAdapterFungibleLedger::mint(asset_id, &who, amount)
-            .map_err(|_| XcmError::FailedToTransactAsset("Failed Mint"))?;
+        // If it's non-native asset we want to check with increase in total supply.
+        // Otherwise it will just use false, as it is assumed the native asset supply cannot be changed.
+        MultiAdapterFungibleLedger::can_deposit(asset_id, &who, amount, true).map_err(|_| {
```
Dengjianping, line 370: Just feel a little confused about line `370` and `374`. 
If line 370 receives non-native asset id, total supply will be increased, right?
But line 374 seems will increase total supply as well for non native asset.

Perhaps I don't understand related code here.

runtime/calamari/tests/integration_tests.rs
```diff
@@ -1108,9 +1108,8 @@ fn concrete_fungible_ledger_transfers_work() {
             ),);
 
             // Register and mint for testing.
-            // TODO:: Switch to u128::MAX when we start using https://github.com/paritytech/substrate/pull/11241
-            let amount = INITIAL_BALANCE;
-            assert_ok!(CalamariConcreteFungibleLedger::deposit_can_mint(
+            let amount = Balance::MAX;
+            assert_ok!(CalamariConcreteFungibleLedger::mint(
```
Garandor:
```suggestion
            assert_ok!(CalamariConcreteFungibleLedger::deposit_can_mint(
```

primitives/manta/src/xcm.rs
ghzlatarev: Line 370 will check by increase total supply only for the non-native asset. Internally you can see that the native asset has `false` hard-coded and doesn't use the argument.

Line 374 will again only increase the total supply for non-native assets. For the native asset it will use `deposit_create` internally which doesn't increase the total supply.

primitives/manta/src/assets.rs
```diff
@@ -320,6 +323,8 @@ where
     fn ensure_valid(asset_id: AssetId) -> Result<(), FungibleLedgerError>;
 
     /// Check whether `account` can increase its balance by `amount` in the given `asset_id`.
+    /// Non-native assets will use the `can_increase_total_supply` check, while native assets will not.
+    /// Should be called before `deposit_can_mint`.
```
Garandor: is there any reason for it not to be called before `deposit_can_mint`?
If not, pls wrap them to ensure they'll always be called in order and make the current fns private

Garandor (dismissed):

primitives/manta/src/assets.rs
ghzlatarev: The comment is not really good, I'll change it. What you are asking for was the initial state of the PR

ghzlatarev:
Test will keep failing until https://github.com/Manta-Network/Manta/pull/703 is merged.

pallets/asset-manager/src/lib.rs
Dengjianping: Seems you reverted `?`.

ghzlatarev: Ah yes, I thought this was the cause of the other problem and reverted. But as it turns out it's not the problem so unreverting it now :D

Garandor (approved):
LGTM!

Garandor (changes requested):
Pls add milestone and ACL labels

Garandor (approved):

Garandor:
A-` was missing 

added

Dengjianping (approved):

PR #701, opened 2022-07-21
switch runtime to wasm only

### LINKED ISSUE #469: thread 'ImportQueue' panicked at 'Storage root must match that calculated.'
Version 0.3.0-9d7a3c1d-x86_64-linux-gnu
From substrate/runtime/executive/src/lib.rs:212:3 (see attached screen grab; also note "Runtime: balance too low to send value")
polkadot_0 3 0_error

flame4:
Signed-off-by: Yi <flame0743@gmail.com>

## Description

relates to PR closes: paritytech/substrate#469.

This PR switch the runtime to wasm only, instead of NativeElseWasmExecutor.
It means in compiled binary, it'll never contain a native runtime.
You can get more relative issue on paritytech/substrate#469, but the brief benifit i think is:
- since the existing of forkless upgrade, only some specific version can be speed up with native binary because it can be only compiled once.
- eliminate the abstruction complexity to support two runtime
- lower binary size and compiling time

---

write a little note why.

Situational Notes:


flame4:
Before remove native binary size: 134MB.
After remove native binary size:  123MB. 

@ghzlatarev

Compiler env: cargo 1.61.0 (a028ae42f 2022-04-29) stable on MacPro 2022 M1 Chip

flame4:
Use polkadot-launch raising up a local network. It works well.

ghzlatarev:
> Use polkadot-launch raising up a local network. It works well.

This is also part of the `publish-draft-release` job fyi

node/src/service.rs
```diff
@@ -67,6 +66,15 @@ use sp_transaction_pool::runtime_api::TaggedTransactionQueue;
...
+type HostFunctions = sp_io::SubstrateHostFunctions;
+
+#[cfg(feature = "runtime-benchmarks")]
+type HostFunctions = (
+    sp_io::SubstrateHostFunctions,
+    frame_benchmarking::benchmarking::HostFunctions,
+);
+
 /// Native Manta Parachain executor instance.
 pub struct MantaRuntimeExecutor;
```
Dengjianping, line 79: Seems it's removable? I didn't any place where it's being used.

node/src/service.rs
```diff
@@ -109,22 +117,25 @@ impl NativeExecutionDispatch for DolphinRuntimeExecutor {
     }
 }
 
+/// We use wasm executor only now.
+pub type DefaultExecutorType = WasmExecutor<HostFunctions>;
+
 /// Full Client Implementation Type
-pub type Client<R, E> = TFullClient<Block, R, NativeElseWasmExecutor<E>>;
+pub type Client<RtApi> = TFullClient<Block, RtApi, DefaultExecutorType>;
```
Dengjianping: Can you name it as `RuntimeApi`? It should be more readable.

node/src/service.rs
flame4: Nope.  It's used in commands.rs in benchmark and try-runtime mode.

ghzlatarev (commented):
Can you add this `force-debug` feature for debugging: https://github.com/PureStake/moonbeam/blob/master/runtime/moonbase/Cargo.toml#L252
When we compile for wasm a lot of the debug impl is stripped, so we will be getting `<wasm:stripped>` in logs
Until now we've been working around this by running with native while debugging, but this is better solution
This is the log implementation if you're curios https://paritytech.github.io/substrate/master/src/sp_debug_derive/impls.rs.html#46-175

We could probably also use `disable-logging` feature for releases to shrink the size further.

Can you also write a description for the PR.

flame4:
> 

1

node/src/service.rs
flame4: done.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
 ### Improvements
 - [\#694](https://github.com/Manta-Network/Manta/pull/694) Switch to u128::MAX in fungible ledger transfer integration test.
+- [\#469](https://github.com/Manta-Network/Manta/pull/469) Switch runtime to wasm only.
```
stechu: should be 701.

node/src/service.rs
```diff
@@ -41,12 +41,11 @@ use polkadot_service::CollatorPair;
 
 use futures::lock::Mutex;
 pub use manta_primitives::types::{AccountId, Balance, Block, Hash, Header, Index as Nonce};
-use sc_client_api::ExecutorProvider;
 use sc_consensus::{
     import_queue::{BasicQueue, Verifier as VerifierT},
     BlockImportParams,
 };
-use sc_executor::{NativeElseWasmExecutor, NativeExecutionDispatch};
+use sc_executor::{NativeExecutionDispatch, WasmExecutor};
```
stechu, line 48: do we still need `NativeExecutionDispatch`?

stechu (dismissed):

runtime/calamari/Cargo.toml
```diff
@@ -220,3 +221,11 @@ std = [
...
   'orml-xtokens/std',
 ]
+
+# default any debug printing from wasm execution is stripped for generating runtime size as small as possible.
+# if use `force-debug` feature, the debug printing will be open.
+force-debug = ['sp-debug-derive/force-debug']
+# A feature that should be enabled when the runtime should be build for on-chain
+# deployment. This will disable stuff that shouldn't be part of the on-chain wasm
+# to make it smaller like logging for example.
+on-chain-release-build = [ "sp-api/disable-logging" ]
```
ghzlatarev: Did you get a reading on how much the size shrinks ?

Also please add this to the release workflow https://github.com/Manta-Network/Manta/blob/manta/.github/workflows/publish_draft_releases.yml#L134

pallets/manta-pay/src/lib.rs
```diff
@@ -894,7 +894,7 @@ where
             &proof,
         )
         .ok()?
-        .then(move || (Wrap(()), event))
+        .then_some((Wrap(()), event))
```
flame4: This line block the newest version clippy checking so just refine it.

node/src/service.rs
flame4: Yeah. This one used in benchmark and try-runtime mode in commands.rs.  I don't think i have a deep knowing about them so it should be kept.

CHANGELOG.md
flame4: Ahh...  How stupid im

ghzlatarev: Don't worry about the changelog, just revert the whole change. We'll fill it out when we release because it's causing merge  conflicts non stop right now. Not just on your PR.

flame4:
> Can you add this `force-debug` feature for debugging: https://github.com/PureStake/moonbeam/blob/master/runtime/moonbase/Cargo.toml#L252 When we compile for wasm a lot of the debug impl is stripped, so we will be getting `<wasm:stripped>` in logs Until now we've been working around this by running with native while debugging, but this is better solution This is the log implementation if you're curios https://paritytech.github.io/substrate/master/src/sp_debug_derive/impls.rs.html#46-175
> 
> We could probably also use `disable-logging` feature for releases to shrink the size further.
> 
> Can you also write a description for the PR.

for the `force-debug` feature, it will lead to a compile error.
the reason is that some code on pallet-xcm with polkadot-v0.9.22 is not compatible with `force-debug`. They have already fixed it in newer version, but we can't upgrade version so simple for other potential problems.
so in this pr the `force-debug` line is removed, but log it as new issue.

`disable-logging` make binary smaller from 123MB to 121MB.

runtime/calamari/Cargo.toml
flame4: compile size[without on-chain-release-build]
manta size: 79MB
calamari_runtime.compact.compressed.wasm: 692K
dolphin_runtime.compact.compressed.wasm: 898K
kusama_runtime.compact.compressed.wasm: 1.2M
manta_runtime.compact.compressed.wasm: 543K
polkadot_runtime.compact.compressed.wasm: 1.1M

[with on-chain-release-build]
manta size: 77MB
calamari_runtime.compact.compressed.wasm: 666K
dolphin_runtime.compact.compressed.wasm: 875K
kusama_runtime.compact.compressed.wasm: 1.2M
manta_runtime.compact.compressed.wasm: 521K
polkadot_runtime.compact.compressed.wasm: 1.1M

@ghzlatarev

ghzlatarev: Meh, better than. nothing. Might as well include it.

.github/workflows/publish_draft_releases.yml
```diff
@@ -131,7 +131,7 @@ jobs:
           SCCACHE_DIR: /home/runner/.cache/sccache
         run: |
           source ${HOME}/.cargo/env
-          RUSTC_BOOTSTRAP=1 cargo build --profile production --verbose
+          RUSTC_BOOTSTRAP=1 cargo build --profile production --verbose --features "std,on-chain-release-build"
```
ghzlatarev: no need for the `std`

.gitignore
```diff
@@ -4,3 +4,4 @@
 Cargo.lock
 runtime/**/target
 .vscode
+.idea
```
ghzlatarev, line 7: what is this for ?

runtime/calamari/Cargo.toml
ghzlatarev: https://github.com/paritytech/substrate/blob/master/primitives/api/src/lib.rs#L44-L54 fyi

.gitignore
flame4: same reason for vscode. It's configuration file on Clion-Rust.

.github/workflows/publish_draft_releases.yml
flame4: done.

ghzlatarev (dismissed):

ghzlatarev (dismissed):

Garandor:
> relates to PR closes

The OR was not a typo, you delete the part that doesn't apply.
If the issue has multiple PRs or other steps, it's `relates to`, if there's nothing else to do, it's `closes`

FYI @ghzlatarev @Apokalip

Garandor:
A conceptual question here - please DO NOT MERGE until after discussion:

**Is it worth making non-consensus nodes ( full nodes handling RPC etc ) spend >2x more computation time executing the block import WASM over running native just to save 2 MB on the binary size?**

Have you measured the CPU busy % of a full node keeping up with the chain before and after?
While it has no impact on benchmarks, it *may* have impact on other node services like RPC servers.
It's difficult to measure because it means we have to simulate full-ish blocks to generate enough load for this to matter, but I bet that RPC performance will go down ( either in round trip timing or capacity of parallel connections handled per second ) after this change.
How much is an open question.

Please convince me that this is a good idea with other reasons than `statemine did it`

~~Code complexity is obviously not it, since this PR saves 14 lines~~
I get the code complexity point, but that can't be all

ghzlatarev:
> A conceptual question here - please DO NOT MERGE until after discussion:
> 
> **Is it worth making non-consensus nodes ( full nodes handling RPC etc ) spend >2x more computation time executing the block import WASM over running native just to save 2 MB on the binary size?**
> 
> Have you measured the CPU busy % of a full node keeping up with the chain before and after? While it has no impact on benchmarks, it _may_ have impact on other node services like RPC servers. It's difficult to measure because it means we have to simulate full-ish blocks to generate enough load for this to matter, but I bet that RPC performance will go down ( either in round trip timing or capacity of parallel connections handled per second ) after this change. How much is an open question.
> 
> Please convince me that this is a good idea with other reasons than `statemine did it`
> 
> ~Code complexity is obviously not it, since this PR saves 14 lines~ I get the code complexity point, but that can't be all

RPC is part of the client, this change is referring to the runtime. Also recall that in all of our production nodes the client is already only communicating with a wasm-executed runtime. So there should not be any performance loss/gain from this change.

Following the link chain you can find the broader discussion https://github.com/paritytech/polkadot-sdk/issues/62. The main problem with native runtime (imo) is that there's a chance that the network can run runtimes compiled with different compiler versions, which was partly the reason for a major stall on [Polkadot already. ](https://polkadot.network/blog/a-polkadot-postmortem-24-05-2021/)

@flame4 I'd put some of these more important discussion issues and linked PRs in the description so they can be in plain view of reviewers.

Garandor:
> RPC is part of the client, this change is referring to the runtime. Also recall that in all of our production nodes the client is already only communicating with a wasm-executed runtime. So there should not be any performance loss/gain from this change.
> Following the link chain you can find the broader discussion [paritytech/polkadot-sdk#62](https://github.com/paritytech/polkadot-sdk/issues/62). The main problem with native runtime (imo) is that there's a chance that the network can run runtimes compiled with different compiler versions, which was partly the reason for a major stall on [Polkadot already. ](https://polkadot.network/blog/a-polkadot-postmortem-24-05-2021/)

Runtime *Executor* is part of the client as well. Check the diffs on this PR.
I remember Rob is running our nodes with the wasm execution CLI flag, yes.
Doesn't mean he must continue to be doing so - at least not for the non-collating nodes.
I refreshed my memory by re-reading this, but it mainly talks about the stalling problems you referenced.
That's about *block production* i.e. only relevant for *collator* nodes.
In the freak case that a compiler problem occurs, losing our full nodes and have their RPC connectivity freeze on some block while the chain continues to produce blocks is not nearly as bad as stalling the chain.

The argument here is, what's wrong with the CLI flag?

Second argument is - if this performance difference is negligible, this whole discussion doesn't matter.
Would be nice to see some performance numbers while the chain is processing large amounts of transactions before/after this change @flame4

ghzlatarev:
> > RPC is part of the client, this change is referring to the runtime. Also recall that in all of our production nodes the client is already only communicating with a wasm-executed runtime. So there should not be any performance loss/gain from this change.
> > Following the link chain you can find the broader discussion [paritytech/polkadot-sdk#62](https://github.com/paritytech/polkadot-sdk/issues/62). The main problem with native runtime (imo) is that there's a chance that the network can run runtimes compiled with different compiler versions, which was partly the reason for a major stall on [Polkadot already. ](https://polkadot.network/blog/a-polkadot-postmortem-24-05-2021/)
> 
> Runtime _Executor_ is part of the client as well. Check the diffs on this PR. I remember Rob is running our nodes with the wasm execution CLI flag, yes. Doesn't mean he must continue to be doing so - at least not for the non-collating nodes. I refreshed my memory by re-reading this, but it mainly talks about the stalling problems you referenced. That's about _block production_ i.e. only relevant for _collator_ nodes. In the freak case that a compiler problem occurs, losing our full nodes and have their RPC connectivity freeze on some block while the chain continues to produce blocks is not nearly as bad as stalling the chain.
> 
> The argument here is, what's wrong with the CLI flag?
> 
> Second argument is - if this performance difference is negligible, this whole discussion doesn't matter. Would be nice to see some performance numbers before/after this change @flame4
> 
> > Following the link chain you can find the broader discussion [paritytech/polkadot-sdk#62](https://github.com/paritytech/polkadot-sdk/issues/62). The main problem with native runtime (imo) is that there's a chance that the network can run runtimes compiled with different compiler versions, which was partly the reason for a major stall on [Polkadot already. ](https://polkadot.network/blog/a-polkadot-postmortem-24-05-2021/)
> > @flame4 I'd put some of these more important discussion issues and linked PRs in the description so they can be in plain view of reviewers.

Not sure i'm following. But 1. we will continue running our runtime in WASM and 2.The client will always be native.
Again this whole change is related to how the runtime runtime is executed.

Garandor:
> 1. we will continue running our runtime in WASM

No argument there for *collator* nodes. My point is that for *full* nodes, forcing WASM over native runtime is wasteful.

ghzlatarev:
> > 1. we will continue running our runtime in WASM
> 
> No argument there for _collator_ nodes. My point is that for _full_ nodes, forcing WASM over native runtime is wasteful.

I guess I'm still not following, but can you ask your question in the upstream discussion. Perhaps they can answer.

Garandor:
I'll do that. In the meantime - CC @stechu for a second opinion

ghzlatarev:
> I'll do that. In the meantime - CC @stechu for a second opinion

Actually I think I get what you mean, but we need to measure the system as a whole to see whether RPC nodes are the bottleneck. Definitely out of scope of this PR. And in the end of the day you can always spawn more RPC nodes if that's the bottleneck.
As mentioned earlier this change will have no performance impact on all of our wasm-executed nodes.
EDIT: correction our nodes are not using running with the wasm flag yet..

ghzlatarev:
> > I'll do that. In the meantime - CC @stechu for a second opinion
> 
> Actually I think I get what you mean, but we need to measure the system as a whole to see whether RPC nodes are the bottleneck. Definitely out of scope of this PR. And in the end of the day you can always spawn more RPC nodes if that's the bottleneck. As mentioned earlier this change will have no performance impact on all of our wasm-executed nodes.

Probably it would be best to compile a list of all the benefits in 1 place.

Garandor:
My concern was answered here https://github.com/paritytech/polkadot-sdk/issues/62

We're good to merge this :)

Garandor (changes requested):
Actually, we're not:

- Please add A- C- and L- labels :)

ghzlatarev (approved):

stechu:
@Garandor @ghzlatarev  @flame4  for the RPC performance discussion, I think it really depends on kind of RPC we are doing. To make it simple, let's imagine two extreme scenarios:
1. RPC `A`, almost all the execution happens in the runtime
2. RPC `B`, almost all the execution happens in the non-wasm client part

Clearly, switching to wasm only execution only affect `A` a lot, and has very minimal impact on `B`. A general design is we should never design something like `A`. That's why @Dengjianping is working on indexers. We may revisit this after adding PVM tho.

flame4:
> Actually, we're not:
> 
> * Please add A- C- and L- labels :)

Already Add. Please review it.

Garandor (approved):

Issue #686, opened 2022-07-16
Indexer Ledger(Tracking Issue)


Dengjianping:
- [ ] indexer(Jamie)
  - [ ] db design and IO benchmarking
  - [ ] Concurrency issues
  - [ ] `Shards` reconstruction
  - [ ] Benchmarking and performance
  - [ ] Threading model
- [ ] rpc forwarding(Yi)
  - [ ] Figure out must-have rpc methods to be forwarded
  - [ ] How to forward subscription rpc
- [ ] monitoring
  - [ ] what metrics need to be monitored
- [ ] How to globally configure the indexer server correctly
- [ ] ...

Garandor:
tracked in https://github.com/Manta-Network/manta-indexer

PR #712, opened 2022-07-25
Add RPC for latest checkpoint

bhgomes:

## Description

Adds an RPC method for querying the latest checkpoint. This is useful for getting a loading bar for synchronization.

---

write a little note why.

Situational Notes:


Kevingislason (dismissed):

stechu:
why a new RPC end-point? Should we just add the total number to the `pull_ledger_diff` (to reduce number of RPCs).

stechu:
This PR requires merging #723  first.

stechu (dismissed):

Kevingislason (dismissed):

pallets/manta-pay/src/lib.rs
```diff
@@ -557,10 +557,14 @@ pub mod pallet {
             let (more_receivers, receivers) =
                 Self::pull_receivers(*checkpoint.receiver_index, max_receivers);
             let (more_senders, senders) = Self::pull_senders(checkpoint.sender_index, max_senders);
+            let receivers_total = (0..=255)
+                .map(|i| ShardTrees::<T>::get(i).current_path.leaf_index as u128)
```
stechu, line 568: @bhgomes  do we need +1 here?

stechu: should not +1 here.

stechu (approved):

pallets/manta-pay/src/lib.rs
```diff
@@ -557,10 +557,15 @@ pub mod pallet {
             let (more_receivers, receivers) =
                 Self::pull_receivers(*checkpoint.receiver_index, max_receivers);
             let (more_senders, senders) = Self::pull_senders(checkpoint.sender_index, max_senders);
+            let senders_receivers_total = (0..=255)
```
ghzlatarev, line 567: I still don't quite get why this is better, than just making a separate storage item to keep track of the utxo set size.

Garandor (approved):
got the same thought as @ghzlatarev  in https://github.com/Manta-Network/Manta/pull/712/files#r935325448 that it would be cheaper to bump the total number in storage on each TX and read a StorageValue instead of collecting the 256 shards on *every single rpc call*.

Intuitively, it sounds far more computationally expensive to have 257 storage reads per RPC instead of 2, especially in light of the recently merged #469 

I think this'll need proper benchmarking - or some caching in the #686 where it only pulls a new value from the RT once per block and serves the rest from cache.

If we do the latter thing this LGTM. ~~@bhgomes could you pls open a followup issue?~~ Nvm, just realized that the #686 indexer IS the followup issue

ghzlatarev (approved):

Dengjianping (approved):

0xduracell:
> got the same thought as @ghzlatarev in https://github.com/Manta-Network/Manta/pull/712/files#r935325448 that it would be cheaper to bump the total number in storage on each TX and read a StorageValue instead of collecting the 256 shards on _every single rpc call_.
> 
> Intuitively, it sounds far more computationally expensive to have 257 storage reads per RPC instead of 2, especially in light of the recently merged #469
> 
> I think this'll need proper benchmarking - or some caching in the #686 where it only pulls a new value from the RT once per block and serves the rest from cache.
> 
> If we do the latter thing this LGTM. ~@bhgomes could you pls open a followup issue?~ Nvm, just realized that the #686 indexer IS the followup issue

Yeah, we have discussed this. Will do it in the `MantaPay 1.0.0`.

PR #715, opened 2022-07-26
Update xcm-onboarding and release templates

### LINKED ISSUE #600: Tracking: Improve documentation

ghzlatarev:
## Description

closes: #600 

* Update release update to mention Calamari @ Moonbase-Relay as well as Dolphin @ Baikal internal testnets. Dolphin @ Rococo should not be part of the testing process, because this is a user-facing (semi-production) environment that we don't want to screw up accidentally without having it tested.
* Update the xcm onboarding template to include the process for the production chains to make it clearer. Also add an instruction for opening governance proposals on Calamari.

---

write a little note why.

Situational Notes:


.github/ISSUE_TEMPLATE/release.md
```diff
@@ -60,6 +60,7 @@ Note: Usually update client first then runtime.
 - [ ] Coordinate with marketing team for documentation updates and other relevant tasks.
 - [ ] Update changelog.
 - [ ] If the release contains any changes that break/change functionality used in https://github.com/Manta-Network/sdk (e.g. RPC changes, see also [extrinsic ordering](#extrinsic-ordering)), raise a PR there and **block this release** until your PR has been merged and incorporated in a new SDK release.
+- [ ] During the [burn-in](#burn-in) make sure to look out for anomalies in our nodes' memory, cpu, disk and network usage via the [Grafana Node Explorer](https://grafana.pulse.pelagos.systems/d/rYdddlPWk/node-exporter-full).
```
Garandor:
```suggestion
- [ ] Before declaring a successful [burn-in](#burn-in) make sure to look out for anomalies in our nodes' memory, cpu, disk and network usage via the [Grafana Node Explorer](https://grafana.pulse.pelagos.systems/d/rYdddlPWk/node-exporter-full).
```

.github/ISSUE_TEMPLATE/release.md
```diff
@@ -60,7 +71,7 @@ Note: Do not publish draft releases from PR branches, because those branches wil
 
 ### Burn In
 
-Ensure that Manta DevOps has run the new release on Como and Baikal nodes
+Ensure that Manta DevOps has run the new release on Baikal nodes
```
Garandor, line 73: 12 hours won't be enough. we really shouldn't clear a release that hasn't been running without anomalies for at least 2 days, this RAM leak only became apparent on baikal after 2 days

.github/ISSUE_TEMPLATE/release.md
```diff
@@ -31,15 +31,25 @@ The following checks can be performed after we have frozen our release candidate
...
 ### Client Releases
 
 - [ ] Verify that each crate's `version` has been bumped from previous release.
+- [ ] Update client of Baikal relay nodes.
+- [ ] Update client of Calamari-Testnet @ Baikal nodes.
+- [ ] Update client of Calamari-Testnet @ Moonbase-Relay nodes.
+- [ ] Update client of Dolphin @ Baikal nodes.
 - [ ] Check that the new client versions have [burned-in](#burn-in) without issue for at least 12 hours.
 
+
```
Garandor, line 49:
```suggestion

```

.github/ISSUE_TEMPLATE/release.md
```diff
@@ -31,15 +31,25 @@ The following checks can be performed after we have frozen our release candidate
...
-- [ ] Execute runtime upgrade to Baikal and verify network stability.
-- [ ] Execute runtime upgrade to Rococo and verify network stability.
+- [ ] Execute runtime upgrade to Baikal relay and verify network stability.
+- [ ] Execute runtime upgrade to Calamari @ Baikal and verify network stability.
+- [ ] Execute runtime upgrade to Calamari @ Moonbase-Relay and verify network stability.
+- [ ] Execute runtime upgrade to Dolphin @ Baikal and verify network stability.
+- [ ] Check that the new runtime versions have [burned-in](#burn-in) without issue for at least 12 hours.
 - [ ] Prepare a governance post and submit to our forum with description and motivation for changes.
 
+Note: Usually update client first then runtime.
```
Garandor, line 40: I think we should do this from what a user will do.
Many ppl will upgrade after the RT is already live ( except overzealous collator runners )
We should test for that case and

`Update runtime first, then client`

.github/ISSUE_TEMPLATE/release.md
Garandor: also this is duplicated with line 38 https://github.com/Manta-Network/Manta/pull/715/files#diff-4482ca0941ca8f0e3db46d20770ecf9158cc00322eee6e760460c47a4a5932c2R38

so maybe remove this

.github/ISSUE_TEMPLATE/release.md
ghzlatarev: I'm not sure if most of our node runners wait for the RT before updating their client. We should double check that, cuz we've been kind of aiming for the opposite. At least internally.

.github/ISSUE_TEMPLATE/release.md
ghzlatarev: Yeah let's make it 3 days, not sure where this 12 hours was coming from.

.github/ISSUE_TEMPLATE/release.md
Garandor: That's why i'm calling them overzealous 😋 

The normal polkadot usecase is that  you rarely update the client and simply use the updated runtimes that get passed to you from the chain.
There's rarely a reason to even upgrade your client at all, so i'd say no matter what our collator runners do currently ( maybe we should tell them to upgrade only after the chain vote) we should do runtime first

ghzlatarev: I think the other reason for that ordering was that historically there have been a lot more breaking client changes

Garandor: right, everytime we fiddle around with RPC or consensus we need to force a client upgrade.
but even in these cases there's no reason to prefer one sequence over the other
- client first and
- runtime first
must both work.

It's just that in these cases a client update is mandatory, it doesn't matter tho in which order.
In the RPC case, if you upgrade the client and ppl start using the new RPC call, the RPCs will just fail because the runtime isnt there to handle the request, and if you upgrade the runtime first, nobody will be able to use it until they update client.

Only when consensus is affected ( like with nimbus ), upgrading in sequence is necessary because upgrading client too early will make the node lose sync

ghzlatarev: Again the problem we've had is that there have been more breaking client changes than breaking runtime changes. This affects the node runners and our CI, and it is why we had set it up this way, to have the least amount of headaches. Our runtime upgrade tests assume that the client is on latest version before the runtime upgrade is executed. If we do it the opposite way we'd just have to do more exceptions and patches to our process and CI again.
So let's keep it this way until it practically starts giving us more problems than it solves.

Garandor (approved):

stechu (approved):

Garandor:
@ghzlatarev pls don't close tracking issues, at least for those, use `relates to`.

Unless we can agree that `closes #xyz` against an already closed PR is valid, then we can just leave all tracking issues closed and link PRs against the closed ones using `closes`.

this close -> reopen -> close cycle is something i want to avoid

Issue #320, opened 2022-01-04
Integrate github actions reusable workflows into CI/CD


ghzlatarev:
Research github actions reusable workflows - https://github.blog/2021-11-29-github-actions-reusable-workflows-is-generally-available/
There's a lot of duplications in our github actions currently that can probably be addressed with these features.

Currently, many workflows individually run `build_node_current`, needlessly duplicating a 50min task.
We can save a bunch of AWS compute by running this once and reusing its build output for all other tasks

PR #743, opened 2022-08-08
Split all testing from `publish_draft_release` workflow and filter execution by labels on the PRs

### LINKED ISSUE #608: Investigate splitting up the publish_draft_releases workflow
[publish_draft_releases.yml](https://github.com/Manta-Network/Manta/blob/manta/.github/workflows/publish_draft_releases.yml) currently contains

- integration tests
- docker test
on normal PRs

and the draft release generation on release candidates.

I open the file, see 1152 lines and immediately feel regret for doing so.

To reduce complexity and in the spirit of single-responsibility we should refactor this into multiple files, so that `publish_draft_releases` does exactly what the name implies and nothing more

look into https://docs.github.com/en/actions/using-workflows/reusing-workflows for modularzation

Garandor:

## Description
Refactors the `publish_draft_release` workflow so it does only what it's meant to do - publish a draft release.

Changes in functionality:
  1. Only run dolphin integration test if the PR touches the dolphin RT ( `A-dolphin` on PR )
  2. Only run calamari with its long integration test if the PR touches calamari ( `A-calamari` on PR )
  3. Only run docker_image_test with calamari_integration_test
  4. only run runtime_upgrade_test on release branches and merge to manta
  5. we run everything and the kitchen sink when it's merged to `manta`

All testing happens in a runtime-specific file and is triggered only when the PR touches the corresponding runtime as defined by the PR label.

I think this saves
- developer sanity maintaining the thing
- aws compute usage
- calamari and dolphin test workflows can be easily diffed against each other

**Please discuss and contribute!**

Also removes the `manta` runtime docker build from releases as these images are not currently used anywhere

closes #608

---

write a little note why.

Situational Notes:


Garandor:
It looks like having `if:` on each job is spammy in the list, anyone know if `if:` is allowed above the `job:` level in gh actions?

Garandor:
There are likely duplicated tests in this PR. I'm opening this as a first attempt for discussion. please feel free to pick this up and push more commits

Most likely the `runtime_upgrade_test` can be refactored out of these as well

Garandor:
One more optimization could be to use `https://github.com/rust-lang/docker-rust-nightly`

as in `runs_on: rustlang/rust:nightly` instead of running rustup at our compute cost on every run

ghzlatarev (dismissed):
lgtm. the only reason it was like this was to not dupe the code, but we've ended up doing that in a lot of other places and we can take care of that problem separately. Probably less painful than maintaining this monstrosity.

Apokalip (dismissed):
lgmt! I don't know what part of this will be moved to the type script tests, but as seeing how the signer is now on my plate too, that will take some time.

.github/workflows/integration_test_calamari.yml
```diff
@@ -0,0 +1,826 @@
...
+  AWS_REGION: us-east-1
+  AWS_SUBNET_ID: subnet-08c26caf0a52b7c19
+  AWS_SECURITY_GROUP_ID: sg-0315bffea9042ac9b
+  AWS_INSTANCE_TYPE: c5.9xlarge
+  AWS_INSTANCE_ROOT_VOLUME_SIZE: 32
+  AWS_IMAGE_SEARCH_PATTERN: ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-*
+  AWS_IMAGE_SEARCH_OWNERS: '["099720109477"]'  # canonical
+jobs:
+  get-rust-versions:
+    if: contains(github.event.pull_request.labels.*.name, 'A-calamari')
```
Garandor: actually, do not merge yet, i think the if needs an OR section that makes sure all tests run on push to manta, aka. when the PR is merged

ghzlatarev:
Not sure if this does something for the build times, it's always building the same binary but not it needs to build it twice sometimes.
Overall 50% - 50% on the core idea of doing less CI and addon with labels, instead of doing all CI and limiting based on some conditions. I guess it's more compute efficient to do the former, but then we also have more human overhead of making sure labeling is correct. And i'm really not sure this is a bottleneck.
I think optimally you want a workflow that just builds the binary and then the split (for easier maintenance) test-workflows can wait for the first workflow and download the binary and then do their thing.

Garandor:
> Not sure if this does something for the build times, it's always building the same binary but not it needs to build it twice sometimes. 

That's why i'm saying the last big optimization is to make a reusable workflow ( #320 ) and that runs once and all others use the common build output from that.
I think it's worth the extra compute for now given that it 
- saves time if the integration tests and RT upgrades are not run on every commit
- is significantly more maintainable

Garandor:
Plus I for one can't reason about the dependencies of the CI jobs in a >1k spaghetti code workflow.
Did you know that `calamari-integration-test` depends on `buiild-runtimes` when it doesn't actually use it?

There seems to be a problem with aws scheduling tho
https://github.com/Manta-Network/Manta/runs/8036038006?check_suite_focus=true#step:2:29
```
Error: aws scheduled instance starting error
Error: VcpuLimitExceeded: You have requested more vCPU capacity than your current vCPU limit of 2064 allows for the instance bucket that the specified instance type belongs to. Please visit http://aws.amazon.com/contact-us/ec2-request to request an adjustment to this limit.
```

ghzlatarev:
> > Not sure if this does something for the build times, it's always building the same binary but not it needs to build it twice sometimes.
> 
> That's why i'm saying the last big optimization is to make a reusable workflow and that runs once and all others use the common build output from that. I think it's worth the extra compute for now given that it
> 
> * saves time if the integration tests and RT upgrades are not run on every commit
> * is significantly more maintainable

I agree it would make things easier to maintain, just saying it won't do much about build times, which you included as a reason in the description.
My only concern here is the extra human effort of doing the labeling correctly, so the right workflows are triggered, and then the reviewers reviewing it correctly. So I'd keep the split of the workflows but no the labeling.

Garandor:
> I agree it would make things easier to maintain, just saying it won't do much about build times, which you included as a reason in the description.

Looking at the build times, I agree. The `check_tests` workflow that always runs takes ~90mins which is the same or more as all the integration tests which run in parallel, so skipping those won't speed up the build unfortunately :(
It will however save ~1 hour of pointlessly wasted cloud compute per skipped integration test

> My only concern here is the extra human effort of doing the labeling correctly, so the right workflows are triggered, and then the reviewers reviewing it correctly. So I'd keep the split of the workflows but no the labeling.

We're labelling these anyway because our changelog depends on them being set correctly and
- laziness is not an excuse ( we need to add it to the PR template tho )
- the workflows will run on merge at the latest ( whether labelled or not )

Dengjianping:
I saw we're building docker image for manta(actually it's the same image as calamari's). It doesn't make much sense to me.
https://hub.docker.com/r/mantanetwork/manta/tags

We should not push manta image to dockerhub.
We can enable it when manta wins the auction.

Garandor:
> I saw we're building docker image for manta(actually it's the same image as calamari's). It doesn't make much sense to me. https://hub.docker.com/r/mantanetwork/manta/tags
> 
> We should not push manta image to dockerhub. We can enable it when manta wins the auction.

Thx for the suggestion, i added it

in the future, pls add new stuff that isn't related with the current PR as a new issue, I'd rather not have a PR that's meant to be about one thing become larger in scope ( and delay being merged )

ghzlatarev:
> > I agree it would make things easier to maintain, just saying it won't do much about build times, which you included as a reason in the description.
> 
> Looking at the build times, I agree. The `check_tests` workflow that always runs takes ~90mins which is the same or more as all the integration tests which run in parallel, so skipping those won't speed up the build unfortunately :( It will however save ~1 hour of pointlessly wasted cloud compute per skipped integration test
> 
> > My only concern here is the extra human effort of doing the labeling correctly, so the right workflows are triggered, and then the reviewers reviewing it correctly. So I'd keep the split of the workflows but no the labeling.
> 
> We're labelling these anyway because our changelog depends on them being set correctly and
> 
> * laziness is not an excuse ( we need to add it to the PR template tho )
> * the workflows will run on merge at the latest ( whether labelled or not )

I'm 50/50 so cool to try it, but as you mention let's add a line to the PR template as a reminder that labeling needs to be done correctly, otherwise CI won't run. As well as a reminder that just because the code diff doesn't include some runtime, it doesn't mean that CI shouldn't be run for it (eg: changing the CI integration test workflows themselves)

ghzlatarev (approved):

Dengjianping (approved):

PR #761, opened 2022-08-17
Support all xTokens extrinsics and the canonical relative-view self-location where necessary

### LINKED ISSUE #452: Investigate (0, Here) for `SelfReserve`

ghzlatarev:
## Description

closes: #452 

* Unfilter all xtokens extrinsic to improve UX and flexibility when sending cross chain token transfers. Doesn't impact security.
* Keep `AbsoulteReserveProvider` as a reserve provider of xtokens. This allows users to select our native asset via relative view when sending to other chains with the new extrinsics. Implemented via `AbsoluteAndRelativeReserve` which always converts to the absolute view. Got rid of some related custom code in primitives.
* Rename `IsNativeConcrete` to `IsNativeConcreteAbsoluteAndRelative` to make it more explicit that this implementation supports both relative and absolute view.
* Add some more tests to show that users can select (0, Here) when sending the native token to another chain. And that we can register the the native asset as both relative and absolute location and it will work.
* All in all, there's really no need to explicitly change all runtime and asset configurations to use relative self-reserve location instead of the absolute view, because in every relevant case we support both views anyway. 
* The only place that doesn't support relative view is the `IsReserve` filter of xcm-executor, which is a filter used in `ReserveAssetDeposited` instruction, so supposedely receiving some reserve coin of another consensus system. So this can only be a problem if someone manages to send a message from our chain to our chain, which is disallowed by xtokens and pallet-xcm.

---

write a little note why.

Situational Notes:


runtime/calamari/src/xcm_config.rs
```diff
@@ -208,16 +210,19 @@ impl Config for XcmExecutorConfig {
     // Defines how to Withdraw and Deposit instruction work
     type AssetTransactor = MultiAssetTransactor;
     type OriginConverter = XcmOriginToCallOrigin;
-    // Combinations of (Location, Asset) pairs which we trust as reserves.
-    type IsReserve = MultiNativeAsset;
+    // Filter to the reserve withdraw operations.
+    // This only applies to withdraw operations of foreign assets because
+    // pallet-xcm calls are filtered and xtokens does not allow sending xcm
+    // to the local chain, so we only need to support absolute views.
+    type IsReserve = MultiNativeAsset<AbsoluteReserveProvider>;
```
ghzlatarev, line 217: Two points here:

1. This configuration hasn't really changed. The old `MultiNativeAsset` struct was just implementing the exact same code that `MultiNativeAsset<AbsoluteReserveProvider>` is using, but in `primitives/xcm.rs` . So it's a personal preference to use the upstream code, that I can revert.

2. As stated in the description the relative view is not needed for this filter, because it is only used in `ReserveAssetDeposited`, the instruction that is supposed to signify that some reserve chain has reserved some foreign assets. This can only become an issue if someone manages to send a xcm message to the origin chain, which is prevented in a bunch of ways like not having hrmp channels with yourself.

runtime/common/tests/xcm_mock/parachain.rs
```diff
@@ -694,6 +703,11 @@ pub(crate) fn create_asset_metadata(
 }
 
 pub(crate) fn create_asset_location(parents: u8, para_id: u32) -> AssetLocation {
+    if parents == 0 {
+        return AssetLocation(VersionedMultiLocation::V1(MultiLocation::new(
+            parents, Here,
+        )));
+    }
     AssetLocation(VersionedMultiLocation::V1(MultiLocation::new(
         parents,
```
Dengjianping: Maybe it's a stupid question.

Does the caller check `parents`?
Basically, it should be `0` or `1`.

ghzlatarev: Yeah should be 0 or 1 for now, added a panic otherwise.

Dengjianping (dismissed):

Apokalip (dismissed):
lgmt

primitives/manta/src/xcm.rs
```diff
@@ -29,52 +29,37 @@ use frame_support::{
...
-        // We only care about concrete location now.
-        if let xcmAssetId::Concrete(location) = self.id.clone() {
-            let first_interior = location.first_interior();
-            let parents = location.parent_count();
-            match (parents, first_interior) {
-                (0, Some(Parachain(id))) => Some(MultiLocation::new(0, X1(Parachain(*id)))),
-                (1, Some(Parachain(id))) => Some(MultiLocation::new(1, X1(Parachain(*id)))),
-                (1, _) => Some(MultiLocation::parent()),
-                _ => None,
+/// This struct offers uses RelativeReserveProvider to output absolute views of multilocations.
```
Garandor:
```suggestion
/// This struct uses RelativeReserveProvider to output absolute views of multilocations.
```

Garandor (commented):
I'm not well-versed enough to know whether `transfer_multicurrencies` or allowing relative Multilocation as input really doesn't open any new attack vectors, so I'll defer to the other two reviewers.

should we vendor ORML (similar to nimbus) seeing how all of our XCM code depends on it?

primitives/manta/src/xcm.rs
ghzlatarev: > I'm not well-versed enough to know whether `transfer_multicurrencies` or allowing relative Multilocation as input really doesn't open any new attack vectors, so I'll defer to the other two reviewers.
> 
> should we vendor ORML (similar to nimbus) seeing how all of our XCM code depends on it?

Actually it's already vendored. https://github.com/Manta-Network/Manta/blob/manta/runtime/calamari/Cargo.toml#L87

Dengjianping (approved):

Garandor:
> I'm not well-versed enough to know whether `transfer_multicurrencies` or allowing relative Multilocation as input really doesn't open any new attack vectors, so I'll defer to the other two reviewers.

@ghzlatarev if you can drop me a sentence why these two are safe i'll approve it

ghzlatarev:
> > I'm not well-versed enough to know whether `transfer_multicurrencies` or allowing relative Multilocation as input really doesn't open any new attack vectors, so I'll defer to the other two reviewers.
> 
> @ghzlatarev if you can drop me a sentence why these two are safe i'll approve it

Oops didn't see this earlier. So:
`transferWithFee ` is just like `transfer` with a separate parameter to specify a fee amount, so the fee doesn't need to be subtracted from the transfer asset.
`transferMultiAsset` is just like `transfer` but where you can input the asset multilocation directly instead of with the asset-id key in the AssetId-Location map.
`transferMultiAssetWithFee` gives you another parameter to specify fee asset and amount.
`transferMultiassets` just lets you send an array of assets instead of 1.

So all of these are just how you specify the asset to send and pay fee with. You can try sending assets you don't have or that don't exist but that will fail. There are also a bunch internal xToken invariants like fee assets being the same as transfer assets, all the assets in an array being the same, etc.

Issue #740, opened 2022-08-06
Removal of Aura pallets, keys, storage


Garandor:
- [x] pallet_aura_ext
  - [ ] Authorities storage item

PR #763, opened 2022-08-17
Support verification of historic Aura blocks

Garandor:
## Description

relates to #669 
replaces #762 

Implements a custom verifier and block import queue that wraps the ones provided by Aura and Nimbus.
Depending on if the block contains an aura or nimbus seal and the runtime has AuraApi or NimbusApi, calls are dispatched to their respective consensus implementations.

The client uses Nimbus for block *proposing* only, so it will not be able to add new blocks to the Aura chain

Relevant reading: https://substrate.stackexchange.com/questions/1319/how-to-properly-split-logic-between-block-verifier-and-block-import

### The block import sequence:
diagram-export-8_19_2022, 12_21_20 PM

---

write a little note why.

Situational Notes:


node/src/aura_or_nimbus_consensus.rs
```diff
@@ -0,0 +1,194 @@
...
+                .map_err(Into::into)
+                .await
+        } else {
+            // if block_params.header.number().is_zero(){
+            //     // NOTE: We assume the genesis block comes unsealed
+            //     info!(target: LOG_TARGET, "Skipping Seal Verification for Genesis Block");
+            //     Ok((block_params,None))
+            // } else {
+                Err("NoSealFound".to_string())
+            // }
```
ghzlatarev: forgotten commented code

runtime/dolphin/src/lib.rs
```diff
@@ -1048,6 +1048,20 @@ impl cumulus_pallet_parachain_system::CheckInherents<Block> for CheckInherents {
...
+// }
+
+// impl frame_support::traits::ExecuteBlock<Block> for AllExecutors {
+// 	fn execute_block(block: Block) {
+//         // if <v330
+//         aura.execute_block(block)
+//         // else >=v330
+// 		nimbus.execute_block(block);
+// 	}
+// }
```
ghzlatarev: forgotten commented code

node/src/aura_or_nimbus_consensus.rs
```diff
@@ -0,0 +1,194 @@
...
+    traits::{Block as BlockT, Header as HeaderT},
+};
+use std::{marker::PhantomData, sync::Arc};
+
+use nimbus_primitives::CompatibleDigestItem as NimbusDigestItem;
+use sc_consensus_aura::CompatibleDigestItem as AuraDigestItem;
+
+const LOG_TARGET: &str = "aura-nimbus-consensus";
+
+struct AuraOrNimbusVerifier<C, Block: BlockT, AuraCIDP, NimbusCIDP> {
```
ghzlatarev: nit: now all the generics are full names, might as well change the C as well

node/src/aura_or_nimbus_consensus.rs
```diff
@@ -0,0 +1,194 @@
...
+        AuraCIDP: CreateInherentDataProviders<Block, ()> + 'static,
+        NimbusCIDP: CreateInherentDataProviders<Block, ()> + 'static,
+    {
+        Self {
+            aura_verifier: sc_consensus_aura::build_verifier(BuildVerifierParams {
+                client: client.clone(),
+                create_inherent_data_providers: create_inherent_data_providers_aura,
+                // NOTE: We only support verification of historic aura blocks, not new block proposals using aura
+                can_author_with: NeverCanAuthor {},
+                check_for_equivocation: sc_consensus_aura::CheckForEquivocation::No,
```
ghzlatarev: Why are we not checking for equivocation btw, and it's actually not clear who is checking who here

Garandor: Because Aura has no code actually *handling* equivocation beside a log print here: substrate/client/consensus/aura/src/import_queue.rs

ghzlatarev: well, isn't that good enough, to at least spot suss behaviour, especially with the new log aggregation.

Garandor: nobody will be producing blocks on Aura after v330 anymore, so if anything this will be of only historic relevance.

I can add flip it to yes tho, shouldn't hurt

ghzlatarev: oof yeah, nvm. I forgot that we're not using aura anymore ... :D

ghzlatarev (dismissed):

ghzlatarev (approved):

Apokalip (approved):
LGMT, from what I understood we will need to support this forever because of the chain history being Aura blocks?

Garandor:
> LGMT, from what I understood we will need to support this forever because of the chain history being Aura blocks?

In the client - yes, in the on-chain runtime - no.

We could have already removed the Aura pallet from the runtime along with its session key, but that's left for a cleanup job in https://github.com/Manta-Network/Manta/issues/740

ghzlatarev:
Btw the docker test is now passing which shows that the new v3.3.0 client can sync with the production 3210 runtime.
https://github.com/Manta-Network/Manta/runs/7920446175?check_suite_focus=true

JoshOrndorff (commented):
Looks basically good to me.

One issue I see with the verifier is that it doesn't have logic to enforce a one-time switch from aura to nimbus. Consider that after the switch to nimbus, some Byzantine node authors an otherwise-valid block using aura logic and includes a valid aura seal. It seems the verifier would still accept that block.

A few moths back I was considering demonstrating a PoW -> Aura consensus handoff inspired by the upcoming Ethereum merge. I had brainstormed exactly this approach for writing the verifier. But I got stuck with the block authoring logic. Where are you telling the nodes to stop authoring with Aura and start authoring with Nimbus?

Garandor:
Thank you for taking the time to review!

> One issue I see with the verifier is that it doesn't have logic to enforce a one-time switch from aura to nimbus. Consider that after the switch to nimbus, some Byzantine node authors an otherwise-valid block using aura logic and includes a valid aura seal. It seems the verifier would still accept that block.

I believe this is addressed in a different place: After runtime upgrade the AuraApi will be replaced with a no-op implementation that always returns empty authorities so the client won't be able to verify an aura-produced block, even if it's valid.
At least so I thought. It turns out I forgot to make that change 😅. Thanks for pointing this out!

> A few moths back I was considering demonstrating a PoW -> Aura consensus handoff inspired by the upcoming Ethereum merge. I had brainstormed exactly this approach for writing the verifier. But I got stuck with the block authoring logic. Where are you telling the nodes to stop authoring with Aura and start authoring with Nimbus?

That's simple: We aren't :)
I sidestepped this tricky part by accepting that upgraded nodes will not be able to author with Aura and there will be a network partition between non-upgraded and upgraded nodes causing block time degradation around the runtime upgrade. Our product is not yet live so I can get away with that ;)
However, if you're looking for inspiration on how to do the block production handoff seamlessly too, Aura for cumulus got you covered: https://github.com/paritytech/cumulus/blob/e43b8b878a6fd0ca8b5e88d19822c4d777d3c677/polkadot-parachain/src/service.rs#L1026
They do something similar to onboard Aura on an already existing Parachain without it.

JoshOrndorff:
> However, if you're looking for inspiration on how to do the block production handoff seamlessly too, Aura for cumulus got you covered

Thanks for the pointer!

PR #771, opened 2022-08-24
Update tx-pause pallet to make it easier to pause stuff

### LINKED ISSUE #769: TX-pause updates & Maintenance Mode
Some improvements that will make our life easier if we need to respond to a hack:
- [x] 2 members of technical committee should be able to pause
- [x] Unpause via governance (root)
- [x]  Disallow pausing of balances and democracy pallets
- [ ] Maintenance mode design doc:
https://www.notion.so/mantanetwork/Parachain-Emergency-Mode-TX-Pause-Updates-Design-c4e6ebb80cc04586831f19a5d462a3db

ghzlatarev:
## Description

closes: #769 

* Splits the `UpdateOrigin` to `PauseOrigin` and `UnpauseOrigin` . `PauseOrigin` can be both technical committee (1/2) or governance. `UnpauseOrigin` is only governance
* Added a list of unpausable pallets - governance + balances
* ~~Add a special case for function name `txp_pause_all_pallets` which signifies that the whole pallet is filtered and added a new pause extrinsic that accepts an array of pallet names, to make it easier to pause en-masse quickly. Again the pause-origin is used.~~
* add `pause_pallets` and `unpause_pallets` where parameter is `pallet_name`, while `function_names` is get from metadata by `GetMetadata` trait
* add `pause_transactions` and `unpause_transactions` which parameter is both `pallet_name` and `Vec<function_name>`
* **NOTE: `MaxCallNames` is used for (pre-charge) weight calculation, and it means the extrinsic count of one pallet, current setting to 10. or we may increase to 20?**

image

---

write a little note why.

Situational Notes:


pallets/tx-pause/src/lib.rs
```diff
@@ -135,21 +142,200 @@ pub mod pallet {
...
+                if deposit_event {
+                    Self::deposit_event(Event::TransactionPaused(
+                        pallet_name.clone(),
+                        function_name.clone(),
+                    ));
+                }
+            }
+        });
+        Ok(())
+    }
```
ferrell-code, line 318: It may be ok to just emit a bunch of events for simplicity in the event mapping. Its not a call that will be used regularly so the overhead incurred isn't particularly costly.

This way is fine as is too, I don't feel strongly

pallets/tx-pause/src/lib.rs
```diff
@@ -64,15 +76,23 @@ pub mod pallet {
...
         InvalidCharacter,
+        /// can not resume
+        CannotResume,
+        /// call of pallet too many
+        TooManyCalls,
     }
 
     #[pallet::event]
-    #[pallet::generate_deposit(fn deposit_event)]
+    #[pallet::generate_deposit(pub fn deposit_event)]
```
ferrell-code:
```suggestion
    #[pallet::generate_deposit(pub(super) fn deposit_event)]
```

runtime/calamari/src/lib.rs
```diff
@@ -173,12 +173,20 @@ parameter_types! {
     pub const SS58Prefix: u8 = manta_primitives::constants::CALAMARI_SS58PREFIX;
 }
 
+parameter_types! {
+    pub UnpausablePallets: Vec<Vec<u8>> = vec![b"Democracy".to_vec(), b"Balances".to_vec(), b"Council".to_vec(), b"CouncilCollective".to_vec(), b"TechnicalCommittee".to_vec(), b"TechnicalCollective".to_vec()];
```
ghzlatarev: Do you think there's a way to no use strings here. Just wondering

pallets/tx-pause/src/lib.rs
```diff
@@ -135,21 +142,200 @@ pub mod pallet {
...
+            Ok(())
+        }
+
+        /// Pause all the calls of the listed pallets in `pallet_names`.
+        /// This logic is in its own extrinsic in order to not have to pause calls 1 by 1.
+        #[pallet::call_index(4)]
+        #[pallet::weight({
+            let len = pallet_names.len();
+            let max = T::MaxCallNames::get();
+            T::WeightInfo::pause_transaction().saturating_mul(len as Weight).saturating_mul(max as Weight)
```
ghzlatarev, line 204: I don't really understand what is this annotation used for, when we have an explicit return at the bottom.

ferrell-code: IIRC the explicit return is to refund some of the weight.

Worst case scenario is charged up front, then explicit return says how much is weight was actually consumed and caller is refunded the difference

zqhxuyuan: right, documentation here: https://docs.substrate.io/build/tx-weights-fees/#post-dispatch-weight-correction. But I need manual test to make sure it works. As current we're using root which can't test this scene. I'll try set to other account.

runtime/calamari/src/lib.rs
zqhxuyuan: As extrinsic parameter only accept `Vec<u8>`, the configuration has to be strings. My first thought of this improvement is align to `Call`, while we can config in runtime using `Call`, but inside exstrinsic, there're no other way to convert `Vec<u8>` to `Call`.

pallets/tx-pause/src/lib.rs
```diff
@@ -135,21 +142,200 @@ pub mod pallet {
...
+
             Ok(())
         }
+
+        /// Pause extrinsics by passing the extrinsic and corresponding pallet names.
+        /// Use names as they are written in the source code of the pallet.
+        #[pallet::call_index(2)]
+        #[pallet::weight({
+            let len = pallet_and_funcs.iter().flat_map(|item| {item.clone().1}).count();
+            T::WeightInfo::pause_transaction().saturating_mul(len as Weight)
```
ferrell-code, line 155: Could make these calls `DispatchClass::Operational` to ensure they get included in a block if someone is spamming transactions

Or make the priority higher for these transactions... could be done in a future PR

runtime/calamari/src/lib.rs
```diff
@@ -173,12 +173,20 @@ parameter_types! {
...
+}
+
 impl pallet_tx_pause::Config for Runtime {
     type Event = Event;
-    type UpdateOrigin = EitherOfDiverse<
+    type Call = Call;
+    type MaxCallNames = ConstU32<10>;
```
ferrell-code: Pallet Asset is the pallet with the most calls at 25 looking at the metadata...  25 maybe a reasonable parameter to set here for that reason as @ghzlatarev mentioned in top comment, although that would make it so less pallets could be paused in a single transaction.

We could address this with a maintenance mode later, but we would have to pause calls for `pallet_assets` individually right now.

ferrell-code (dismissed):
Logic looks sound to me. a few suggestions

pallets/tx-pause/src/lib.rs
ferrell-code: This is likely not relevant as pallet Collective calls are already `DispatchClass::Operational`

Garandor:
One thing to consider, if the network is being spammed, can we somehow force the `pause_system` extrinsic to be preferred in the client? Otherwise we could run the case where we run `pause` but it never gets included in a block

runtime/calamari/src/lib.rs
Garandor: `UnpausablePallets` should be `NonPausablePallets` otherwise it can be confused with `UnpauseOrigin`

runtime/manta/src/lib.rs
```diff
@@ -175,9 +176,17 @@ parameter_types! {
...
+parameter_types! {
+    pub UnpausablePallets: Vec<Vec<u8>> = vec![b"Democracy".to_vec(), b"Balances".to_vec(), b"Council".to_vec(), b"CouncilCollective".to_vec(), b"TechnicalCommittee".to_vec(), b"TechnicalCollective".to_vec()];
+}
+
 impl pallet_tx_pause::Config for Runtime {
     type Event = Event;
-    type UpdateOrigin = EnsureRoot<AccountId>;
+    type Call = Call;
+    type MaxCallNames = ConstU32<25>;
+    type PauseOrigin = EnsureRoot<AccountId>;
```
Garandor, line 187: i think this should also be >2-of-council, right?

runtime/calamari/src/lib.rs
Garandor: Non-pausable pallets don't **_need_** to be an extrinsic parameter, right? there is no extrinsic like `set_nonpausables` that would need to get the pallet names from the user.

We can just have a Vec<pallet_index_type> in the runtime. We would only update this with runtime upgrades and nothing else and the code just has a
```
let nonpausables = vec![
            <calamari_runtime::Runtime as frame_system::Config>::PalletInfo::index::<calamari_runtime::Democracy>(),
            <calamari_runtime::Runtime as frame_system::Config>::PalletInfo::index::<calamari_runtime::TechnicalCommittee>()
];
```
to match against. The boilerplate can be behind a helper `fn`

runtime/calamari/src/lib.rs
```diff
@@ -173,12 +173,20 @@ parameter_types! {
...
+
 impl pallet_tx_pause::Config for Runtime {
     type Event = Event;
-    type UpdateOrigin = EitherOfDiverse<
+    type Call = Call;
+    type MaxCallNames = ConstU32<25>;
+    type PauseOrigin = EitherOfDiverse<
         EnsureRoot<AccountId>,
-        pallet_collective::EnsureProportionMoreThan<AccountId, TechnicalCollective, 1, 2>,
+        pallet_collective::EnsureProportionMoreThan<AccountId, TechnicalCollective, 2, 6>,
```
Garandor: i think `EnsureProportionMoreThan` is wrong here. We want two or more people, not 1/3 of the committee

Garandor: I think `EnsureMembers` could work here

pallets/tx-pause/src/lib.rs
```diff
@@ -64,15 +76,23 @@ pub mod pallet {
         CannotPause,
         /// invalid character encoding
         InvalidCharacter,
+        /// can not resume
+        CannotResume,
```
Garandor: this shouldnt ever happen, right?

pallets/tx-pause/src/lib.rs
```diff
@@ -103,25 +123,12 @@ pub mod pallet {
             pallet_name: Vec<u8>,
             function_name: Vec<u8>,
```
Garandor, line 122: does it really make sense to want to pause individual extrinsics?

Garandor (changes requested):
Some questions and suggestions

runtime/manta/src/lib.rs
zqhxuyuan: We don't have council config yet on manta, so leave root here.

pallets/tx-pause/src/lib.rs
zqhxuyuan: right, It's in some of my dev iteration, forgot to delete.

pallets/tx-pause/src/lib.rs
zqhxuyuan: individual extrinsic pause/unpause is first ported from acala. As we've `pause_transactions` for multiple extrinsics, we could remove individual extrinsic pause/unpause. But I feel it's better to keep it for simply in some case we just want to pause/unpause one single extrinsic. Anyway, it's open for discussion.

runtime/calamari/src/lib.rs
zqhxuyuan: `PalletInfo::index::<calamari_runtime::Democracy>` return index of `Democracy` pallet in runtime. however `ensure_can_pause` only accept `pallet_name: Vec<u8>` as its parameter. I don't think `pallet_name` can be convert into `index`, thus they can't comparing. Unless we're changing `palelt_name` parameter type of pause extrinsics to `pallet_index`.

zqhxuyuan: renamed.

runtime/calamari/src/lib.rs
zqhxuyuan: done, use `EnsureMembers<AccountId, TechnicalCollective, 2>`.

runtime/calamari/src/lib.rs
Garandor: Right, i think we can just use the pallet index for pausing - these are not public-use extrinsics anyway.

Using index keeps the code simpler (since we can use the direct pallet types) but makes usage harder.
I personally think the tradeoff is worth it ( if we also get rid of individually pausing extrinsics ).

We'll have to build a tool to create batch pransactions for pausing anyway since we'll want to pause many pallets if we have to and calling this stuff individually is gonna take hours

pallets/tx-pause/src/lib.rs
Garandor: i meant to ask why we pause extrinsics and not pallets.
Is there any pallet where we want to pause some extrinsics and not others?
I cannot think of of one.
So i'd say pausing `pallets` is good enough and we could use either pallet names or pallet IDs to select which ones to pause.
This would help address @ghzlatarev 's question if we could avoid using `String`s

Garandor (commented):
See open question at https://github.com/Manta-Network/Manta/pull/771#discussion_r1015966280

I'm not convinced we need string parameters and extrinsic pausing

runtime/calamari/src/lib.rs
zqhxuyuan: If we use pallet index and the `PausedTransactions` storage now is: `Map(pallet_index, ())`, but the entry point from `Call` is still `pallet_name`, which means there needs to be some way convert `pallet_name` to `pallet_index`, so that we can check if given `pallet_name` of `Call` exists. I don't think we have this conversion.
https://github.com/Manta-Network/Manta/blob/0365ac76bacfaf41d5599e2a044dbf572026ae97/pallets/tx-pause/src/lib.rs#L152-L157

Garandor (dismissed):
LGTM

Garandor (commented):
LGTM after run-tests fix

Garandor (approved):

ferrell-code (approved):

PR #781, opened 2022-08-31
Proper XCM weights benchmark

### LINKED ISSUE #552: Integrate xcm benchmarked weights
We can add benchmarked weights to xcm pallets now that https://github.com/paritytech/polkadot/pull/4442 has landed

ghzlatarev:
## Description

closes: #552 

* Properly benchmarks XCM instruction calls, the same way as with normal runtime extrinsics, to get more accurate weights, instead of the fixed estimates we were using.
* Before 1 instruction was 1_000_000_000 and now on average for the different instructions it's 50 times less
* The newly generated weights:
    - Calamari: https://github.com/Manta-Network/Manta/actions/runs/2966879906
    - Dolphin: https://github.com/Manta-Network/Manta/actions/runs/2966880994
    - Update the benchmarks script and CI with the new weights

---

write a little note why.

Situational Notes:


.github/resources/xcm-weight-template.hbs
```diff
@@ -0,0 +1,67 @@
+// Copyright 2020-2022 Manta Network.
```
Dengjianping, line 1: Where can I find xcm weight template?

.github/resources/xcm-weight-template.hbs
```diff
@@ -0,0 +1,67 @@
...
+//! EXECUTION: {{cmd.execution}}, WASM-EXECUTION: {{cmd.wasm_execution}}, CHAIN: {{cmd.chain}}, DB CACHE: {{cmd.db_cache}}
+
+// Executed Command:
+{{#each args as |arg|}}
+// {{arg}}
+{{/each}}
+
+#![cfg_attr(rustfmt, rustfmt_skip)]
+#![allow(unused_parens)]
+#![allow(unused_imports)]
```
Dengjianping, line 31: Perhaps you need to add one more lint here:
```rust
#![allow(clippy::unnecessary_cast)]
```

runtime/calamari/src/weights/xcm/mod.rs
```diff
@@ -0,0 +1,178 @@
+// Copyright 2020-2022 Manta Network.
```
Dengjianping, line 1: Is this file automatically generated as well?

runtime/calamari/src/lib.rs
```diff
@@ -998,14 +1000,101 @@ impl_runtime_apis! {
...
+
+                fn claimable_asset() -> Result<(MultiLocation, MultiLocation, MultiAssets), BenchmarkError> {
+                    let origin = KmaLocation::get();
+                    let assets: MultiAssets = (Concrete(KmaLocation::get()), 1_000 * KMA).into();
+                    let ticket = MultiLocation { parents: 0, interior: Here };
+                    Ok((origin, ticket, assets))
+                }
+            }
+
             let whitelist: Vec<TrackedStorageKey> = vec![
```
Dengjianping, line 1099: Is there any xcm key we should whitelist?

.github/resources/xcm-weight-template.hbs
ghzlatarev: It's in the .github/resource folder like the other one.

runtime/calamari/src/weights/xcm/mod.rs
ghzlatarev: No, this one is manual.

runtime/calamari/src/lib.rs
ghzlatarev: I don't think so.
On Kusama they whitelist storage read to `ActiveConfig` https://github.com/paritytech/polkadot/pull/4442/files#diff-fc698698c7826b51219876977c4813bc7d173e8ecc1cdb963ca422ea3d300427R112
But this doesn't show up in my weights results, and anyway it whitelisting doesn't change my results noticeably.

.github/resources/xcm-weight-template.hbs
ghzlatarev: Not sure this is needed, clippy is not complaining at the moment.

.github/resources/xcm-weight-template.hbs
Dengjianping: Sorry, I didn't express it clearly. I mean where the template is from.

.github/resources/xcm-weight-template.hbs
Dengjianping: Yeah, it's fine if there's no clippy warning.

.github/resources/xcm-weight-template.hbs
ghzlatarev: It's from the Cumulus repo https://github.com/paritytech/cumulus/blob/master/templates/xcm-bench-template.hbs

.github/workflows/integration_test_dolphin.yml
```diff
@@ -394,7 +394,7 @@ jobs:
           aws-image-search-pattern: ${{ env.AWS_IMAGE_SEARCH_PATTERN }}
           aws-image-search-owners: ${{ env.AWS_IMAGE_SEARCH_OWNERS }}
   stop-node-builder-current:
-    if: ${{ always() }}
+    if: contains(github.event.pull_request.labels.*.name, 'A-dolphin' || github.ref == 'refs/heads/manta')
```
Garandor: I don't think this is safe, i thought of doing it this way before but then switched to always.
If someone labels the PR, the job runs. If they then unlabel the PR before CI is complete ( labelled by mistake ) this job will be skipped, leaving a zombie instance behind.

I think running always and ignoring success is the way to go here

ghzlatarev: Actually just labeled and unlabeled and the job just finished the whole run. But if you then try to add the label again then it will cancel the running job and make it hang.

ghzlatarev: Also adding a label restarts both calamari and dolphin, if one of them was already present.

ghzlatarev: I'll revert this change and leave it for a separate PR.

runtime/calamari/src/lib.rs
```diff
@@ -998,14 +1000,101 @@ impl_runtime_apis! {
...
             impl cumulus_pallet_session_benchmarking::Config for Runtime {}
 
+            use pallet_xcm_benchmarks::asset_instance_from;
+            use xcm_config::{LocationToAccountId, XcmExecutorConfig};
+
+            parameter_types! {
+                pub const TrustedTeleporter: Option<(MultiLocation, MultiAsset)> = None;
+                pub const TrustedReserve: Option<(MultiLocation, MultiAsset)> = Some((
+                    KsmLocation::get(),
+                    MultiAsset { fun: Fungible(1_000_000_000_000), id: Concrete(KsmLocation::get()) },
```
Dengjianping, line 1019: Can you leave a comment here, why `1_000_000_000_000` though you have explained it in the PR?
Perhaps I will forget why it's this number next time I go through this piece of code.

ghzlatarev: Sure, this is just a random number, doesn't really matter

Garandor (approved):
LGTM!

Dengjianping (approved):

PR #782, opened 2022-09-01
Run manta-pay randomized tests in a loop 10 times

### LINKED ISSUE #704: Run manta-pay randomized tests in a loop in CI
MantaPay tests are currently randomized to a degree, which is great, however sometimes it could take a few tries to catch bugs, depending on what kind of random values are used in each instance.
That's why we should run these tests say 2 - 5 time, however there's a tradeoff that this will slow down the CI, which should be balanced.
This could be a good first issue for someone as it is just changes in CI and measurements.

ghzlatarev:

## Description

closes: #704

* Run the randomized manta-pay tests in a loop a few items to avoid false positives.

---

write a little note why.

Situational Notes:


pallets/manta-pay/src/test/payment.rs
```diff
@@ -313,155 +314,186 @@ fn initialize_test(id: AssetId, value: AssetValue) {
...
-        mint_tokens(
-            asset_id,
-            &value_distribution(5, total_free_supply, &mut rng),
-            &mut rng,
-        );
-    });
+    for _ in 0..RANDOMIZED_TESTS_ITERATIONS {
+        let mut rng = OsRng;
+        new_test_ext().execute_with(|| {
+            let asset_id = rng.gen();
```
Dengjianping, line 320: Can be the random number sampled from a range like `(0, u128::MAX)`?

Dengjianping: I mean the `amount` not the asset id.

ghzlatarev: no, asset-id is still u32.

Dengjianping: Some of balances can be from (0, u128::MAX) or (0, u64::MAX), depending on the type of integer.

pallets/manta-pay/src/test/payment.rs
```diff
@@ -313,155 +314,186 @@ fn initialize_test(id: AssetId, value: AssetValue) {
...
-        let asset_id = rng.gen();
-        let total_free_supply = AssetValue(rng.gen());
-        initialize_test(asset_id, total_free_supply + DEFAULT_ASSET_ED);
-        mint_tokens(
-            asset_id,
-            &value_distribution(5, total_free_supply, &mut rng),
-            &mut rng,
-        );
-    });
+    for _ in 0..RANDOMIZED_TESTS_ITERATIONS {
```
Garandor: nit: start from 1 so even if someone puts RANDOMIZED_TESTS_ITERATIONS: usize = 0;
it'll run at least once

Garandor (approved):
LGTM

Garandor (dismissed):
Actually, fix linter

ghzlatarev:
> Actually, fix linter

yeah im waiting for #781 where I fixed the linter.

Garandor (approved):
LGTM!

Dengjianping (approved):

PR #791, opened 2022-09-08
Fix CI by ignoring failure of `stop-` jobs

### LINKED ISSUE #608: Investigate splitting up the publish_draft_releases workflow

Garandor:
stop-jobs run always even if their corresponding `start-` job has been skipped in order to avoid situations where we could get left-over zombie runners.

To not fail CI if a runner can't be stopped (e.g. because it has never been started), we ignore failure of stop- jobs

## Description

closes: #608

---

write a little note why.

Situational Notes:


Garandor:
Well, I was hoping the CI status would show success in this case, not the dreaded `X` for failure, guess this needs another solution, maybe we could add a second step to each stop job that just returns true and have `continue_on_error` on the aws stop step of the job.

@ghzlatarev could you maybe have a stab at this? The relevant doc is here: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

Dengjianping (dismissed):

ghzlatarev:
> Well, I was hoping the CI status would show success in this case, not the dreaded `X` for failure, guess this needs another solution, maybe we could add a second step to each stop job that just returns true and have `continue_on_error` on the aws stop step of the job.
> 
> @ghzlatarev could you maybe have a stab at this? The relevant doc is here: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

Will look into it some more, all of my ideas were not going to work as well :D

ghzlatarev:
@Garandor we can go with always running the start/stop jobs and just skip the middle parts, like in my last commit.
That's going to resolve our problem without `continue-on-error: true` but it's a bit less efficient.
I'm ok with this, just copy the solution for the other jobs if you agree.

.github/workflows/integration_test_calamari.yml
```diff
@@ -408,7 +408,7 @@ jobs:
           if [ $? == 1 ]; then echo "Failed to finalize the target block - 6"; exit 1; fi
   # HELPER JOBS BELOW
   start-node-builder-current:
-    if: contains(github.event.pull_request.labels.*.name, 'A-calamari' || github.ref == 'refs/heads/manta')
+    if: ${{ always() }}
```
ghzlatarev: revert this change

.github/workflows/integration_test_calamari.yml
```diff
@@ -444,8 +445,10 @@ jobs:
           aws-region: ${{ needs.start-node-builder-current.outputs.aws-region }}
           runner-label: ${{ needs.start-node-builder-current.outputs.runner-label }}
           aws-instance-id: ${{ needs.start-node-builder-current.outputs.aws-instance-id }}
+      - name: discard stopper success/failure
+        run: true
   start-calamari-integration-tester:
-    if: contains(github.event.pull_request.labels.*.name, 'A-calamari' || github.ref == 'refs/heads/manta')
+    if: ${{ always() }}
```
ghzlatarev: and this

.github/workflows/integration_test_calamari.yml
```diff
@@ -481,8 +485,10 @@ jobs:
           aws-region: ${{ needs.start-calamari-integration-tester.outputs.aws-region }}
           runner-label: ${{ needs.start-calamari-integration-tester.outputs.runner-label }}
           aws-instance-id: ${{ needs.start-calamari-integration-tester.outputs.aws-instance-id }}
+      - name: discard stopper success/failure
+        run: true
   start-docker-image-tester:
-    if: contains(github.event.pull_request.labels.*.name, 'A-calamari' || github.ref == 'refs/heads/manta')
+    if: ${{ always() }}
```
ghzlatarev: and this

ghzlatarev (approved):

ghzlatarev:
nice

flame4 (approved):

Dengjianping (approved):

PR #748, opened 2022-08-08
Upgrade to Manta v1.0.0

bhgomes:
Signed-off-by: Brandon H. Gomes <bhgomes@pm.me>

## Description

relates to OR closes: #XXXX

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Linked to Github issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [x] Added **one** label out of the `L-` group to this PR
- [ ] Added **one or more** labels out of `A-` and `C-` groups to this PR
- [ ] Re-reviewed `Files changed` in the Github PR explorer.

Situational Notes:
- If adding functionality, write unit tests!
- If importing a new pallet, choose a proper module index for it, and allow it in `BaseFilter`. Ensure **every** extrinsic works from front-end. If there's corresponding tool, ensure both work for each other.
- If needed, update our Javascript/Typescript APIs. These APIs are officially used by exchanges or community developers.
- If modifying existing runtime storage items, make sure to implement storage migrations for the runtime and test them with `try-runtime`. This includes migrations inherited from upstream changes, and you can search the diffs for modifications of `#[pallet::storage]` items to check for any.
- If runtime changes, need to update the version numbers properly:
   * `authoring_version`: The version of the authorship interface. An authoring node will not attempt to author blocks unless this is equal to its native runtime.
   * `spec_version`: The version of the runtime specification. A full node will not attempt to use its native runtime in substitute for the on-chain Wasm runtime unless all of spec_name, spec_version, and authoring_version are the same between Wasm and native.
   * `impl_version`: The version of the implementation of the specification. Nodes are free to ignore this; it serves only as an indication that the code is different; as long as the other two versions are the same then while the actual code may be different, it is nonetheless required to do the same thing. Non-consensus-breaking optimizations are about the only changes that could be made which would result in only the impl_version changing.
   * `transaction_version`: The version of the extrinsics interface. This number must be updated in the following circumstances: extrinsic parameters (number, order, or types) have been changed; extrinsics or pallets have been removed; or the pallet order in the construct_runtime! macro or extrinsic order in a pallet has been changed. You can run the `metadata_diff.yml` workflow for help. If this number is updated, then the `spec_version` must also be updated
- Verify benchmarks & weights have been updated for any modified runtime logics

PR #814, opened 2022-09-29
feat: upgrade asset manager

### LINKED ISSUE #808: AssetManager AssetId migration
AssetId type changes from u32 to u128
Migrate all existing storage items using it on Calamari

bhgomes:

## Description

closes: #808 

* Switch to u128 for `AssetId` of Calamari asset-manager pallet and add migration code for asset-manager and pallet-assets storage items that use `AssetId`. MantaPay `AssetId` to follow in a different PR. Only switch Calamari not Dolphin, because we don't need the migration before we restart it.
* The invariants that are checked are that there is no data in the new keys prior to the migration and that the same data is successfully migrated to the new keys. We also check manually that the old keys no longer hold any data.
* Turn `AssetId` and `Balance` of `AssetConfig` into generics
* Refactor some of the FungibleLedger functions to have try-versions. Rename some of the api functions.
* General improvements of related code

* Main files to review: 
    * [pallets/asset-manager/src/lib.rs](https://github.com/Manta-Network/Manta/pull/814/files#diff-0da041c944b5a8ca26aaae94530c2697f5cabe129ca90cb76b79c4f823d5c5be)
    * [primitives/manta/src/assets.rs](https://github.com/Manta-Network/Manta/pull/814/files#diff-2dbbc7f7ea01e95e39511d79b63953f7851509a2aa5cda9ecc73cc079b17e9e8)
    * runtime/calamari/src/migrations/asset_id.rs

Related question to why we can't assert that the old asset-id storage keys are empty in the post-upgrade try-runtime hook - https://substrate.stackexchange.com/questions/5214/potential-storage-issue-with-try-runtime-tool
But this was manually checked on a local testnet that the old storage keys are indeed drained.

Result of try-runtime:
```
2022-11-07 20:53:45 Connection established to target: Target { sockaddrs: [], host: "ws.calamari.systems", host_header: "ws.calamari.systems:443", _mode: Tls, path_and_query: "/" }
2022-11-07 20:53:45 scraping key-pairs from remote @ 0x089effe921771abf95df0623ee32ead2aab3d2b9e809b45af5300e0f72f0fbf9    
2022-11-07 20:53:45 downloading data for all pallets.    
2022-11-07 20:54:30 adding data for hashed key: 26aa394eea5630e07c48ae0c9558cef7f9cce9c888469bb1a0dceaa129672ef8    
2022-11-07 20:54:30 extending externalities with 1 manually injected key-values    
2022-11-07 20:54:30 👩‍👦 scraping child-tree data from 0 top keys    
2022-11-07 20:54:30 Custom("[backend]: frontend dropped; terminate client")
2022-11-07 20:54:30 injecting a total of 64079 top keys    
2022-11-07 20:54:31 injecting a total of 0 child keys    
2022-11-07 20:54:31 initialized state externalities with storage root 0xddadddff54a674394eeaeeef9d15b8b3571648c977c5a75334fb92a28c1e0e69    
2022-11-07 20:54:32 Connection established to target: Target { sockaddrs: [], host: "ws.calamari.systems", host_header: "ws.calamari.systems:443", _mode: Tls, path_and_query: "/" }
2022-11-07 20:54:32 found matching spec name: "calamari"    
2022-11-07 20:54:32 found matching spec version: 3432    
2022-11-07 20:54:32 Custom("[backend]: frontend dropped; terminate client")
2022-11-07 20:54:32 Starting migration for AssetManager...    
2022-11-07 20:54:32 Storage migration for AssetManager's AssetIdLocation storage item has been executed.    
2022-11-07 20:54:32 Storage migration for AssetManager's LocationAssetId storage item has been executed.    
2022-11-07 20:54:32 Storage migration for AssetManager's AssetIdMetadata storage item has been executed.    
2022-11-07 20:54:32 Storage migration for AssetManager's UnitsPerSecond storage item has been executed.    
2022-11-07 20:54:32 Storage migration for AssetManager's NextAssetId storage item has been executed.    
2022-11-07 20:54:32 ✅ Storage migration for AssetManager has been executed successfully and storage version has been update to: 2.    
2022-11-07 20:54:32 Starting migration for pallet-assets...    
2022-11-07 20:54:32 Storage migration for Assets' Asset storage item has been executed.    
2022-11-07 20:54:32 Storage migration for Assets' Account storage item has been executed.    
2022-11-07 20:54:32 Storage migration for Assets' Metadata storage item has been executed.    
2022-11-07 20:54:32 ✅ Storage migration for Assets has been executed successfully and storage version has been update to: 1.    
2022-11-07 20:54:32 ✅ no migration for System    
2022-11-07 20:54:32 ✅ no migration for CalamariVesting    
2022-11-07 20:54:32 ✅ no migration for AssetManager    
2022-11-07 20:54:32 ✅ no migration for Assets    
2022-11-07 20:54:32 ✅ no migration for Multisig    
2022-11-07 20:54:32 ✅ no migration for Utility    
2022-11-07 20:54:32 ✅ no migration for XTokens    
2022-11-07 20:54:32 ✅ no migration for DmpQueue    
2022-11-07 20:54:32 ✅ no migration for CumulusXcm    
2022-11-07 20:54:32 ⚠️ PolkadotXcm declares internal migrations (which *might* execute). On-chain `StorageVersion(0)` vs current storage version `StorageVersion(0)`    
2022-11-07 20:54:32 ⚠️ XcmpQueue declares internal migrations (which *might* execute). On-chain `StorageVersion(1)` vs current storage version `StorageVersion(1)`    
2022-11-07 20:54:32 ✅ no migration for Scheduler    
2022-11-07 20:54:32 ✅ no migration for Preimage    
2022-11-07 20:54:32 ✅ no migration for Treasury    
2022-11-07 20:54:32 ✅ no migration for Aura    
2022-11-07 20:54:32 ✅ no migration for Session    
2022-11-07 20:54:32 ✅ no migration for CollatorSelection    
2022-11-07 20:54:32 ✅ no migration for Authorship    
2022-11-07 20:54:32 ✅ no migration for AuraAuthorFilter    
2022-11-07 20:54:32 ✅ no migration for AuthorInherent    
2022-11-07 20:54:32 ✅ no migration for ParachainStaking    
2022-11-07 20:54:32 ✅ no migration for TechnicalMembership    
2022-11-07 20:54:32 ✅ no migration for TechnicalCommittee    
2022-11-07 20:54:32 ✅ no migration for CouncilMembership    
2022-11-07 20:54:32 ✅ no migration for Council    
2022-11-07 20:54:32 ✅ no migration for Democracy    
2022-11-07 20:54:32 ✅ no migration for TransactionPayment    
2022-11-07 20:54:32 ✅ no migration for Balances    
2022-11-07 20:54:32 ✅ no migration for TransactionPause    
2022-11-07 20:54:32 ✅ no migration for ParachainInfo    
2022-11-07 20:54:32 ✅ no migration for Timestamp    
2022-11-07 20:54:32 ⚠️ ParachainSystem declares internal migrations (which *might* execute). On-chain `StorageVersion(1)` vs current storage version `StorageVersion(1)`    
2022-11-07 20:54:32 ✅ Storage migration for AssetManager's AssetIdLocation storage item has been executed successfully.    
2022-11-07 20:54:32 ✅ Storage migration for AssetManager's LocationAssetId storage item has been executed successfully.    
2022-11-07 20:54:32 ✅ Storage migration for AssetManager's AssetIdMetadata storage item has been executed successfully.    
2022-11-07 20:54:32 ✅ Storage migration for AssetManager's UnitsPerSecond storage item has been executed successfully.    
2022-11-07 20:54:32 ✅ Storage migration for AssetManager's NextAssetId storage item has been executed successfully.    
2022-11-07 20:54:32 ✅ Storage migration for Assets' Asset storage item has been executed successfully.    
2022-11-07 20:54:32 ✅ Storage migration for Assets' Account storage item has been executed successfully.    
2022-11-07 20:54:32 ✅ Storage migration for Assets' Metadata storage item has been executed successfully.    
2022-11-07 20:54:32 proof: 00000000000000000000000000000000244d6f6f6e7269766572104d4f5652120000000000000000000000000000000000304163616c6120446f6c6c617210415553440c0000000000000000000000000000000000444b6172757261204c6971756964204b534d104c4b534d0c00000000000000000000000000000000004c4b6172757261204e617469766520546f6b656e0c4b41520c002043616c616d6172690c4b4d410c000000e8764817000000000000000000000001244d6f6f6e7269766572104d4f565212000000008a5d78456301000000000000000001250f64652af8a98da6f42e6eec6d15f964ab324d46889d2f9e67eb3ab6afedf07de95ac7304163616c6120446f6c6c617210415553440c000000e40b54020000000000000000000000013f072c12285b5d4551f88e8f6e7eb52b810100000085814a7236281c16b0fda4cfb8703c2d812e4895a67bd0a55cf8a1708235807a3f080436db8bf2d563cb5d2999a82809eded0b0000004b6fe6541b739c2191023bbb5678e77d4668c868fd7506763680f4376c51fb453f080436db8bf2d563cb5d2999a82809eded0b000000b379f4b0c9897881aa84d53a72e71854805da327dd6d45d1d4d8b1ad9338defd3f080436db8bf2d563cb5d2999a82809eded0b000000cb4c9e2e6aa6069223e22ab5a99733c7e4ac72249fd17125776307fad169b8ad3f0805...06aca508a2c0a73c18ff700d5ab9b7a70aa15a76d84d95cb61adfc8bf8b8ffb80c000000200480323ccbf2d2073d768b1dc4546b42999aef966d7b058fc49abb8cbe805b04c8cc80af4f4109ea0c9d14b5c82609227b9f653d04c4c62c1395873a1b1519037798aca70e19bf46066d74446ee15c8c2715f7030d000000280480cd3e3bf5669614bac02cd1179605106e99aaf3f221a73d929e71957054e22f2280419d54780bded1897cff3f854e27db5701302ec42bb8c0bd15d3eae29e3a3355806a7a38a2bf0e9f1c98b6a411f70116427de6a80b123a6a53440fd0cb1d9222ada70e1f3931028cc05c2e18a319e8f64f9e08000000210480191fc2e0dbc7baa0c1b06d2d92d7dde8848ccdfa73e04b9d6392abb411b565ad80d3d816fc1f7cf147f744017f5fab0aa253a527bae0078808968beaa1daedd661808537bdbb3acc2e767192560c45fd15cc48d4ac698528c3609390a7690df92607bf300cd15a3fd6e04e47bee3922dbfa92c8d826869651d2963939b862dcfb8bc6b7d419e3ce659b3c1704b7dea3ec3ee2745de72e8948330af2d400480bb196235f93acd40fb1562587e0964bfc739619fb39d307ead8f03393c1617df805142ad38e5b94dbbf1e1fab52b777a99344a4c7baf611d40f67b8f0c87be0753c10340000080853ad76c7f4a8d11ae49f5ecb01a7a216fa5d149fdb098a600f2dba9f41e392f / 103 nodes    
2022-11-07 20:54:32 proof size: 879.43 KB (900538 bytes)    
2022-11-07 20:54:32 compact proof size: 876.74 KB (897781 bytes)    
2022-11-07 20:54:32 zstd-compressed compact proof 876.77 KB (897808 bytes)    
2022-11-07 20:54:32 TryRuntime_on_runtime_upgrade executed without errors. Consumed weight = 5150000000, total weight = 500000000000 (0.0103)  
```

---

Situational Notes:


primitives/manta/Cargo.toml
```diff
@@ -17,7 +17,7 @@ scale-info = { version = "2.1.2", default-features = false, features = ["derive"
 smallvec = "1.8.0"
 
 # manta-rs dependencies
-manta-accounting = { git = "https://github.com/manta-network/manta-rs.git", tag = "v0.5.4", default-features = false }
+# manta-pay = { git = "https://github.com/manta-network/manta-rs.git", branch = "feat/new-circuits", default-features = false }
```
ghzlatarev: commented code

primitives/manta/src/assets.rs
```diff
@@ -14,128 +14,200 @@
...
+    /// to `true`, then non-zero balances may be stored without a `consumer` reference (and thus an
+    /// existential deposit in the balances pallet or whatever else is used to control user-account
+    /// state growth).
+    ///
+    /// If this flag is set to `false`, a fresh account cannot receive XCM tokens.
     pub is_sufficient: bool,
 }
 
+/*
 impl Default for AssetRegistrarMetadata {
```
ghzlatarev: As we discussed we probably don't need this, but if we keeping it let's make it more appropriate.

primitives/manta/src/assets.rs
```diff
@@ -217,52 +278,52 @@ impl From<AssetLocation> for Option<MultiLocation> {
...
 }
 
-/// Converter struct implementing `Convert`.
-/// This enforce the `AssetInfoGetter` implements `AssetIdLocationGetter`
-pub struct AssetIdLocationConvert<AssetLocation, AssetInfoGetter>(
-    PhantomData<(AssetLocation, AssetInfoGetter)>,
-);
-impl<AssetLocation, AssetInfoGetter> xcm_executor::traits::Convert<MultiLocation, AssetId>
-    for AssetIdLocationConvert<AssetLocation, AssetInfoGetter>
+///
```
ghzlatarev: missind docs

pallets/asset-manager/src/lib.rs
```diff
@@ -537,8 +638,11 @@ pub mod pallet {
         }
     }
 
-    /// Check the multilocation is supported by calamari/manta.
-    impl<T: Config> Contains<MultiLocation> for Pallet<T> {
+    impl<T> Contains<MultiLocation> for Pallet<T>
```
ghzlatarev, line 609: missing docs

pallets/asset-manager/src/lib.rs
```diff
@@ -559,29 +663,13 @@ pub mod pallet {
         }
     }
 
-    /// Get min-xcm-fee by multilocation.
-    impl<T: Config> GetByKey<MultiLocation, Option<u128>> for Pallet<T> {
+    impl<T> GetByKey<MultiLocation, Option<u128>> for Pallet<T>
```
ghzlatarev, line 635: missing docs

pallets/asset-manager/src/lib.rs
```diff
@@ -510,7 +611,7 @@ pub mod pallet {
                     | Junctions::X5(Junction::Parachain(para_id), ..)
                     | Junctions::X6(Junction::Parachain(para_id), ..)
                     | Junctions::X7(Junction::Parachain(para_id), ..)
-                    | Junctions::X8(Junction::Parachain(para_id), ..) => Some(*para_id),
+                    | Junctions::X8(Junction::Parachain(para_id), ..) => Some(para_id),
```
zqhxuyuan: We can simplify with `first_interior` like this:
```rust
            location.and_then(|location| {
                if let Some(Parachain(para_id)) = location.first_interior() {
                    Some(para_id)
                } else {
                    None
                }
            })
```

ghzlatarev: great

pallets/asset-manager/Cargo.toml
```diff
@@ -1,52 +1,56 @@
 [package]
-authors = ['Manta Network']
+authors = ["Manta Network"]
 edition = "2021"
-homepage = 'https://manta.network'
-license = 'GPL-3.0'
+homepage = "https://manta.network"
+license = "GPL-3.0"
 name = "pallet-asset-manager"
 repository = 'https://github.com/Manta-Network/Manta/'
```
Dengjianping, line 7: If you want to move all `'` to `"`, it might be changed as well.

pallets/asset-manager/Cargo.toml
```diff
@@ -1,52 +1,56 @@
...
 [dependencies]
-codec = { package = "parity-scale-codec", version = '3.1.2', default-features = false }
-frame-benchmarking = { git = 'https://github.com/paritytech/substrate.git', branch = "polkadot-v0.9.26", default-features = false, optional = true }
+codec = { package = "parity-scale-codec", version = "3.1.2", default-features = false }
+frame-benchmarking = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.26", default-features = false, optional = true }
 frame-support = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.26", default-features = false }
 frame-system = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.26", default-features = false }
 log = { version = "0.4.0", default-features = false }
 manta-primitives = { path = "../../primitives/manta", default-features = false }
+# manta-util = { git = "https://github.com/manta-network/manta-rs", branch = "feat/new-circuits", default-features = false }
```
Dengjianping: Can commented line be removed?

pallets/asset-manager/src/lib.rs
```diff
@@ -51,61 +52,59 @@ pub mod pallet {
...
         type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
 
-        /// Asset configuration, e.g. AssetId, Balance, Metadata
-        type AssetConfig: AssetConfig<Self>;
+        /// Asset Id Type
+        type AssetId: CheckedIncrement
+            + Default
+            + Parameter
+            + Serialize
+            + for<'de> Deserialize<'de>
```
Dengjianping: Can it be replaced `MaybeSerializeDeserialize` by ? If it works, `serde` can be removed.

pallets/asset-manager/Cargo.toml
```diff
@@ -1,52 +1,56 @@
...
-pallet-assets = { git = 'https://github.com/paritytech/substrate.git', branch = "polkadot-v0.9.26" }
-pallet-balances = { git = 'https://github.com/paritytech/substrate.git', branch = "polkadot-v0.9.26" }
+pallet-assets = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.26" }
+pallet-balances = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.26" }
 sp-core = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.26" }
 sp-io = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.26" }
 
 [features]
 default = ["std"]
 runtime-benchmarks = [
```
Dengjianping, line 34: For what I saw, `pallet-assets` is optional for this pallet, only for benchmakring, so it's unnecessary to built into runtime. You can mark `pallet-assets` as `optional = true`, then change this feature like 
```toml
runtime-benchmarks = [
  "dep:frame-benchmarking",
  "frame-support/runtime-benchmarks",
  "frame-system/runtime-benchmarks",
  "manta-primitives/runtime-benchmarks",
  "pallet-assets?/runtime-benchmarks",
]
```
Now when compile the pallet, the count of crates needed to be compiled(`cargo b`) will be reduced from `356` to `336`.
See this cargo feature: https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html#new-syntax-for-cargo-features

But we have to bump `rust-version`.

pallets/asset-manager/src/migrations.rs
```diff
@@ -23,35 +25,35 @@ use frame_support::{
...
+                    if *para_id != 2084 {
+                        let _ = Pallet::<T>::increase_count_of_associated_assets(*para_id);
                         reads += 1; // There's one read in method increase_count_of_associated_assets.
                         writes += 1; // There's one write in method increase_count_of_associated_assets.
                     }
                 }
             });
-
             // Update storage version.
             StorageVersion::new(1u16).put::<T>();
```
Dengjianping, line 56: Should it be bumped?

pallets/asset-manager/src/migrations.rs
```diff
@@ -65,22 +67,18 @@ impl<T: GetStorageVersion + Config + PalletInfoAccess> OnRuntimeUpgrade
     #[cfg(feature = "try-runtime")]
     fn pre_upgrade() -> Result<(), &'static str> {
         let storage_version = <T as GetStorageVersion>::on_chain_storage_version();
-
         if storage_version >= 1 {
```
Dengjianping, line 71: If it's bumped, change `1` to `2`.

pallets/asset-manager/src/migrations.rs
```diff
@@ -65,22 +67,18 @@ impl<T: GetStorageVersion + Config + PalletInfoAccess> OnRuntimeUpgrade
...
 
     #[cfg(feature = "try-runtime")]
     fn post_upgrade() -> Result<(), &'static str> {
         let storage_version = <T as GetStorageVersion>::on_chain_storage_version();
-
         if storage_version < 1 {
             return Err("Storage version is >= 1, the migration won't be executed.");
         }
-
         let acala = (2000, 3); // karura has 3 asset locations on calamari.
```
Dengjianping, line 83: How many parachains we're supporting, it should be more than these, right?

primitives/manta/Cargo.toml
```diff
@@ -16,9 +16,6 @@ log = "0.4.16"
 scale-info = { version = "2.1.2", default-features = false, features = ["derive"] }
 smallvec = "1.8.0"
 
-# manta-rs dependencies
-manta-accounting = { git = "https://github.com/manta-network/manta-rs.git", tag = "v0.5.4", default-features = false }
-
 # Substrate primitives
 frame-benchmarking = { git = 'https://github.com/paritytech/substrate.git', branch = "polkadot-v0.9.26", default-features = false, optional = true }
```
Dengjianping, line 20: `'`

primitives/manta/Cargo.toml
```diff
@@ -16,9 +16,6 @@ log = "0.4.16"
...
 frame-support = { git = 'https://github.com/paritytech/substrate.git', default-features = false, branch = "polkadot-v0.9.26" }
```
Dengjianping, line 21: `'`

primitives/manta/src/assets.rs
```diff
@@ -14,178 +14,237 @@
...
+            metadata: AssetStorageMetadata {
+                name: b"Default".to_vec(),
+                symbol: b"DEF".to_vec(),
+                decimals: 12,
+                is_frozen: false,
+            },
             evm_address: None,
-            is_frozen: false,
-            min_balance: DEFAULT_ASSET_ED,
+            min_balance: TEST_DEFAULT_ASSET_ED,
```
Dengjianping, line 222: For me, if I use `AssetRegistryMetadata::default()` to create an instance, I don't know it's for testing.

primitives/manta/src/assets.rs
```diff
@@ -217,52 +278,53 @@ impl From<AssetLocation> for Option<MultiLocation> {
...
+        M::asset_id(&location.borrow().clone().into()).ok_or(())
     }
 
-    fn reverse_ref(id: impl Borrow<AssetId>) -> Result<MultiLocation, ()> {
-        AssetInfoGetter::get_asset_location(*id.borrow())
+    #[inline]
+    fn reverse_ref(asset_id: impl Borrow<M::AssetId>) -> Result<MultiLocation, ()> {
+        M::location(asset_id.borrow())
             .and_then(Into::into)
             .ok_or(())
```
Dengjianping, line 316: Hope we can change the dummy error to concrete one.

primitives/manta/src/util/num.rs
```diff
@@ -0,0 +1,91 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Manta.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Numeric Traits
+
+/// Checked Addition
+pub trait CheckedAdd<Rhs = Self> {
```
Dengjianping: I'm not sure why we have this rust file. Basically, if we derive `AtLeast32BitUnsigned` or `BaseArithmetic`, that will have such methods like `checked_increment`.

runtime/calamari/src/migrations/asset_id.rs
```diff
@@ -0,0 +1,607 @@
...
+                &Blake2_128Concat::hash(&new_key.encode()),
+                value,
+            );
+            num_reads += 1;
+            num_writes += 1;
+        }
+
+        // LocationAssetId
+
+        let pallet_prefix: &[u8] = b"AssetManager";
```
Dengjianping, line 160: Seems it can be removed.

runtime/calamari/src/migrations/asset_id.rs
```diff
@@ -0,0 +1,607 @@
...
+                &Blake2_128Concat::hash(&old_key.encode()),
+                new_key,
+            );
+            num_reads += 1;
+            num_writes += 1;
+        }
+
+        // AssetIdMetadata
+
+        let pallet_prefix: &[u8] = b"AssetManager";
```
Dengjianping, line 185: Removable.

runtime/calamari/src/migrations/asset_id.rs
```diff
@@ -0,0 +1,607 @@
...
+                &Blake2_128Concat::hash(&new_key.encode()),
+                value,
+            );
+            num_reads += 1;
+            num_writes += 1;
+        }
+
+        // UnitsPerSecond
+
+        let pallet_prefix: &[u8] = b"AssetManager";
```
Dengjianping, line 222: Removable.

runtime/calamari/src/migrations/asset_id.rs
```diff
@@ -0,0 +1,607 @@
...
+                &Blake2_128Concat::hash(&new_key.encode()),
+                value,
+            );
+            num_reads += 1;
+            num_writes += 1;
+        }
+
+        // NextAssetId
+
+        let pallet_prefix: &[u8] = b"AssetManager";
```
Dengjianping, line 247: Removable

pallets/asset-manager/Cargo.toml
ghzlatarev: Yeah sure. These changes were automatic from taplo fmt, not sure why it didn't change the repository string as well.

pallets/asset-manager/src/migrations.rs
ghzlatarev: Yeah, forgot that, thanks. Also will add a check that storage version is correct for both pallets.

pallets/asset-manager/src/migrations.rs
ghzlatarev: This migration is not actually executed, but i'll add these checks to the asset-id migration as well.

pallets/asset-manager/src/migrations.rs
ghzlatarev: I think when we did this migration that was all of them. Right now we have KSM and PHA as well
But again this migration won't be executed now. This is just formatting change most likely.

primitives/manta/Cargo.toml
ghzlatarev: What do you mean here ?

runtime/calamari/src/migrations/asset_id.rs
ghzlatarev: This is clearer to read imo

primitives/manta/Cargo.toml
Dengjianping: Seems this pr is going to replace all `'` to `"`

ghzlatarev: That's not the intent of this PR. All of these type of changes are either from taplo fmt or accidental.
We can do this separately if we want to.

pallets/asset-manager/Cargo.toml
ghzlatarev: Yep

pallets/asset-manager/src/lib.rs
ghzlatarev: Cool thanks, looks like this can work.

primitives/manta/src/assets.rs
ghzlatarev: Right, I see how this can be confusing.
How about I make a trait with a more explicit name like `DefaultForTesting`

ghzlatarev: Ok I see how this can be confusing, especially since there's no such thing as a default asset metadata.
But this is just an easy convenient way to get a function that returns some asset that is of the correct type for testing/benchmarking in multiple modules.
I can add some comments for this or if you have a proper solution please let me know.

primitives/manta/src/util/num.rs
ghzlatarev: checked_increment is not included in those 2 traits

ghzlatarev: Ok I see what you mean , why no just use `checked_add(1)` instead of implementing `checked_increment`
@bhgomes did you have something specific in mind here ?

pallets/asset-manager/Cargo.toml
ghzlatarev: Actually it's not even needed for benchmarking so I'll remove it altogether.

primitives/manta/src/assets.rs
ghzlatarev: I don't think we can give it a different type as this is the signature of the trait.
But we already map the error to a concrete error where we use this impl - https://github.com/Manta-Network/Manta/pull/814/files#diff-d734041aae7312744fcd9c95049692edcd71695c761ff962916334a619aaf3b5R345

primitives/manta/src/assets.rs
ghzlatarev: Nvm I think I can just make another trait with better name `TestingDefault`

primitives/manta/src/assets.rs
```diff
@@ -217,52 +285,53 @@ impl From<AssetLocation> for Option<MultiLocation> {
...
 }
 
 /// Converter struct implementing `Convert`.
-/// This enforce the `AssetInfoGetter` implements `AssetIdLocationGetter`
-pub struct AssetIdLocationConvert<AssetLocation, AssetInfoGetter>(
-    PhantomData<(AssetLocation, AssetInfoGetter)>,
-);
-impl<AssetLocation, AssetInfoGetter> xcm_executor::traits::Convert<MultiLocation, AssetId>
-    for AssetIdLocationConvert<AssetLocation, AssetInfoGetter>
+/// This enforces that `AssetInfoGetter` implements `AssetIdLocationGetter`
```
zqhxuyuan: doc need update.

primitives/manta/src/assets.rs
```diff
@@ -14,178 +14,244 @@
...
+        metadata: Self::Metadata,
+    ) -> Result<(), Self::Error>;
 }
 
-pub trait AssetConfig<C>: 'static + Eq + Clone
+/// Asset Configuration
+pub trait AssetConfig<C>: AssetIdType + BalanceType + LocationType
 where
-    C: frame_system::Config,
+    C: Config,
```
zqhxuyuan, line 120: Do we really need frame_system::Config trait bound? Can we just use `<AccountId>` type parameter?
In `primitives` I guess we should depend as minimum as possible.

ghzlatarev: I'm actually not sure we can get away with just `AccountId` generic (or maybe i'm misunderstanding you)
You can log and give it a try separately, i feel like it will take a while to disentangle the code if it can be implemented that way.

primitives/manta/src/assets.rs
ghzlatarev: done, thanks!

primitives/manta/src/util/num.rs
bhgomes: @ghzlatarev yeah feel free to use those traits (and delete these files) as long as they work for `u32` and `u128`, I didn't know about them at the time of writing the changes

pallets/asset-manager/Cargo.toml
```diff
@@ -1,52 +1,53 @@
...
+  "log/std",
   "manta-primitives/std",
   "orml-traits/std",
-  'log/std',
-  'xcm/std',
+  "scale-info/std",
+  "sp-runtime/std",
+  "sp-std/std",
+  "xcm/std",
+  "pallet-assets/std",
```
Dengjianping: Seems it can be removed.

pallets/asset-manager/src/mock.rs
```diff
@@ -240,14 +261,16 @@ pub(crate) fn create_asset_metadata(
...
-        name: name.as_bytes().to_vec(),
-        symbol: symbol.as_bytes().to_vec(),
-        decimals,
+) -> AssetRegistryMetadata<Balance> {
+    AssetRegistryMetadata {
+        metadata: AssetStorageMetadata {
+            name: name.as_bytes().to_vec(),
+            symbol: symbol.as_bytes().to_vec(),
+            decimals,
+            is_frozen,
```
Dengjianping, line 265: 1. `AssetRegistrarMetadata` is renamed to `AssetRegistryMetadata`, so does old storage need to be removed from here?: https://github.com/Manta-Network/Manta/pull/814/files#diff-40225a2ea5f6a026936004b2aab2d0126e1f4c481c089528d70f81b819c507e9L195 
2. Because the filed `is_frozen` is moved to `metadata`, we might need do migration for `AssetRegistryMetadata`.

pallets/asset-manager/Cargo.toml
ghzlatarev: indeed

pallets/asset-manager/src/mock.rs
ghzlatarev: good catch
1. This one is not a storage item. It needs to be annotated with `storage` to get a storage key, so I don't think we need to remove anything https://github.com/paritytech/substrate/blob/master/frame/support/src/lib.rs#L215-L233
2. I would've assumed so, but actually executing this migration locally manages to migrate the `AssetRegistryMetadata`, which doesn't really make sense to me ...
But anyway I'll make this migration explicit in the code.

Dengjianping: But this is a storage to store asset metadata: https://github.com/Manta-Network/Manta/pull/814/files#diff-0da041c944b5a8ca26aaae94530c2697f5cabe129ca90cb76b79c4f823d5c5beR330

ghzlatarev: Right, I migrated it -> https://github.com/Manta-Network/Manta/pull/814/files#diff-e29c5b8bac12f21073b76e6b3971b09b03d1c986c9918c8be66b581f4ca5cb17R171-R202

ghzlatarev: Or do you mean something else ?

Dengjianping: Sorry, I didn't see it.

pallets/asset-manager/src/lib.rs
```diff
@@ -51,61 +52,58 @@ pub mod pallet {
...
     use sp_runtime::{
-        traits::{AccountIdConversion, One},
+        traits::{AccountIdConversion, MaybeSerializeDeserialize, One},
         ArithmeticError,
     };
     use xcm::latest::prelude::*;
 
-    const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
+    /// Storage Version
+    pub const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
```
ghzlatarev: update

pallets/asset-manager/src/lib.rs
```diff
@@ -484,38 +576,37 @@ pub mod pallet {
...
+        T: Config,
+    {
+        /// Returns and increments the [`NextAssetId`] by one.
+        #[inline]
+        fn next_asset_id_and_increment() -> Result<T::AssetId, DispatchError> {
+            NextAssetId::<T>::try_mutate(|current| {
                 let id = *current;
-                *current = current.checked_add(1u32).ok_or(ArithmeticError::Overflow)?;
+                *current = current
+                    .checked_add(&<T as Config>::AssetId::from(1u32))
```
Dengjianping: `<T as Config>::AssetId::from(1u32)` -> `One::one()`.
But I think both are fine, depends on you.

ghzlatarev: Nice, forgot about this.

Dengjianping (dismissed):

pallets/manta-pay/src/lib.rs
```diff
@@ -105,6 +105,12 @@ pub mod rpc;
 #[cfg(feature = "runtime")]
 pub mod runtime;
 
+/// Standard Asset Id
+pub type StandardAssetId = u32;
```
stechu, line 109: why set `StandardAssetId` here to `u32`?

ghzlatarev: Because this is not PR for manta-pay. That change is done in #748

Dengjianping (dismissed):

primitives/manta/src/types.rs
```diff
@@ -14,42 +14,56 @@
...
+/// This index is used to look up accounts.
 pub type AccountIndex = u32;
 
-/// Balance of an account.
-pub type Balance = AssetValueType;
+/// Calamari Asset Id Type
+pub type CalamariAssetId = u128;
+
+/// Dolphin Asset Id Type
+pub type DolphinAssetId = u32;
```
Garandor, line 48: Though you're saying "we won't need u128 asset ID until we restart it", we're gonna restart it next week for v3, so although we don't need the migration for dolphin, dolphin should use the new u128 asset ID as well, no?

primitives/manta/src/lib.rs
```diff
@@ -14,12 +14,17 @@
...
+#![cfg_attr(doc_cfg, feature(doc_cfg))]
+#![forbid(rustdoc::broken_intra_doc_links)]
+#![forbid(missing_docs)]
+
+extern crate alloc;
 
 pub mod assets;
 pub mod constants;
 pub mod types;
+///
```
Garandor: why this empty doccomment?

primitives/manta/src/xcm.rs
```diff
@@ -69,21 +71,27 @@ impl Reserve for MultiAsset {
...
         asset.reserve().map(|r| r == *origin).unwrap_or(false)
     }
 }
 
-pub struct AccountIdToMultiLocation<AccountId>(PhantomData<AccountId>);
-impl<AccountId> Convert<AccountId, MultiLocation> for AccountIdToMultiLocation<AccountId>
+/// AccountId to [`MultiLocation`] Converter
+pub struct AccountIdToMultiLocation;
+
+impl<A> Convert<A, MultiLocation> for AccountIdToMultiLocation
```
Garandor, line 91: the generics changes and doc enhancements are great, but they make the PR big and hard to reason about.
For the future please try keeping a single-responsibility-principle in PRs where there is only one functional change per PR that is not conflated with other changes.
While I understand that both changes fall under "Upgrade asset manager", if there's a large many-files refactoring, it's well worth to make that its own PR.

If you submit one PR with the non-functional changes, all we need to review is that they actually don't change any behavior - and then having second PR with only one functional change will make it much easier to verify that all changes only influence the intended behavior from the PR description in the intended way.
Combining multiple small functional changes is fine too as long as their code diffs don''t overlap too much and the PR doesn't grow too large.

As a PR author you know which changes belong to which feature, as a reviewer, I see the code for the first time and lose time trying to map a line to one of multiple changes.

I would have liked to see:
- PR1 Switch to u128 for AssetId of Calamari asset-manager pallet and add migration code for asset-manager and pallet-assets storage items that use AssetId
- PR2 Turn AssetId and Balance of AssetConfig into generics
- PR3 All of the docstring and format changes maybe with the session_keys cleanup + Refactor some of the FungibleLedger functions to have try-versions. 
- PR4 Rename some of the api functions

primitives/manta/src/assets.rs
```diff
@@ -306,210 +374,267 @@ impl FungibleLedgerError {
...
+    /// Deposit `amount` of an asset with the given `asset_id` to `account`.
+    fn deposit_minting(
+        asset_id: Self::AssetId,
+        account: &Self::AccountId,
+        amount: Self::Balance,
+    ) -> Result<(), FungibleLedgerError<Self::AssetId, Self::Balance>>;
+
+    /// Checks if `amount` of `asset_id` can be deposited into `account` and then deposits it if
+    /// that is successful.
+    fn try_deposit_minting(
```
Garandor, line 417: where is this used?

primitives/manta/src/assets.rs
```diff
@@ -306,210 +374,267 @@ impl FungibleLedgerError {
...
     }
 
     #[inline]
-    fn deposit_can_mint(
-        asset_id: AssetId,
-        beneficiary: &C::AccountId,
-        amount: Balance,
-    ) -> Result<(), FungibleLedgerError> {
-        Self::ensure_valid(asset_id)?;
+    fn try_deposit_minting(
```
Garandor: technically, both `deposit` and `try_deposit` are fallible and return result, the `try` naming is misleading, maybe something like `deposit_with_mint_check`?

Garandor: or `deposit_minting_with_check`

primitives/manta/src/types.rs
ghzlatarev: https://github.com/Manta-Network/Manta/issues/883

primitives/manta/src/lib.rs
ghzlatarev: forgotten, will add

primitives/manta/src/xcm.rs
ghzlatarev: yeah totally, I think Brandon and I were rushing again : (

primitives/manta/src/assets.rs
ghzlatarev: Personally I prefer the explicit naming like `deposit_minting_with_check` so I'll change it and revert if @bhgomes feels strongly about the try naming

primitives/manta/src/lib.rs
ghzlatarev: Oh no, just move the comment to the XCM file

runtime/calamari/tests/xcm_mock/parachain.rs
```diff
@@ -538,41 +548,52 @@ impl AssetRegistrar<Runtime, ParachainAssetConfig> for CalamariAssetRegistrar {
...
+        },
         min_balance: 1,
         evm_address: None,
-        is_frozen: false,
         is_sufficient: true,
     };
     pub const AssetManagerPalletId: PalletId = ASSET_MANAGER_PALLET_ID;
 }
 
+///
```
Garandor: empty doccomment

runtime/calamari/tests/xcm_mock/parachain.rs
```diff
@@ -694,14 +714,16 @@ pub(crate) fn create_asset_metadata(
...
-        name: name.as_bytes().to_vec(),
-        symbol: symbol.as_bytes().to_vec(),
-        decimals,
+) -> AssetRegistryMetadata<Balance> {
+    AssetRegistryMetadata {
+        metadata: AssetStorageMetadata {
+            name: name.as_bytes().to_vec(),
+            symbol: symbol.as_bytes().to_vec(),
+            decimals,
+            is_frozen,
```
Garandor, line 722: could you explain your thoughts behind grouping some as metadata and some not?
e.g. why is `is_frozen` metadat and `is_sufficient` not?

runtime/calamari/tests/integrations_mock/integration_tests.rs
```diff
@@ -1113,27 +1116,10 @@ fn concrete_fungible_ledger_transfers_work() {
                 0
             );
 
-            // Transferring invalid asset ID should not work.
```
Garandor, line 1116: `FungibleLedgerError::InvalidAssetId` still exists, why remove the test?

runtime/calamari/src/xcm_config.rs
```diff
@@ -286,17 +278,20 @@ impl cumulus_pallet_dmp_queue::Config for Runtime {
     type ExecuteOverweightOrigin = EnsureRoot<AccountId>;
 }
 
-// We wrap AssetId for XToken
+/// We wrap AssetId for XToken
 #[derive(Clone, Eq, Debug, PartialEq, Ord, PartialOrd, Encode, Decode, TypeInfo)]
 pub enum CurrencyId {
-    MantaCurrency(AssetId),
+    ///
```
Garandor: more doccomments

runtime/calamari/src/xcm_config.rs
```diff
@@ -286,17 +278,20 @@ impl cumulus_pallet_dmp_queue::Config for Runtime {
...
+    MantaCurrency(CalamariAssetId),
 }
 
-pub struct CurrencyIdtoMultiLocation<AssetXConverter>(sp_std::marker::PhantomData<AssetXConverter>);
+///
```
Garandor: more doccomments

pallets/manta-pay/src/test/payment.rs
```diff
@@ -294,20 +297,30 @@ where
...
-    assert_ok!(MantaAssetRegistrar::create_asset(
+    let metadata = AssetRegistryMetadata {
+        metadata: AssetStorageMetadata {
+            name: b"Calamari".to_vec(),
+            symbol: b"KMA".to_vec(),
+            decimals: 12,
+            is_frozen: false,
+        },
+        min_balance: TEST_DEFAULT_ASSET_ED,
+        evm_address: None,
```
Garandor: Is `evm_address` still needed as asset metadata?

pallets/manta-pay/src/benchmark/mod.rs
```diff
@@ -44,30 +44,30 @@ where
 
 /// Init assets for manta-pay
 #[inline]
-pub fn init_asset<T>(owner: &T::AccountId, id: AssetId, value: Balance)
+pub fn init_asset<T>(owner: &T::AccountId, id: u128, value: Balance)
```
Garandor: i thought mantapay does not yet change to `u128` in this PR?
Why not use `StandardAssetId` here?

Garandor:
> Result of try-runtime:

The log shows no indication that the assetmanager migration has executed, that seems wrong

runtime/calamari/src/migrations/asset_id.rs
```diff
@@ -0,0 +1,680 @@
...
+{
+    fn on_runtime_upgrade() -> Weight {
+        let mut num_reads = 0;
+        let mut num_writes = 0;
+
+        let asset_manager_storage_version = <M as GetStorageVersion>::on_chain_storage_version();
+        let assets_storage_version = <A as GetStorageVersion>::on_chain_storage_version();
+        num_reads += 2;
+        if asset_manager_storage_version != 1 || assets_storage_version != 0 {
+            return T::DbWeight::get().reads(num_reads as Weight);
```
Garandor, line 128: I would add a log:info!("not migrating") here

runtime/calamari/src/migrations/asset_id.rs
```diff
@@ -0,0 +1,680 @@
...
+        }
+
+        // AssetIdLocation
+
```
Garandor, line 132: and a log:info!("starting migration") here

runtime/calamari/src/migrations/asset_id.rs
```diff
@@ -0,0 +1,680 @@
...
+            put_storage_value(
+                pallet_prefix,
+                storage_item_prefix,
+                &Blake2_128Concat::hash(&new_key.encode()),
+                value,
+            );
+            num_reads += 1;
+            num_writes += 1;
+        }
+
```
Garandor, line 157: also progress reports like log::info!("Migrated {num_of_locations} AssetIdLocations")

You can get rid of the in-code comments if you just `log` the stages, that gives more info at runtime (especially if - god forbid - the migration fails on mainnet)

runtime/calamari/src/migrations/asset_id.rs
```diff
@@ -0,0 +1,680 @@
...
+    pub name: Vec<u8>,
+    pub symbol: Vec<u8>,
+    pub decimals: u8,
+    pub evm_address: Option<H160>,
+    pub is_frozen: bool,
+    pub min_balance: Balance,
+    pub is_sufficient: bool,
+}
+
+pub struct AssetIdMigration<T, M, A>(PhantomData<T>, PhantomData<M>, PhantomData<A>);
```
Garandor: could you document the meanings of TMA here? the trait bounds aren't enough to infer them

pallets/manta-pay/src/test/payment.rs
Garandor: > RLS — I don't think we needed it any more

Pls remove it with this migration

runtime/calamari/src/migrations/asset_id.rs
```diff
@@ -0,0 +1,680 @@
...
+        if assets_storage_version != 0 {
+            return Err("Assets storage version is not 0, the migration won't be executed.");
+        }
+
+        // We want to test that:
+        // There are no entries in the new storage beforehand
+        // The same number of mappings exist before and after
+        //
+        // We have to manually check that there are no entries
+        // in the old storage afterward due to a potential try-runtime bug
```
Garandor: this comment was resolved as not-a-bug, right?

pallets/asset-manager/src/migrations.rs
```diff
@@ -23,35 +25,35 @@ use frame_support::{
     traits::{Get, OnRuntimeUpgrade, PalletInfoAccess, StorageVersion},
 };
 
+///
```
Garandor: empty doccomment

pallets/asset-manager/src/benchmarking.rs
```diff
@@ -20,121 +20,120 @@ use crate::{Call, Config, Event, Pallet};
...
 
             let location: MultiLocation = MultiLocation::new(0, X1(Parachain(i)));
-            let location = <T::AssetConfig as AssetConfig<T>>::AssetLocation::from(location.clone());
-            let metadata = <T::AssetConfig as AssetConfig<T>>::AssetRegistrarMetadata::default();
+            let location = T::Location::from(location.clone());
+            let metadata = <T::AssetConfig as AssetConfig<T>>::AssetRegistryMetadata::testing_default();
 
             Pallet::<T>::register_asset(RawOrigin::Root.into(), location.clone(), metadata.clone())?;
-            Pallet::<T>::set_units_per_second(RawOrigin::Root.into(), i, 0)?;
+            Pallet::<T>::set_units_per_second(RawOrigin::Root.into(), next, 0)?;
```
Garandor: what does this edit `i` -> `next` change?

pallets/asset-manager/src/lib.rs
```diff
@@ -118,24 +115,93 @@ pub mod pallet {
         type WeightInfo: crate::weights::WeightInfo;
     }
 
+    /// Asset Manager Pallet
+    #[pallet::pallet]
+    #[pallet::generate_store(pub(super) trait Store)]
```
Garandor: pls remove -> https://substrate.stackexchange.com/a/1403/504

pallets/asset-manager/src/lib.rs
```diff
@@ -118,24 +115,93 @@ pub mod pallet {
...
+    /// Genesis Configuration
     #[pallet::genesis_config]
     pub struct GenesisConfig<T: Config> {
-        pub start_id: AssetId,
-        pub _marker: PhantomData<T>,
+        pub start_id: T::AssetId,
+    }
+
+    #[cfg(feature = "std")]
+    impl<T> GenesisConfig<T>
```
Garandor: why do you provide the `GenesisConfig` type here explicitly if you're just forwarding `build` and `assimilate_storage`?
other pallets don't provide this

pallets/asset-manager/src/lib.rs
```diff
@@ -401,13 +502,13 @@ pub mod pallet {
         #[transactional]
         pub fn set_units_per_second(
             origin: OriginFor<T>,
-            #[pallet::compact] asset_id: AssetId,
+            asset_id: T::AssetId,
```
Garandor: why not compac? it's still just a `u128`

Garandor (changes requested):
ok, i'm through. see comments for requested changes.
Most importantly, remove `evm_address`

pallets/manta-pay/src/benchmark/mod.rs
ghzlatarev: Nice catch, will change it. This code is a copy paste from the manta-pay-v1 branch to get it to compile.

ghzlatarev:
> > Result of try-runtime:
> 
> The log shows no indication that the assetmanager migration has executed, that seems wrong

That's because I didn't add any logs, will do. And the migration is not part of the pallet code so that's probably the reason for `2022-10-05 20:29:37 ✅ no migration for AssetManager `

runtime/calamari/src/migrations/asset_id.rs
ghzlatarev: yeah

pallets/asset-manager/src/benchmarking.rs
ghzlatarev: I think i was having some issues with the type checker.
I can just convert `i` an asset-id here

pallets/asset-manager/src/lib.rs
ghzlatarev: Yeah I don't think it's needed. Removing it unless @bhgomes had something else in mind.

pallets/asset-manager/src/benchmarking.rs
```diff
@@ -43,92 +43,91 @@ benchmarks! {
     }
 
     set_units_per_second {
-        let mut next = <T::AssetConfig as AssetConfig<T>>::StartNonNativeAssetId::get();
+        let assets_count = 1000;
         for i in 0..1000 {
```
Garandor: changeto `assets_count` here too

runtime/calamari/src/migrations/asset_id.rs
ghzlatarev: Actually i just need the T, will get rid of the rest, and that should clear things up.

runtime/calamari/tests/integrations_mock/integration_tests.rs
ghzlatarev: There's no DummyAssetId anymore

pallets/asset-manager/src/lib.rs
ghzlatarev: Yeah not sure why it got removed, but yeah it's still u128 so can be added to all the AssetId extrinsic params.

runtime/calamari/tests/xcm_mock/parachain.rs
ghzlatarev: This is all that is required for the pallet-assets metadata https://github.com/paritytech/substrate/blob/master/frame/assets/src/types.rs#L127
There's a comment here  https://github.com/Manta-Network/Manta/blob/manta/primitives/manta/src/assets.rs#L103

Dengjianping (dismissed):

ghzlatarev:
> > Result of try-runtime:
> 
> The log shows no indication that the assetmanager migration has executed, that seems wrong

Updated the try-runtime results with the logs now.

Garandor (approved):
LGTM!

Dengjianping (approved):

PR #835, opened 2022-10-12
align block number type in tests

### LINKED ISSUE #805: Align all `mock.rs` Runtimes to either Calamari or Dolphin Values
Mock runtimes and their tests use different configuration values than our actual runtime does, so we're testing with different config values than we actually run in prod.
We should align all of their constants and configs to the prod version (either calamari or dolphin) to make the tests more realistic

e.g. Blocknumber u64 -> u32

image

zqhxuyuan:
## Description

relates to OR closes: #805 

---

Situational Notes:


zqhxuyuan:
@ghzlatarev We can't use `u32` in mock(test fails), because `Header`(we're using sp_runtime::testing::Header) is using `u64`. unless we change to our own `Header` type.

pallets/parachain-staking/src/mock.rs
```diff
@@ -27,17 +27,17 @@ use frame_support::{
...
 };
 
 pub type AccountId = u64;
 pub type Balance = u128;
-pub type BlockNumber = u64;
 
 type UncheckedExtrinsic = frame_system::mocking::MockUncheckedExtrinsic<Test>;
 type Block = frame_system::mocking::MockBlock<Test>;
+type BlockNumber = u32;
+type Header = sp_runtime::generic::Header<BlockNumber, BlakeTwo256>;
```
Dengjianping: Can these types like `Header/BlockNumber` be imported from `primitives/manta/src/types.rs`?

ghzlatarev (commented):
Looks good, just need to fix the tests compilation and add labels.

ghzlatarev (dismissed):

ghzlatarev (approved):

Garandor (approved):

PR #836, opened 2022-10-12
client trait bound refactor

### LINKED ISSUE #837: node service trait bound refactor
service `RuntimeApi` has lots of common trait bound, should refactor to some trait.

zqhxuyuan:

## Description

relates to OR closes: #837 

---

Situational Notes:


Dengjianping (dismissed):
Good pr, but not sure it should be included with v3.4.2.
@Garandor

node/src/client.rs
```diff
@@ -0,0 +1,96 @@
...
+    + frame_rpc_system::AccountNonceApi<Block, AccountId, Nonce>
+    + cumulus_primitives_core::CollectCollationInfo<Block>
+    + sp_consensus_aura::AuraApi<Block, AuraId>
+    + nimbus_primitives::AuthorFilterAPI<Block, NimbusId>
+    + nimbus_primitives::NimbusApi<Block>
+{
+}
+
+/// Extend RuntimeApi trait bound for Aura
+pub trait RuntimeApiAura<AuraId: AppKey>:
```
Garandor: This trait isn't used anywhere in this PR?

zqhxuyuan: remove for now, previous I was trying change service_aura.rs, but it seems will be replace with nimubus, so yeah we don't need `RuntimeApiAura`.

Garandor:
for this PR L-changed is more appropriate. fixed is really only for bugs

node/src/client.rs
```diff
@@ -0,0 +1,69 @@
...
+        + sp_transaction_pool::runtime_api::TaggedTransactionQueue<Block>,
+    <Self as sp_api::ApiExt<Block>>::StateBackend: sp_api::StateBackend<BlakeTwo256>,
+{
+}
+
+impl<Api> RuntimeApiNimbus for Api where
+    Api: pallet_transaction_payment_rpc::TransactionPaymentRuntimeApi<Block, Balance>
+        + frame_rpc_system::AccountNonceApi<Block, AccountId, Nonce>
+        + cumulus_primitives_core::CollectCollationInfo<Block>
+        + sp_consensus_aura::AuraApi<Block, AuraId>
```
Garandor: technically `sp_consensus_aura` is not a NimbusAPI but instead left-over from Aura consensus (it's still needed though for old block verification on sync)
Since you removed the Aura one, i'd leave this one line outside of the wrapper types

Garandor: like you did in line 121
https://github.com/Manta-Network/Manta/pull/836/files#diff-594440db8adfc5d1b37711309312bfa06b49e488309186ad062a896778cf9fbfR121

Garandor (changes requested):
One comment, otherwise LGTM! :)

node/src/client.rs
zqhxuyuan: I'm also moving `TransactionPaymentRuntimeApi` and `AccountNonceApi` to commonn part, this will make `new_partial` more strict bound then before, But I think it's ok.
PS: can we also move `AuraApi` to common part too, in that case we dont' need add extra AuraApi in service.

ghzlatarev (approved):

ghzlatarev:
Looks good.
Please update the branch and finish the checklist.

node/src/client.rs
Garandor: > I'm also moving `TransactionPaymentRuntimeApi` and `AccountNonceApi` to commonn part, this will make `new_partial` more strict bound then before, But I think it's ok.

Sounds good 👍 

> PS: can we also move AuraApi to common part too, in that case we dont' need add extra AuraApi in service.

Manta won't have the aura+nimbus consensus when it launches so it won't need the Aura API, so IMO Aura shouldn't be in common

I am also thinking we can just remove `service_aura` with this PR. It was needed during the v3.2.1->v3.3.0 migration but it's dead code now

zqhxuyuan: > I am also thinking we can just remove `service_aura` with this PR. It was needed during the v3.2.1->v3.3.0 migration but it's dead code now

`service_aura` using cumulus-aura is used for manta-runtime now, however manta-runtime current not have Nimbus, So we must support Nimbus for manta-runtime, in that case we can remove `service_aura`. Am I right?

Garandor: oh right, that code hasn't been updated yet - it will be before launch, but for now it's needed to compile.
switching this over would take longer than today so leave `service_aura` in for now

Garandor (approved):
LGTM!

ghzlatarev (approved):

Issue #747, opened 2022-08-08
MantaPay 1.0.0


stechu:
The first production MantaPay version.

- [ ] Input parameter sanity checking for extrinsics ( > 0 )

Garandor:
resolved with [#903](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)

PR #866, opened 2022-10-19
Release v3.4.3 (second attempt)

Garandor:

Diff to last release: https://github.com/Manta-Network/Manta/compare/v3.4.2...release-v3.4.3

## Description

relates to #853

---

Situational Notes:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 ### Changed
 - [\#836](https://github.com/Manta-Network/Manta/pull/836) client trait bound refactor [CA]
 - [\#848](https://github.com/Manta-Network/Manta/pull/848) Fix XCM tests [CADO]
+- [\#860](https://github.com/Manta-Network/Manta/pull/860) Don't include testing helpers in release code [CA]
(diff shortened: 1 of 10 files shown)


ferrell-code:
Some of the crate versions are still 3.4.2 or 3.4.0. Not sure if we bump all of them, we don't necessarily have to sync all of them to runtime versions, but figured I would point it out

example: https://github.com/Manta-Network/Manta/blob/79b9036c2a96cf7c37e1110d8b94904ee9f251dd/primitives/manta/Cargo.toml#L8

Garandor:
> Some of the crate versions are still 3.4.2 or 3.4.0. Not sure if we bump all of them, we don't necessarily have to sync all of them to runtime versions, but figured I would point it out
> 
> example:
> 
> https://github.com/Manta-Network/Manta/blob/79b9036c2a96cf7c37e1110d8b94904ee9f251dd/primitives/manta/Cargo.toml#L8

Thanks for checking these, looks like I didn't catch all with search-and-replace.
I think I got em all now - except `manta-pay` which will stay on v0.4.0 until #747 is merged

ferrell-code (approved):

Dengjianping (approved):

PR #879, opened 2022-11-02
add tx-limit feature

### LINKED ISSUE #878: add tx-limit feature to limit to_private transfer
- Choice1: add a new pallet to set asset config
~~- Choice2: asset limit is kind of asset metadata, we can add a new field to asset-metadata~~
- intergation with pallet-manta-pay
  - ~~pallet-manta-pay use `FungibleLedger` trait, add another `ensure_valid(asset_id, amount)` to make sure tx is valid~~
    - amount is allowed when asset_id has limitation
    - or asset_id has no limitation.
  - add another trait and let pallet-manta-pay directly use TransactionLimit as implementation

zqhxuyuan:

## Description

relates to OR closes: #878

- This PR add another trait `TransactionLimitation` while current asset-id-migration has some refactor, so after asset-id-migration PR merged, this need change too.
- My initial thought is add another method into `FungibleLedger`, that can work too.

---

Situational Notes:


pallets/tx-limit/README.md
```diff
@@ -0,0 +1,3 @@
+Simple pallet to pause further executom of any extrinsic in the runtime. Provide pallet name and extrinsic name required to be paused. Use the name of the extrinsics as they are declared in the source code.
```
Garandor: forgot to change this

pallets/manta-pay/src/lib.rs
```diff
@@ -186,6 +190,17 @@ pub mod pallet {
         #[transactional]
         pub fn to_private(origin: OriginFor<T>, post: TransferPost) -> DispatchResultWithPostInfo {
             let origin = ensure_signed(origin)?;
+
+            let asset_id = post
+                .asset_id
+                .ok_or(Error::<T>::PublicUpdateInvalidAssetId)?;
+            for source in post.sources.iter() {
+                ensure!(
```
ferrell-code, line 204: Could be a few ways to rate limit. Right now a user could fill blocks with many `to_private` calls with the max allowed amount of asset. Maybe a Max Amount Allowed per Block so it only makes sense to do one transaction per block. 

Perhaps better to do in the future once #748 is merged. We can spell out the pros and cons to different rate limiting approaches

pallets/tx-limit/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+        #[pallet::weight(T::WeightInfo::set_asset_limit())]
+        #[transactional]
+        pub fn set_asset_limit(
+            origin: OriginFor<T>,
+            #[pallet::compact] asset_id: AssetId,
+            #[pallet::compact] amount: Balance,
+        ) -> DispatchResult {
+            T::UpdateOrigin::ensure_origin(origin)?;
+
+            AssetLimits::<T>::try_mutate(asset_id, |maybe_limit| -> DispatchResult {
```
ferrell-code: `AssetLimits::<T>::insert()`

Don't think we need to read the value, just overwrite it

pallets/manta-pay/src/lib.rs
Garandor: I'm also thinking that a per-extrinsic limit is too simplistic.
The idea is to prevent large amounts per block, not per TX. While we could just make the limit small per-tx so it adds up to the intended limit per block, it'll make the transaction limit unnecesarily small, annoying normal users who might be the only one attempting a 5k value transaction in the current block, which we'd have to make fail.

IMHO, `pallet-tx-limit` should
- have a storage map `LimitUsedThisBlock<asset_id,Balance>`
- subscribe to `to_private` events posted to the chain, read them including transferred balance (we might get Balance by derefing `source`)
- Increment the storage item by the balance from the event
- Query storage on every `to_private` the way it's currently implemented but compare `storage item + tx amount < limit`
- Reset storage on every `on_initialize`

pallets/manta-pay/src/lib.rs
```diff
@@ -186,6 +190,17 @@ pub mod pallet {
...
+                    T::TransactionLimit::ensure_valid(asset_id, *source),
+                    Error::<T>::InvalidSourceAccount
```
Garandor: I think this should have its own Error type e.g. `TransferExceedsBlockLimit`

Garandor (changes requested):
See posted comments, my proposal is just one idea, we can discuss more on Monday

pallets/manta-pay/src/lib.rs
zqhxuyuan: Actually, It's difficult to using polkadot.js to filling blocks with many `to_private`, because `TransferPost` is kind of black box for user that it's generated by signer. The first challenge when I trying use polkadot.js is how do I fill `utxo`, If I use my previous tx information that already finalized, that's also wrong, because `utxo` can't be same. so I think it's fine for per-extrinsic limit.

Garandor: If I understand your concern correctly, it's about pre-computing the extrinsic data for multiple `to_private` txns in sequence without waiting for them to be included on-chain.
This whole issue can be sidestepped by splitting assets into multiple wallets with multiple signers and private ledgers and then sending one TX from each of them

The total number of TXNs that can be executed in parallel is limited by the max block weight however, so the transaction number per-time is still limited, making a per-extrinsic limit workable.

pallets/manta-pay/src/lib.rs
```diff
@@ -186,6 +190,17 @@ pub mod pallet {
         #[transactional]
         pub fn to_private(origin: OriginFor<T>, post: TransferPost) -> DispatchResultWithPostInfo {
             let origin = ensure_signed(origin)?;
+
+            let asset_id = post
+                .asset_id
+                .ok_or(Error::<T>::PublicUpdateInvalidAssetId)?;
+            for source in post.sources.iter() {
+                ensure!(
+                    T::TransactionLimit::ensure_valid(asset_id, *source),
```
Dengjianping, line 205: Would `*source` be `0`? If it's 0 or not 0, and `asset_id` is not set any limit, this line of code always returns `false`.
```rust
amount < AssetLimits::<T>::get(asset_id) => false
```
The query type for `AssetLimits` is `ValueQuery`, so zero will be returned.
Not sure it will affect for those asset ids which don't have tx limit.

zqhxuyuan: current check: `!AssetLimits::<T>::contains_key(asset_id) || amount < AssetLimits::<T>::get(asset_id)`

- If asset limit is not set, contains_key = false, thus result is true.
- If asset limit is set, as long as `source` not exceed limit, even though `source` is 0, the right part return true.

PR #905, opened 2022-12-02
Update mantapay benchmarking to account for native asset-id operations

### LINKED ISSUE #899: Update benchmark macro of MantaPay
Split from https://github.com/Manta-Network/Manta/pull/890#discussion_r1019006832
Should run the benchmark with native asset-id as well, in case the weight for that code path is much worse.

ghzlatarev:

## Description

* Run benchmarks with both native and non native asset ids
* Had to modify the precomputed coin generation script and recompute the coins

closes: #899

---

Situational Notes:


ferrell-code (commented):
Add weights and the necessary new logic to the `#[pallet::weight()]` of the extrinsics?

Looks good to me in benchmark

ghzlatarev:
> Add weights and the necessary new logic to the `#[pallet::weight()]` of the extrinsics?
> 
> Looks good to me in benchmark

The weights are generated during release and are stored in `runtime/calamari/src/weights` folder

Dengjianping (dismissed):

pallets/manta-pay/src/benchmark/mod.rs
```diff
@@ -93,16 +95,34 @@ benchmarks! {
     }
 
     to_public {
+        let x in 0 .. 1;
         let caller: T::AccountId = whitelisted_caller();
         let origin = T::Origin::from(RawOrigin::Signed(caller.clone()));
+        let _ =<T::AssetConfig as AssetConfig<T>>::FungibleLedger::deposit_minting_with_check(1, &caller, INITIAL_VALUE, true);
```
zqhxuyuan: format: `let _ =<`

ghzlatarev: fixed.

zqhxuyuan (dismissed):

ferrell-code:
> The weights are generated during release and are stored in `runtime/calamari/src/weights` folder

Gotcha we will need to change these lines to something like this for example. `#[pallet::weights(cmp::max(to_public(0), to_public(1))]` 

We can do that during the release

ferrell-code (dismissed):

Dengjianping (dismissed):

ghzlatarev:
> > The weights are generated during release and are stored in `runtime/calamari/src/weights` folder
> 
> Gotcha we will need to change these lines to something like this for example. `#[pallet::weights(cmp::max(to_public(0), to_public(1))]`
> 
> We can do that during the release

Oh yeah I missed that. Actually, need to fix tests as well.

pallets/manta-pay/src/lib.rs
```diff
@@ -196,17 +196,38 @@ pub mod pallet {
         /// Transforms some public assets into private ones using `post`, withdrawing the public
         /// assets from the `origin` account.
         #[pallet::call_index(0)]
-        #[pallet::weight(T::WeightInfo::to_private())]
+        #[pallet::weight({
+            let _native_asset_id = <T::AssetConfig as AssetConfig<T>>::NativeAssetId::get();
+			match post.asset_id {
```
ferrell-code: weird whitespace thing?

ghzlatarev: fixed.

ferrell-code (approved):

runtime/calamari/src/weights/pallet_manta_pay.rs
```diff
@@ -45,27 +45,55 @@ use sp_std::marker::PhantomData;
...
+    }
     // Storage: Assets Asset (r:1 w:1)
     // Storage: Assets Account (r:2 w:2)
     // Storage: MantaPay UtxoSet (r:1 w:1)
     // Storage: MantaPay NullifierSetSize (r:1 w:0)
     // Storage: MantaPay ShardTrees (r:1 w:1)
     // Storage: MantaPay UtxoAccumulatorOutputs (r:0 w:1)
     // Storage: MantaPay Shards (r:0 w:1)
-    fn to_private() -> Weight {
-        (39_380_718_000 as Weight)
```
Garandor, line 65: why are each of the two branches 10bil faster than before? I'd expect at least one branch to be roughly the same as the old number. Same for the other extrinsics

pallets/manta-pay/src/weights.rs
```diff
@@ -33,73 +48,199 @@ use frame_system::Config;
...
+    fn to_public_native() -> Weight;
+    fn to_public_non_native() -> Weight;
+    fn private_transfer_native() -> Weight;
+    fn private_transfer_non_native() -> Weight;
+    fn public_transfer_native() -> Weight;
+    fn public_transfer_non_native() -> Weight;
 }
 
-/// Concrete Weight Functions
+/// Weights for pallet_manta_pay using the Substrate node and recommended hardware.pub struct SubstrateWeight<T>(PhantomData<T>);
```
Garandor, line 61: copy-paste prob

pallets/manta-pay/src/weights.rs
```diff
@@ -14,11 +14,26 @@
 // You should have received a copy of the GNU General Public License
 // along with Manta.  If not, see <http://www.gnu.org/licenses/>.
 
-//! Autogenerated weights
+//! Autogenerated weights for pallet_manta_pay
```
Garandor, line 17: this file is autogenerated, not sure it's a good idea to make manual edits here theyll be lost if we update this again for  whatever reason

pallets/manta-pay/src/benchmark/mod.rs
```diff
@@ -76,65 +76,139 @@ where
...
+        let caller: T::AccountId = whitelisted_caller();
+        let origin = T::Origin::from(RawOrigin::Signed(caller.clone()));
+        let _ = <T::AssetConfig as AssetConfig<T>>::FungibleLedger::deposit_minting_with_check(<T::AssetConfig as AssetConfig<T>>::NativeAssetId::get(), &caller, INITIAL_VALUE, true);
+        let _ = <T::AssetConfig as AssetConfig<T>>::FungibleLedger::deposit_minting_with_check(<T::AssetConfig as AssetConfig<T>>::NativeAssetId::get(), &Pallet::<T>::account_id(), INITIAL_VALUE, true);
+        let mint_post = TransferPost::decode(&mut &*TO_PRIVATE[0_usize]).unwrap();
+        let asset = mint_post.source(0).unwrap();
+    }: {
+        Pallet::<T>::to_private(origin.clone(), mint_post).unwrap();
+    } verify {
+        // FIXME: add balance checking
```
Garandor, line 92: fixme?

runtime/calamari/src/weights/pallet_manta_pay.rs
```diff
@@ -45,27 +45,55 @@ use sp_std::marker::PhantomData;
...
+    fn to_private_non_native() -> Weight {
+        (27_130_516_000 as Weight)
```
Garandor, line 80: if native and non-native assets aare just 200mil units apart, why even distinguish between them?
Basically why introduce all this extra complexity if we're only saving our users 0.7% of TX weight/fees?

ferrell-code: Differs by two reads and two writes (still pretty similar)

ghzlatarev: We don't have to.
This is just a sanity check at this point, that would catch a big divergence if it happens.
I don't mind just closing this though, for the sake of lower complexity.

Garandor: 1. If we wanted to keep this, it could be worthwhile to find out whether one can manually invoke the benchmarking function just to compare the outputs (making sure you compare worst-case execution times) of these two in a unit test and then discard the computed weight result.

2. Alternative would be to make it a simple unit test that runs one native and one non-native asset to_private/transfer/public and fails if their returned weights differ by too much.
This doesn't catch the worst-case execution time of both of these tho.

If 1. turns out to be too complex/not possible I would vote for lower complexity here.
The changes to precomputed coins can be merged anyway, right?

PR #904, opened 2022-12-01
Update to new circuits mantapay-v1

Apokalip:
## Description
Changing structs and fixing encodes decodes relative to manta-rs changes and sdk needed changes

##Encode/Decode changes, some of the parity_scale_codec encodes decodes have been removed from manta-rs as custom trait implemenations are a headache needing to use scale_codec errors and not doing more than serde's serialization/deserialization. Most of the struct params are Fields/Groups that can be represented by a byte array [u8; 32] easily.

---

Before we can approve this PR for merge, please make sure that **all** the following items have been checked off:
- [x] Linked to Github issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [x] Added **one** label out of the `L-` group to this PR
- [x] Added **one or more** labels from the `A-` and `C-` groups to this PR
- [x] Explicitly labelled `A-calamari`, `A-dolphin` and/or `A-manta` if your changes are meant for/impact either of these (CI depends on it)
- [x] This PR is targeted against the *current*  Milestone ( otherwise discuss if it can be added in time)
- [ ] Re-reviewed `Files changed` in the Github PR explorer.

Situational Notes:
- If adding functionality, write unit tests!
- If importing a new pallet, choose a proper module index for it, and allow it in `BaseFilter`. Ensure **every** extrinsic works from front-end. If there's corresponding tool, ensure both work for each other.
- If needed, update our Javascript/Typescript APIs. These APIs are officially used by exchanges or community developers.
- If modifying existing runtime storage items, make sure to implement storage migrations for the runtime and test them with `try-runtime`. This includes migrations inherited from upstream changes, and you can search the diffs for modifications of `#[pallet::storage]` items to check for any.
- If runtime changes, need to update the version numbers properly:
   * `authoring_version`: The version of the authorship interface. An authoring node will not attempt to author blocks unless this is equal to its native runtime.
   * `spec_version`: The version of the runtime specification. A full node will not attempt to use its native runtime in substitute for the on-chain Wasm runtime unless all of spec_name, spec_version, and authoring_version are the same between Wasm and native.
   * `impl_version`: The version of the implementation of the specification. Nodes are free to ignore this; it serves only as an indication that the code is different; as long as the other two versions are the same then while the actual code may be different, it is nonetheless required to do the same thing. Non-consensus-breaking optimizations are about the only changes that could be made which would result in only the impl_version changing.
[...]

PR #911, opened 2022-12-07
Feature/manta rs v0.5.8

### LINKED ISSUE #812: Integrate MantaPay V1 changes into production/mock runtimes

zqhxuyuan:
## Description

relates to OR closes: #XXXX

pick commits from https://github.com/Manta-Network/Manta/pull/904 to avoid DCO failed:

---

Situational Notes:


Cargo.lock
```diff
@@ -1711,12 +1742,12 @@ dependencies = [
 [[package]]
 name = "cumulus-client-consensus-common"
 version = "0.1.0"
-source = "git+https://github.com/paritytech/cumulus.git?branch=polkadot-v0.9.26#e43b8b878a6fd0ca8b5e88d19822c4d777d3c677"
+source = "git+https://github.com/paritytech/cumulus.git?branch=polkadot-v0.9.26#be9e23c377555cabb867326ace51e0ab72bee1b9"
```
Garandor, line 1745: looks like there was an implicit upgrade
from: [Use genesis_state_version param instead of StateVersion::V0 (](https://github.com/paritytech/cumulus/commit/e43b8b878a6fd0ca8b5e88d19822c4d777d3c677)
via: [update substrate](https://github.com/paritytech/cumulus/commit/9138c30cdf53c9e285d2871575577e1edb79cf3f)
to: [Ensure that inherents are first and unsigned (](https://github.com/paritytech/cumulus/commit/be9e23c377555cabb867326ace51e0ab72bee1b9)

which contains a new substrate commit
https://github.com/paritytech/cumulus/commit/9138c30cdf53c9e285d2871575577e1edb79cf3f#diff-13ee4b2252c9e516a0547f2891aa2105c3ca71c6d7a1e682c69be97998dfc87eL435

which we dont have in our own cargo.lock
https://github.com/Manta-Network/Manta/blob/feature/manta-rs-v0.5.8/Cargo.lock#L653

zqhxuyuan: Do we need revert to old commit?

Apokalip: @Garandor Should we revert the commit or just manually overwrite?

Garandor: cleared up in discussion. Keeping the new version with only `tracing-core` reduced to v 1.26

ghzlatarev (approved):

Cargo.lock
Garandor: Will follow up investigating impact of dependency changes in the near future.

Special importance to the upgrade of `chrono` which pulls in a lot of new deps

pallets/manta-pay/src/lib.rs
```diff
@@ -759,7 +766,13 @@ where
         &self,
         output: config::UtxoAccumulatorOutput,
     ) -> Option<Self::ValidUtxoAccumulatorOutput> {
-        if UtxoAccumulatorOutputs::<T>::contains_key(encode(output)) {
+        let accumulator_output = fp_encode(output).ok()?;
+        // Checking for an empty(zeroed) byte array.
+        // This happens for UTXOs with value = 0, for which you dont need
+        // a membership proof, but you still need a root(in this case zeroed).
```
Garandor, line 772: I'd like to see a `why` (because zerocoins are created on the fly?) instead of a `what` description here or a link to further documentation

Garandor: However, this is minor and can be added as part of #910

Garandor (approved):

PR #924, opened 2022-12-14
Fix runtime upgrade test by using governance instead of SUDO

### LINKED ISSUE #323: Integrate governance process into the CI runtime upgrade test
Currently the sudo (alice) account is used directly, which is not what happens on the actual chain.
The root account should be used through governance. This means.
Need to use `externalProposeDefault` as it is the only unfiltered call currently.
Use this workflow for reference - https://hackmd.io/0XKjhGxNQS-wcNb493bPIg?both

ghzlatarev:
## Description

closes: #323 

* Change trigger condition of the CI test, before it wasn't getting triggered on tags. Also the if condition for one of the sub-jobs was only being triggered for dolphin label.

* Use governance instead of sudo account to submit runtime upgrade extrinsics. The old method could only upgrade a special ci chain-spec which included sudo in its genesis. That special chain-spec is now removed alltogether and we can trigger with `calamari-dev` chain-spec.
* Migrate the test JS/TS code from Dev-Tools repo to tests folder in this repo
* Use mocha framework to assert the test conditions (that spec_version has increased)
* Update the CI test to use the new mocha test.
* Successful run: https://github.com/Manta-Network/Manta/actions/runs/3720978998

---

Situational Notes:


ferrell-code (dismissed):

Garandor (approved):
LGTM, great PR!

ferrell-code (approved):

tests/runtime_upgrade_test.ts
```diff
@@ -0,0 +1,62 @@
...
+        const oldRuntimeVersion = await api.rpc.state.getRuntimeVersion();
+        const oldSpecVersion = oldRuntimeVersion["specVersion"];
+
+        const code = fs.readFileSync('calamari.wasm').toString('hex');
+        let codeHash = blake2AsHex(`0x${code}`);
+        const authorizeUpgradeCallData = api.tx.parachainSystem.authorizeUpgrade(codeHash);
+        var referendumIndexObject = { referendumIndex: 0 };
+        execute_with_root_via_governance(api, aliceKeyPair, authorizeUpgradeCallData, referendumIndexObject);
+        await delay(60000);
+        api.tx.parachainSystem.enactAuthorizedUpgrade(`0x${code}`).signAndSend(aliceKeyPair, {nonce: -1});
```
Dengjianping, line 46: `enactAuthorizedUpgrade` can be signed or unsigned, anyway, it's fine.

PR #928, opened 2022-12-19
MantaPay stress-test benchmark

### LINKED ISSUE #861: Stress test for MantaPay pallet
- [x] generate lots of storage data
- [x] input storage data into a running network
- [x] perform some test transactions
- [x] generate chain-spec with all the data from genesis

ghzlatarev:
## Description

closes: #861 

* The idea of the stress test is to send lots of transactions on top of a large ledger, and making sure they succeed in a certain amount of time.
* Specifically the amount of transactions is 700 in the form `to-private -> to-private -> to-private -> private-transfer -> to-private -> to-private -> to-public`. 
* The size of the ledger is ~100k transactions of the same form, which populate all the storage items of the pallet.
* The asserts are that transactions per second from start to last finalized transaction is more than 0.5. Also that all expected success events are recorded on the ledger.
* The test runs on the reference hardware that we also use for benchmarking weights

---

Situational Notes:


.github/workflows/integration_test_dolphin.yml
```diff
@@ -256,17 +269,15 @@ jobs:
...
             -- \
             --address=ws://127.0.0.1:9921
-      - uses: actions/checkout@v2
-        with:
-          repository: Manta-Network/Manta
-          path: Manta
       - name: run test suites
         run: |
-          # TODO: implement moonbeam-like js test suite triggers here
           sleep 120
```
ghzlatarev, line 274: todo: update to 240

.github/workflows/integration_test_calamari.yml
```diff
@@ -1,7 +1,7 @@
 name: calamari integration test
 on:
   push:
-    branches: [manta]
+    branches: [ghzlatarev/stress-benchmark-test]
```
ghzlatarev: todo: use reference hardware

tests/append_storage.js
```diff
@@ -0,0 +1,23 @@
+const fs = require('fs');
+
+let prefixes = [
+'0x682a59d51ab9e48a8c8cc418ff9708d2',
+'0x4ae7e256f92e5888372d72f3e4db1003',
+'0xa66d1aecfdbd14d785a4d1d8723b4beb'
+];
+
+async function main() {
+  let storage = JSON.parse(fs.readFileSync('./storage.json', 'utf8'));
```
ghzlatarev: duplicated variable

.github/workflows/integration_test_calamari.yml
```diff
@@ -370,6 +377,39 @@ jobs:
         name: test - calamari alice peered successfully
         run: |
           grep '\[Parachain\] 💤 Idle (${{ matrix.chain-spec.expected.peer-count.para }} peers)' ${{ github.workspace }}/polkadot-launch/9921.log
+      - name: append manta-pay storage
```
ghzlatarev, line 373: add comment why launching separately

tests/stress_benchmark_test.ts
```diff
@@ -0,0 +1,287 @@
...
+    });
+    const keyring = new Keyring({ type: "sr25519" });
+    const sender = keyring.addFromMnemonic(test_config.mnemonic);
+
+    const mintsContent = await readFile("./data/precomputed_mints");
+    const mintsBuffer = mintsContent.subarray(
+      test_config.mints_offset,
+      test_config.mint_size * test_config.total_iterations
+    );
+    let full_transfer_size =
```
ferrell-code: Nit:
```suggestion
    const fullTransferSize =
```

tests/stress_benchmark_test.ts
```diff
@@ -0,0 +1,287 @@
...
+      test_config.mint_size * 2 + test_config.transfer_size;
+    const transfersContent = await readFile("./data/precomputed_transfers");
+    const transfersBuffer = transfersContent.subarray(
+      test_config.transfers_offset,
+      full_transfer_size * test_config.total_iterations
+    );
+    let fullReclaimSize = test_config.mint_size * 2 + test_config.reclaim_size;
```
ferrell-code:
```suggestion
    const fullReclaimSize = test_config.mint_size * 2 + test_config.reclaim_size;
```

tests/append_storage.js
```diff
@@ -0,0 +1,26 @@
+const fs = require('fs');
+
+let prefixes = [
```
ferrell-code, line 4: maybe add comment saying what they are

```// manta-pay, asset-manager and assets storage items```

**Review by ferrell-code — COMMENTED** [review]
Some nits

**Review by ferrell-code — DISMISSED** [review]

File: .gitattributes [hunk 8]
```diff
@@ -0,0 +1 @@
+tests/data/storage.json filter=lfs diff=lfs merge=lfs -text
```
Garandor: I'm not sure lfs is a good way to store this.
If we get multiple unrelated (large) data stored like this, your stress test will have to `git lfs init` the whole repo before it can access its own blob.

would it be possible to store this on a S3 bucket instead?

I've used `git lfs` extensively in the past and i haven't seen an instance and I have yet to see an instance where it didn't eventually lead to problems

`lfs` is good for exactly one thing IMO, and that's if your codebase depends on large blobs (image files, UI assets) for building so that you keep the repo smal, can edit sources without DLing the deps and can in a single action DL all needed deps if you need to build.

This doesn't fit into that usecase so I think it'd be better stored elsewhere

.github/workflows/integration_test_dolphin.yml
```diff
@@ -370,6 +373,46 @@ jobs:
...
+            -- $HOME/.local/share/calamari-pc/${{ matrix.chain-spec.id }}-${GITHUB_SHA:0:7}-launch-config.json
+      - name: run stress test
+        run: |
+          sleep 720
+          cd ${{ github.workspace }}/Manta/tests
+          yarn install
+          yarn
+          yarn stress_benchmark_test --address=ws://127.0.0.1:9921 --exit
+          echo $?
+          if [ $? != 0 ]; then echo "Stress benchmark test failed!"; exit 1; fi
```
Garandor: this probably doesnt work.
`echo $?` will set a 0 exit code and so :406 will never trigger.
assign $? to a variable instead

tests/stress_benchmark_test.ts
```diff
@@ -0,0 +1,287 @@
...
+          )
+          .signAndSend(sender, { nonce: -1 }, ({ events = [], status }) => {
+            if (status.isInBlock) {
+              console.log("Included at block hash", status.asInBlock.toHex());
+              console.log("Events:");
+              events.forEach(({ event: { data, method, section }, phase }) => {
+                let event = section + "." + method;
+                console.log("event: ", event);
+                if ("mantaPay.ToPublic" == event) {
+                  allSuccesses++;
```
Garandor, line 250: why not unsub here and at all the others except line 203 where youre waiting for finalization?

tests/stress_benchmark_test.ts
```diff
@@ -0,0 +1,287 @@
...
+        let tps = (test_config.tests_iterations * 7) / totalTime;
+        console.log("Tps is: ", tps);
+        assert(tps >= test_config.expected_tps);
+        break;
+      }
+    }
+
+    if (lastFinalized == false) {
+      assert(false);
+    }
```
Garandor:
```suggestion
assert(lastFinalized);
```

Garandor (commented):
main comment on avoiding `git lfs`, looks good otherwise

.gitattributes
ghzlatarev: Yeah makes sense, can put it on S3

.github/workflows/integration_test_dolphin.yml
ghzlatarev: Nice catch. Accidentally the tests still worked since the exit code from the js code would terminate the CI job...

tests/stress_benchmark_test.ts
ghzlatarev: No reason , not even sure what the benefit is so ill just remove it

Garandor (approved):

ferrell-code (approved):

Issue #926, opened 2022-12-16
Create Manta Parachain Genesis


Garandor:
- [x] Initial collator session keys
- [x] Initial collator accounts
- [x] Sudo with controllers
- [x] initial balance 1_000_000_000 MA
- [x] Parachain ID

PR #946, opened 2022-12-26
Update Manta Runtime

### LINKED ISSUE #925: Update Manta Runtime
- [x] Enable SUDO, nimbus, parachain-staking pallets
- [x] Disable the rest

Garandor:

## Description

- `SessionKeys::new(unchecked_collator_keys("Dave"))` => `SessionKeys::from_seed_unchecked("Dave")` and adapted chainspec files. This is better as it allows us to have Manta RT have only 2 SessionKeys from genesis, omitting the obsolete Aura key
- Removed all nonessential pallets from manta RT except
  - sudo (which is a multisig)
  - staking (initial collators)
  - nimbus (block production)
  -  Utility (handling operations in batches to make multisig sudo operation less painful)
  - Balances (preparation for token distribution and to make sure other accounts can fill up multisig controllers if need be)
  - Multisig (managing sudo)
  - DMP (saving the para from a SNAFU through relaychain governance)
  - preimage (to store future runtime upgrades)
- node no longer uses `service_aura` for Manta RT
- Set staking inflation to 0
- filter all staking extrinsics ( sudo only )

relates to OR closes: #925 

---

Situational Notes:


node/src/chain_specs/manta.rs
```diff
@@ -157,32 +167,42 @@ fn manta_dev_genesis(
...
+                        account,
+                        manta_runtime::staking::NORMAL_COLLATOR_MINIMUM_STAKE,
+                    )
+                })
+                .collect(),
+            delegations,
+            inflation_config: manta_runtime::staking::inflation_config::<manta_runtime::Runtime>(),
+        },
         parachain_info: manta_runtime::ParachainInfoConfig {
             parachain_id: MANTA_PARACHAIN_ID.into(),
```
ghzlatarev, line 185: The parachain-id needs to change, as this one is already in use.

runtime/manta/src/lib.rs
```diff
@@ -631,28 +487,23 @@ construct_runtime!(
...
-        Scheduler: pallet_scheduler::{Pallet, Call, Storage, Event<T>} = 29,
-
-        // XCM helpers.
-        XcmpQueue: cumulus_pallet_xcmp_queue::{Pallet, Call, Storage, Event<T>} = 30,
-        PolkadotXcm: pallet_xcm::{Pallet, Call, Storage, Event<T>, Origin, Config} = 31,
-        CumulusXcm: cumulus_pallet_xcm::{Pallet, Event<T>, Origin} = 32,
-        DmpQueue: cumulus_pallet_dmp_queue::{Pallet, Call, Storage, Event<T>} = 33,
 
         // Handy utilities.
         Utility: pallet_utility::{Pallet, Call, Event} = 40,
```
ghzlatarev: Multisig for the sudo i guess, but what is Utility needed for ?

runtime/manta/src/lib.rs
```diff
@@ -631,28 +487,23 @@ construct_runtime!(
...
         // Preimage registry.
         Preimage: pallet_preimage::{Pallet, Call, Storage, Event<T>} = 28,
-        // System scheduler
-        Scheduler: pallet_scheduler::{Pallet, Call, Storage, Event<T>} = 29,
-
-        // XCM helpers.
-        XcmpQueue: cumulus_pallet_xcmp_queue::{Pallet, Call, Storage, Event<T>} = 30,
-        PolkadotXcm: pallet_xcm::{Pallet, Call, Storage, Event<T>, Origin, Config} = 31,
-        CumulusXcm: cumulus_pallet_xcm::{Pallet, Event<T>, Origin} = 32,
-        DmpQueue: cumulus_pallet_dmp_queue::{Pallet, Call, Storage, Event<T>} = 33,
```
ghzlatarev: Maybe we should keep the dmp_queue in case the relay chain needs to send us a message, ie: some emergency upgrade that only it can trigger.

runtime/manta/src/lib.rs
ghzlatarev: Btw if we're using multisig for the sudo, we need to make sure that all of the related accounts (and the multisig itself) are funded in the genesis.

node/src/chain_specs/manta.rs
```diff
@@ -157,32 +167,42 @@ fn manta_dev_genesis(
...
         },
         collator_selection: manta_runtime::CollatorSelectionConfig {
```
Dengjianping, line 187: For brand new runtime, do we still need `collator_selection`?
I'm curious.

runtime/manta/Cargo.toml
```diff
@@ -43,7 +45,6 @@ pallet-authorship = { git = 'https://github.com/paritytech/substrate.git', defau
 pallet-balances = { git = 'https://github.com/paritytech/substrate.git', default-features = false, branch = "polkadot-v0.9.26" }
 pallet-multisig = { git = 'https://github.com/paritytech/substrate.git', default-features = false, branch = "polkadot-v0.9.26" }
```
Dengjianping, line 46: Seems ~~`pallet-multisig`~~, `pallet-utility` and `pallet-preimage` can be removed. But not 100% sure.

runtime/manta/src/lib.rs
```diff
@@ -203,32 +184,22 @@ impl Contains<Call> for MantaFilter {
...
+            // Explicitly DISALLOWED calls ( Pallet user extrinsics we don't want used WITH REASONING )
             // Explicitly ALLOWED calls
             | Call::Authorship(_)
             // Sudo also cannot be filtered because it is used in runtime upgrade.
             | Call::Sudo(_)
             | Call::Multisig(_)
-            | Call::TransactionPause(_)
-            | Call::Balances(_) => true,
+            | Call::AuthorInherent(pallet_author_inherent::Call::kick_off_authorship_validation {..}) // executes unsigned on every block
+            | Call::Balances(_)
```
Dengjianping, line 198: `Balances` is allowed?

runtime/manta/src/lib.rs
Garandor: Thanks for pointing this out, i would've only cared about the controller accounts otherwise.

> Multisig for the sudo i guess, but what is Utility needed for ?

Multisig sudo is annoying to operate, with batch calls we can bundle operations and sign them once

runtime/manta/Cargo.toml
Garandor: won't we need `pallet-preimage` to do runtime upgrades through the `authorize_upgrade` => `enact_authorized_upgrade` flow?

For utility reasoning see https://github.com/Manta-Network/Manta/pull/946#discussion_r1057814981

runtime/manta/src/lib.rs
Garandor: I suppose we don't need to do it now, but this way we won't need another runtime upgrade after distributing tokens to allow users to transfer them.

I don't see a downside to having it enabled since nobody will have a balance to transfer anyway

node/src/chain_specs/manta.rs
Garandor: Yes we do unfortunately, the pallet is still used as a nimbus filter stage
https://github.com/Manta-Network/Manta/blob/3c1e15d9100ba20e31ce8c3ccc0e2dd1054471b7/runtime/manta/src/lib.rs#L473

We do not need to give it any configuration tho

runtime/manta/src/lib.rs
Garandor: @ghzlatarev can the `cumulus_pallet_dmp_queue` be used standalone without PolkadotXcm/CumulusXcm?

Garandor:
xcm_config is missing from your commit @ghzlatarev 

i'll try adding it

node/src/chain_specs/manta.rs
Garandor: chain spec changes will go into followup PR #926

node/src/chain_specs/manta.rs
```diff
@@ -157,30 +167,43 @@ fn manta_dev_genesis(
...
+                        acc,          // validator id
+                        session_keys, // collator session keys
                     )
                 })
                 .collect(),
         },
-        aura_ext: Default::default(),
         parachain_system: Default::default(),
-        polkadot_xcm: manta_runtime::PolkadotXcmConfig {
+        polkadot_xcm: calamari_runtime::PolkadotXcmConfig {
```
ferrell-code, line 206: use manta_runtime?

Garandor: thanks, this was a copy-pasta mistake. I fixed it in #952

ghzlatarev (approved):
LGTM, but please update the description to mention the reasoning for Utility, Balances, Multisig, DMP

ferrell-code (approved):

flame4 (approved):

PR #948, opened 2022-12-27
add zero balance check

### LINKED ISSUE #880: Add parameter sanity checks to `mantapay` public extrinsics
> Yes, to_private and to_public should add other kinds of checks like > 0 and < L for asset values

### LINKED ISSUE #947: Split `>0` checks from pallet-tx-limit

zqhxuyuan:

## Description

relates to OR closes: #947

---

Situational Notes:

### DIFF
--- a/pallets/manta-pay/src/lib.rs
+++ b/pallets/manta-pay/src/lib.rs
@@ -201,12 +201,6 @@ pub mod pallet {
         #[transactional]
         pub fn to_private(origin: OriginFor<T>, post: TransferPost) -> DispatchResultWithPostInfo {
             let origin = ensure_signed(origin)?;
-            for source in post.sources.iter() {
-                ensure!(
-                    asset_value_decode(*source) > 0u128,
-                    Error::<T>::ZeroTransfer
-                );
-            }
             ensure!(
                 post.sources.len() == 1
                     && post.sender_posts.len() == 0
@@ -214,6 +208,12 @@ pub mod pallet {
                     && post.sinks.len() == 0,
                 Error::<T>::InvalidShape
             );
+            for source in post.sources.iter() {
+                ensure!(
+                    asset_value_decode(*source) > 0u128,
+                    Error::<T>::ZeroTransfer
+                );
+            }
             Self::post_transaction(None, vec![origin], vec![], post)
         }
 
@@ -224,16 +224,16 @@ pub mod pallet {
         #[transactional]
         pub fn to_public(origin: OriginFor<T>, post: TransferPost) -> DispatchResultWithPostInfo {
             let origin = ensure_signed(origin)?;
-            for sink in post.sinks.iter() {
-                ensure!(asset_value_decode(*sink) > 0u128, Error::<T>::ZeroTransfer);
-            }
             ensure!(
                 post.sources.len() == 0
                     && post.sender_posts.len() == 2
-                    && post.receiver_posts.len() == 0
+                    && post.receiver_posts.len() == 1
                     && post.sinks.len() == 1,
                 Error::<T>::InvalidShape
             );
+            for sink in post.sinks.iter() {
+                ensure!(asset_value_decode(*sink) > 0u128, Error::<T>::ZeroTransfer);
+            }
             Self::post_transaction(None, vec![], vec![origin], post)
         }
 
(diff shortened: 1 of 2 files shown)


ghzlatarev (dismissed):

pallets/manta-pay/src/lib.rs
```diff
@@ -201,6 +201,9 @@ pub mod pallet {
         #[transactional]
         pub fn to_private(origin: OriginFor<T>, post: TransferPost) -> DispatchResultWithPostInfo {
             let origin = ensure_signed(origin)?;
+            for source in post.sources.iter() {
+                ensure!(asset_value_decode(*source) > 0u128, Error::<T>::BalanceLow);
+            }
```
ghzlatarev: the front-end doesn't do 0 value txs anymore, right ?

zqhxuyuan: The frontend UI has 0 value checking, but sdk don't have this check now.

Garandor: Yeah, IMHO it's best to add this simple check on every level of the stack, so @zqhxuyuan pls make a note/issue to add it to SDK also

pallets/manta-pay/src/lib.rs
```diff
@@ -201,6 +201,9 @@ pub mod pallet {
         #[transactional]
         pub fn to_private(origin: OriginFor<T>, post: TransferPost) -> DispatchResultWithPostInfo {
             let origin = ensure_signed(origin)?;
+            for source in post.sources.iter() {
+                ensure!(asset_value_decode(*source) > 0u128, Error::<T>::BalanceLow);
```
Garandor: I think these errors would be better represented through `ZeroTransfer`

Garandor: Also, are you sure that *each* source&sink must be nonzero?
Isn't it suffient for *at-least-one* source to be?

Garandor: I pushed a change for this, pls resolve if you agree

Garandor:
It would also make sense to add this same check to the `public_transfer` extrinsic

pallets/manta-pay/src/lib.rs
zqhxuyuan: we can change 0 to exist deposit of each token. ~~amount less than ED is also invalid.~~

Garandor: not sure if existential deposit makes sense here.

Our private ledger doesn't have notion of ED so I can only see it on the `to_public` call.
But a sink for `to_public` must already exist, otherwise you can't pay the gas fees for the `to_public`, so you *can always* `to_public` amounts smaller than ED without risk of losing it.

Am I missing something?

zqhxuyuan: > Also, are you sure that _each_ source&sink must be nonzero? Isn't it suffient for _at-least-one_ source to be?

Ok, you're saying that maybe we don't loop all sources? In that case, the sources len will be only 1. I added another check that we discuss before.

zqhxuyuan:
> It would also make sense to add this same check to the `public_transfer` extrinsic

I don't think we need this check, as public transfer not touch our private ledger. it's ok for user just transfer 0 amount and cost some fee.

pallets/manta-pay/src/lib.rs
```diff
@@ -202,8 +202,18 @@ pub mod pallet {
...
             let origin = ensure_signed(origin)?;
             for source in post.sources.iter() {
-                ensure!(asset_value_decode(*source) > 0u128, Error::<T>::ZeroTransfer);
+                ensure!(
+                    asset_value_decode(*source) > 0u128,
+                    Error::<T>::ZeroTransfer
+                );
             }
+            ensure!(
+                post.sources.len() == 1
```
ghzlatarev: Isn't 2 in and 2 out is also valid though ?

zqhxuyuan: in the previous version, when private transfer need one zero coin to fillout 2 in, it'll also invoke to_private extrinc, but current version already get rid this part. so to_private is only for public token to private token. and it only has 1 sources, 1 receiver posts. For private transfer, it have 2 sender posts and 2 receiver posts.

ghzlatarev: Ah yeah, saw it here as well https://github.com/Manta-Network/manta-rs/blob/6572219b4793a5fed9ae4c8ecc1281eadbc038a4/manta-accounting/src/transfer/canonical.rs#L95

ghzlatarev (approved):

Garandor (approved):

PR #952, opened 2022-12-28
update genesis

### LINKED ISSUE #926: Create Manta Parachain Genesis

Garandor:
## Description

relates to OR closes: #926 

---

Situational Notes:


node/src/chain_specs/manta.rs
```diff
@@ -157,30 +173,38 @@ fn manta_dev_genesis(
...
+                        acc,          // validator id
+                        session_keys, // collator session keys
                     )
                 })
                 .collect(),
         },
-        aura_ext: Default::default(),
         parachain_system: Default::default(),
-        polkadot_xcm: manta_runtime::PolkadotXcmConfig {
+        polkadot_xcm: calamari_runtime::PolkadotXcmConfig {
```
ferrell-code: manta_runtime?

node/src/chain_specs/manta.rs
```diff
@@ -198,4 +222,198 @@ pub fn manta_testnet_config() -> Result<MantaChainSpec, String> {
...
+                        (
+                            collator.acc.clone(),  // account id
+                            collator.acc.clone(),  // validator id
+                            collator.keys.clone(), // collator session keys
+                        )
+                    })
+                    .collect(),
+            },
+            parachain_system: Default::default(),
+            polkadot_xcm: calamari_runtime::PolkadotXcmConfig {
```
ferrell-code: manta_runtime?

node/src/chain_specs/manta.rs
```diff
@@ -198,4 +222,198 @@ pub fn manta_testnet_config() -> Result<MantaChainSpec, String> {
...
+            (
+                hex!("3e31eaf98e0f6dd9b088c095757fac00dc9e3fe6af29e42caea94f7d76dd3444").into(),
+                INITIAL_CONTROLLER_BALANCE,
+            ), // CONTROLLER 3 ss58 12QYoFyrrEomSq4tMPxVxnWvPwVVKQHUpDWZuDxcg5Gv3c9E
+            (
+                hex!("0619e4018531059b677d70b476b72431d3e88fe71d2c27bba5d945975e824956").into(),
+                INITIAL_CONTROLLER_BALANCE,
+            ), // COLLATOR 1
+            (
+                hex!("0202d1760e33f60f977fcff88b611cb95e66d0b3b74c12ac6c8037b8b9764a31").into(),
```
Garandor, line 270: @grenade pls doublecheck the collator addresses match with what you gave me

node/src/chain_specs/manta.rs
```diff
@@ -198,4 +222,198 @@ pub fn manta_testnet_config() -> Result<MantaChainSpec, String> {
...
+            parachain_system: Default::default(),
+            polkadot_xcm: calamari_runtime::PolkadotXcmConfig {
+                safe_xcm_version: Some(SAFE_XCM_VERSION),
+            },
+        }
+    }
+
+    pub fn manta_mainnet_config() -> MantaChainSpec {
+        let genesis_collators: Vec<Collator> = vec![
+        Collator::new( // c1: dfWh1oFWNxrmrHdiGdVpQQD6Pp3SmY5NrNsAGNRCxgaqJuZif
```
Garandor, line 349: @grenade pls doublecheck that collator addresses, session keys and nodeID match with what you gave me

node/src/chain_specs/manta.rs
Garandor: thanks, fixed!

node/src/chain_specs/manta.rs
Garandor: thanks, fixed!

flame4 (dismissed):

node/src/chain_specs/manta.rs
```diff
@@ -198,4 +222,203 @@ pub fn manta_testnet_config() -> Result<MantaChainSpec, String> {
 /// Returns the Manta mainnet chainspec.
 pub fn manta_config() -> Result<MantaChainSpec, String> {
     MantaChainSpec::from_json_bytes(&include_bytes!("../../../genesis/manta-genesis.json")[..])
+    // Ok(mainnet::manta_mainnet_config())
```
ferrell-code, line 225: Remove commented line?

Garandor: this is a bit of a topic for discussion, i want to preserve the code that generated the genesis json.
If possible, i'd actually want to keep 225 active and delete 224.
tbh I don't even know the reason why we do this file read in the first place.

@ghzlatarev do you know the history?

I can only imagine that the `mod mainnet` code will eventually break compilation if we don't keep it updated, so this file preserves the record of mainnet as it was used to spawn the network even when we have to comment out/remove the code that generates this file

ghzlatarev: When we make it into a file we can use it with any version of the node and it should start from the original runtime.
If it's in-place it would start with the latest runtime version.
I think that's the reason.
Also maybe obfuscating some account ids.

Garandor: makes sense, the RT blob gets built into the file.

This actually means we can only generate the config after bumping spec versions, otherwise it'll report as `manta/4000`

I'm wondering how best to preserve this code tho, maybe splitting the module into its own file?
I don't think we need to obfuscate any of these pubkeys

ferrell-code: I think its because the genesis file must be etched into stone can never be changed once network is live. 

Aka if we change the genesis than the node can't sync to manta network as it has a different genesis block

I think leaving this comment here is good thing, as we may want to change the genesis before launching manta network

ferrell-code (approved):

node/src/chain_specs/manta.rs
Garandor: the line will be needed in #957 and we can use that issue to figure out where to put the code that generated it.

node/src/chain_specs/manta.rs
grenade: all verified (c1 - c5, node, collator, nimbus, vrf).

node/src/chain_specs/manta.rs
grenade: all verified (c1 - c5, node, collator, nimbus, vrf).

grenade: all verified (c1 - c5, node, collator, nimbus, vrf).

node/src/chain_specs/manta.rs
grenade: all verified (c1 - c5, node, collator, nimbus, vrf).

grenade (approved):
all verified (c1 - c5, node, collator, nimbus, vrf).

node/src/chain_specs/manta.rs
```diff
@@ -221,4 +222,203 @@ pub fn manta_testnet_config() -> Result<MantaChainSpec, String> {
...
+            SessionKeys::new((
+                hex!("1eda33797b226fed66ea3ac327a77f26b594b90bc9e9af7416a651bef07a7c72").unchecked_into(), // ss58: dfXFU5LYyQj5vzHeaHbVfhyenuibJJPGSpPonKCPNw6qsUwSm
+                hex!("bec2a422e8b72b6f5423eb2ec59d9727e580227e0e0327247d516593ec7b5312").unchecked_into(), // ss58: dfas8jnCTfPXtVvwAu15PtX6WAc3rw4iacCHeuj7qniepWtJE
+            ))
+        ),
+    ];
+        let genesis_collators_clone = genesis_collators.clone(); // so we can move it into the constructor closure
+
+        MantaChainSpec::from_genesis(
+            "Shark Parachain",
```
zqhxuyuan, line 393: rename to manta parachain? or the manta symbol is Shark?

node/src/chain_specs/manta.rs
```diff
@@ -221,4 +222,203 @@ pub fn manta_testnet_config() -> Result<MantaChainSpec, String> {
...
+                hex!("7a5cef43508640b6839c37a41272dec83dda4df28503bd87436c197d213f692c").into(),
+                INITIAL_COLLATOR_BALANCE,
+            ), // COLLATOR 4
+            (
+                hex!("4039a894994d326e32801a0fc19a267383b6231e74bc9c3fd1021d949d849406").into(),
+                INITIAL_COLLATOR_BALANCE,
+            ), // COLLATOR 5
+            (
+                root_key.clone(),
+                TOTAL_ISSUANCE - 4 * INITIAL_CONTROLLER_BALANCE - 4 * INITIAL_COLLATOR_BALANCE,
```
zqhxuyuan, line 287: is this calculation right? we have 3 controller, and 5 collator? or 4 controller and 4 collator?

node/src/chain_specs/manta.rs
```diff
@@ -221,4 +222,203 @@ pub fn manta_testnet_config() -> Result<MantaChainSpec, String> {
...
+                INITIAL_CONTROLLER_BALANCE,
+            ), // CONTROLLER 2 ss58 14j2dssnCyQD2xWezTvZmbZPSKDBrwKcAcBtXrE6XgeGfzGx
+            (
+                hex!("3e31eaf98e0f6dd9b088c095757fac00dc9e3fe6af29e42caea94f7d76dd3444").into(),
+                INITIAL_CONTROLLER_BALANCE,
+            ), // CONTROLLER 3 ss58 12QYoFyrrEomSq4tMPxVxnWvPwVVKQHUpDWZuDxcg5Gv3c9E
+            (
+                hex!("0619e4018531059b677d70b476b72431d3e88fe71d2c27bba5d945975e824956").into(),
+                INITIAL_CONTROLLER_BALANCE,
+            ), // COLLATOR 1
```
zqhxuyuan, line 268: should be CONTROLLER4. and bellow also need to rename the right order

Garandor: good point, the name `INITIAL_CONTROLLER_BALANCE` is misleading here.
We just decided to give one collator more starting balance for emergency recovery.
It is not an actual multisig contoller account, i'd keep it this way, hope this explanation is enough

zqhxuyuan (approved):

PR #976, opened 2023-01-09
Move all imported constants to our code base

### LINKED ISSUE #596: All constants should be controlled by us completely except special reason.
Like these constants:
```rust
use frame_support::weights::constants::{BlockExecutionWeight, ExtrinsicBaseWeight, WEIGHT_PER_SECOND};
...
use polkadot_runtime_common::{BlockHashCount, RocksDbWeight, SlowAdjustingFeeUpdate};
```

ghzlatarev:
## Description

Move constants to our code-base, in order to not have them change without us realizing:
* BlockExecutionWeight
* ExtrinsicBaseWeight
* WEIGHT_PER_SECOND
* BlockHashCount
* RocksDbWeight
* SlowAdjustingFeeUpdate

---

Situational Notes:


primitives/manta/src/constants.rs
```diff
@@ -96,3 +100,21 @@ pub const MANTA_PAY_PALLET_ID: PalletId = PalletId(*b"mantapay");
...
+pub const WEIGHT_PER_MILLIS: Weight = WEIGHT_PER_SECOND / 1000;
+/// 1_000_000
+pub const WEIGHT_PER_MICROS: Weight = WEIGHT_PER_MILLIS / 1000;
+/// 1_000
+pub const WEIGHT_PER_NANOS: Weight = WEIGHT_PER_MICROS / 1000;
+
+parameter_types! {
+    /// By default, Substrate uses RocksDB, so this will be the weight used throughout
+    /// the runtime.
+    pub const RocksDbWeight: RuntimeDbWeight = RuntimeDbWeight {
```
Dengjianping, line 116: Because we're going to control all constants, but if there's any change from upstream, maybe we don't know it, so we better write some test cases to cover it, like
```rust
#[test]
fn diff_constants_with_upstream() {
    ...
    assert!(manta_primitives::RocksDbWeight == frame_support::weights::RuntimeDbWeight);
    ...
}
```
`RocksDbWeight` is a primitive constant, we should cover primitive constants only.

runtime/common/src/lib.rs
```diff
@@ -77,3 +77,37 @@ parameter_types! {
 /// https://research.web3.foundation/en/latest/polkadot/overview/2-token-economics.html#-2.-slow-adjusting-mechanism
 pub type SlowAdjustingFeeUpdate<R> =
     TargetedFeeAdjustment<R, TargetBlockFullness, AdjustmentVariable, MinimumMultiplier>;
+
+parameter_types! {
+    /// Time to execute an empty block.
+    /// Calculated by multiplying the *Average* with `1` and adding `0`.
```
ferrell-code, line 83: Not entirely sure what adding zero means, but I imagine this is copied from the substrate repo, so probably not important

primitives/manta/src/constants.rs
ghzlatarev: I already moved `RocksDbWeight` to manta-primitives, because it is not just needed in the runtime/ folder.

runtime/common/src/lib.rs
ghzlatarev: Yeah this is a copy pasted comment, and yeah I don't really get it as well. Multiply by 1 is just as strange.

ferrell-code: Yeah this whole comment reads like its a bot or something

ferrell-code (dismissed):

Dengjianping:
what about this `RocksDbWeight` in weight files? Maybe you have to update weight template.
https://github.com/Manta-Network/Manta/blob/manta/runtime/calamari/src/weights/calamari_vesting.rs#L43

primitives/manta/src/constants.rs
Dengjianping: Not sure I get your point. I mean maybe we won't keep it up-to-date, we should know changes from upstream 's `RocksDbWeight`. So writing test case to compare both `RocksDbWeight` is a solution.

runtime/common/src/lib.rs
Garandor: these are likely "wiggle factors" and offsets they have as inputs for the benchmark output script that generated these stats. They started with higher numbers and eventually decided the benchmark is accurate enough to not need manual "safety buffers"

primitives/manta/src/constants.rs
Garandor: I agree with Jamie here that there should be a test whether our `manta_primitives::RocksDbWeight` constant is equal to the parity-provided `frame_support::weights:: RocksDbWeight` so we get pushed to investigate the change if they desync.

ghzlatarev: Oooh I get what you guys mean now, yeah no problem.

ghzlatarev:
> 

They will be regenerated in the release PR, but I can also do it here if you guys want.

runtime/manta/src/fee.rs
```diff
@@ -16,10 +16,10 @@
 
 use crate::currency;
 use frame_support::weights::{
-    constants::ExtrinsicBaseWeight, WeightToFeeCoefficient, WeightToFeeCoefficients,
-    WeightToFeePolynomial,
+    WeightToFeeCoefficient, WeightToFeeCoefficients, WeightToFeePolynomial,
 };
 use manta_primitives::types::Balance;
+use runtime_common::ExtrinsicBaseWeight;
```
Garandor, line 22: i think we should have these tests below too

https://github.com/paritytech/substrate/blob/master/frame/support/src/weights/extrinsic_weights.rs#L68

And then another assert if our numbers are outside of 80-120% of theirs.

LGTM otherwise

Dengjianping:
> > 
> 
> They will be regenerated in the release PR, but I can also do it here if you guys want.

Yeah, we can fix it in next release.

runtime/manta/src/fee.rs
ghzlatarev: Ill add an equals assert, but not sure what's the point of the 80-120 range. We can make a judgement when the assert_eq fails.

Garandor: The point is that we don't want to have to change on every small random variance difference from upstream rerunning the benchmark, which they likely do every release. We only want to notice larger changes.
Could also be 90-110, i just don't want to investigate `assert_eq!(4_000_000,4_000_001)` every time we upgrade deps.

Range should probably be 90 - 101%. if they get slower for some reason, we need to investigate if we have to get slower too, otherwise we might break execution time limits.

ghzlatarev: They don't seem to change often. This specific one `ExtrinsicBaseWeight` didn't change for 9 months.

Dengjianping (approved):

ferrell-code (approved):

Issue #496, opened 2022-04-18
Better error monitoring and error propagation from manta-rs, pallet-assets, pallet-balances in pallet-manta-pay


ghzlatarev:
For example manta-pay's `public_transfer` returns generic error message for failures like insufficient funds for existential deposit.
This can be improved.

https://github.com/Manta-Network/Manta/pull/977#discussion_r1071076348 Otherwise we can't really do monitoring of unexpected errors that break the ledger.

ghzlatarev:
Needs https://github.com/paritytech/substrate/pull/10242 in 0.9.19 upgrade

PR #977, opened 2023-01-09
Suspend MantaPay when InternalLedgerError

ghzlatarev:

## Description

The following errors mean that the ledger is uncrecovarble broken:
* `ReceiverLedgerError::ChecksumError`
*  `ReceiverLedgerError::MerkleTreeCapacityError`
* `TransferLedgerError::ChecksumError`
* `TransferLedgerError::VerifiyingContextDecodeError`

When they occur we just trigger tx-pause `pause_pallets` for MantaPay, in order to shut down any further operations, until we figure out a way forward.

---

Situational Notes:


pallets/manta-pay/src/lib.rs
```diff
@@ -126,6 +126,17 @@ pub mod pallet {
 
     const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
 
+    ///
```
ghzlatarev: todo: comment

pallets/manta-pay/src/lib.rs
```diff
@@ -474,23 +488,35 @@ pub mod pallet {
...
+        T: Config,
+    {
         #[inline]
         fn from(err: SenderPostError<SenderLedgerError>) -> Self {
             match err {
                 SenderPostError::AssetSpent => Self::AssetSpent,
                 SenderPostError::InvalidUtxoAccumulatorOutput => Self::InvalidUtxoAccumulatorOutput,
-                SenderPostError::UnexpectedError(_) => Self::InternalLedgerError,
+                SenderPostError::UnexpectedError(_) => {
+                    let _ = T::Suspender::suspend_manta_pay_execution();
```
ghzlatarev: todo: should we handle this result ?

pallets/manta-pay/src/mock.rs
```diff
@@ -260,6 +260,7 @@ impl crate::Config for Test {
     type WeightInfo = crate::weights::SubstrateWeight<Self>;
     type PalletId = MantaPayPalletId;
     type AssetConfig = MantaAssetConfig;
+    type Suspender = ();
```
ghzlatarev: todo: add a unit test

pallets/manta-pay/src/lib.rs
ferrell-code: Maybe dropping a `log::error`  would be the way to go. 

Not much we can do if it fails

pallets/manta-pay/src/lib.rs
```diff
@@ -126,6 +127,17 @@ pub mod pallet {
 
     const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
 
+    ///
+    pub trait SuspendMantaPay {
+        fn suspend_manta_pay_execution() -> DispatchResultWithPostInfo;
+    }
```
ferrell-code, line 134: We could have this function not return anything if we don't use error in pallet

then we could handle error and `log::error` when implementing this trait in the runtime files

pallets/manta-pay/src/lib.rs
ferrell-code: still todo

pallets/manta-pay/src/lib.rs
```diff
@@ -1084,10 +1194,47 @@ pub enum TransferLedgerError {
...
+where
+    T: Config,
+{
+    #[inline]
+    fn from(value: TransferLedgerError<T>) -> Self {
+        match value {
+            TransferLedgerError::InvalidProof => Self::InvalidProof,
+            TransferLedgerError::InvalidTransferShape => Self::InvalidShape,
+            TransferLedgerError::SenderLedgerError(err) => Self::Sender(err.into()),
+            TransferLedgerError::ReceiverLedgerError(err) => Self::Receiver(err.into()),
```
ferrell-code, line 1199: I think here receiver be unknown error and not shutdown

This actually does still work :) the `err.into()` still triggers the shutdown in a nested fashion

maybe add this test case

```rust
let _e = Error::<Test>::from(TransferPostError::<Test>::Receiver(ReceiverPostError::UnexpectedError(ReceiverLedgerError::<Test>::MerkleTreeCapacityError)));
assert_manta_pay_suspension();
```

pallets/manta-pay/src/lib.rs
ghzlatarev: Yeah this seems simpler.

pallets/manta-pay/src/lib.rs
ghzlatarev: Added,

pallets/manta-pay/src/lib.rs
ghzlatarev: Commented.

ferrell-code (dismissed):
Nice, LGTM

pallets/manta-pay/src/test/payment.rs
```diff
@@ -616,3 +619,84 @@ fn pull_ledger_diff_should_work() {
...
+        Some(())
+    );
+
+    assert_eq!(
+        TransactionPause::paused_transactions((b"MantaPay".to_vec(), b"to_public".to_vec())),
+        Some(())
+    );
+
+    assert_eq!(
+        TransactionPause::paused_transactions((b"MantaPay".to_vec(), b"public_transfer".to_vec())),
```
Dengjianping: I know it's test case, but once we really get this ledger error, `public_transfer` should not be shutdown,  imo.

Dengjianping:
I forget or don't have the context. why do we suspend mantapay if this error happen?

pallets/manta-pay/src/lib.rs
```diff
@@ -485,12 +502,24 @@ pub mod pallet {
...
-                ReceiverPostError::UnexpectedError(_) => Self::InternalLedgerError,
+                ReceiverPostError::UnexpectedError(e) => {
+                    match e {
+                        ReceiverLedgerError::ChecksumError
+                        | ReceiverLedgerError::MerkleTreeCapacityError => {
+                            T::Suspender::suspend_manta_pay_execution();
+                        }
+                        _ => {}
+                    };
+                    Self::InternalLedgerError
```
SupremoUGH, line 521: Is it possible to do modify the definition of `InternalLedgerError` so we can return `Self::InternalLedgerError(e)` to keep track of the unexpected errors that don't shut down the ledger?

pallets/manta-pay/src/lib.rs
```diff
@@ -561,7 +597,16 @@ pub mod pallet {
...
-                TransferPostError::<T>::UnexpectedError(_) => Self::InternalLedgerError,
+                TransferPostError::<T>::UnexpectedError(e) => {
+                    match e {
+                        TransferLedgerError::ChecksumError
+                        | TransferLedgerError::VerifyingContextDecodeError(_) => {
+                            T::Suspender::suspend_manta_pay_execution();
+                        }
+                        _ => {}
+                    };
+                    Self::InternalLedgerError
```
SupremoUGH, line 608: Same as above.

ghzlatarev:
> I forget or don't have the context. why do we suspend mantapay if this error happen?

If these errors happen, the ledger is unrecoverable broken, so we should just shut it down.
Let me add a description, forgot about that.

SupremoUGH (dismissed):

pallets/manta-pay/src/test/payment.rs
ghzlatarev: Actually you're probably right, pallet-assets transfer is disabled.
But we can just enable pallet-assets transfer instead.

pallets/manta-pay/src/lib.rs
ghzlatarev: Will take care of this in https://github.com/Manta-Network/Manta/issues/496

pallets/manta-pay/src/test/payment.rs
ghzlatarev: Ok , removed public transfer puase for now.

SupremoUGH (dismissed):

ferrell-code (dismissed):

ferrell-code (dismissed):

ferrell-code (approved):
A few more tests that make sure errors that shouldn't suspend mantapay indeed do not maybe useful too. Code LGTM though

pallets/manta-pay/src/lib.rs
```diff
@@ -953,14 +1001,27 @@ pub enum ReceiverLedgerError {
     ///
     /// The asset has already been registered with the ledger.
     AssetRegistered,
+
+    /// Type Marker Parameter
+    Marker(PhantomData<T>),
```
Garandor, line 1006: PhantomData as enum field looks problematic to me, specifically if one was to iterate over all enum members (see other comment) because now `ReceiverLedgerError::Marker` is a valid enum field.
There are a few attempts at working around this [here](https://github.com/rust-lang/rust/issues/32739#issuecomment-205656212)

(the discussion on `exhaustive_patterns` in that issue features a cameo from one of our very own https://github.com/rust-lang/rust/issues/51085#issuecomment-974479661 )

Leaving this as resolved since we don't do iteration in this PR anyway

Garandor:
> A few more tests that make sure errors that shouldn't suspend mantapay indeed do not maybe useful too. Code LGTM though

Each invocation of `suspend_manta_pay_execution` is behind a match, so i think this is reasonably certain not to blow up randomly even without tests. 
Not sure how one would test absence of pausing for all other error conditions, [the overflow](https://stackoverflow.com/a/55056427) recommends an external crate for this purpose. I would like an iter-all-possible-enum-value-combinations-and-match style test more for assurance that *ONLY* the listed combinations pause the ledger (adding single combination negative tests is fragile on adding new error types), but this is good enough already. 
Merging into 401

Garandor (approved):

PR #981, opened 2023-01-10
Upgrade to manta-rs v0.5.9, switch to mainnet keys, fix ledger error handling

SupremoUGH:
Signed-off-by: SupremoUGH <francisco@manta.network>

## Description

Manta-rs v0.5.9 contains the calamari trusted setup keys

Added the improved error handling as in the latest commits of https://github.com/Manta-Network/manta-rs/pull/294.

Disabled stress-test to be fixed in #984 

---

Before we can approve this PR for merge, please make sure that **all** the following items have been checked off:
- [x] Connected to an issue with discussion and accepted design using zenhub "Connect issue" button below
- [x] Added **one** label out of the `L-` group to this PR
- [x] Added **one or more** labels from the `A-` and `C-` groups to this PR
- [x] Explicitly labelled `A-calamari`, `A-dolphin` and/or `A-manta` if your changes are meant for/impact either of these (CI depends on it)
- [x] Re-reviewed `Files changed` in the Github PR explorer.

Situational Notes:
- If adding functionality, write unit tests!
- If importing a new pallet, choose a proper module index for it, and allow it in `BaseFilter`. Ensure **every** extrinsic works from front-end. If there's corresponding tool, ensure both work for each other.
- If needed, update our Javascript/Typescript APIs. These APIs are officially used by exchanges or community developers.
- If modifying existing runtime storage items, make sure to implement storage migrations for the runtime and test them with `try-runtime`. This includes migrations inherited from upstream changes, and you can search the diffs for modifications of `#[pallet::storage]` items to check for any.

PR #982, opened 2023-01-11
Fix codec issue for dense-pull-ledger-diff

### LINKED ISSUE #979: `senders` and `receivers` are empty in the response of `dense_pull_ledger_diff`
When runtime(wasm) gets the data from related storages, then pass it to client(host, jsonrpsee), `codec(skip)` will skip to encode both two fields.
The response like 
```
{
  should_continue: false,
  receivers: '',
  senders: '',
  senders_receivers_total: '0x02000000000000000000000000000000',
  next_checkpoint: null
}
```
This is caused by the macro `codec(skip)`
```rust
ub struct DensePullResponse {
    ...
    #[codec(skip)]
    pub receivers: String,

    #[codec(skip)]
    pub senders: String,
    ...
}
```
But runtime(codec) doesn't allow to use `String`, removing `#[codec(skip)]` will report compilation error.
```
#[derive(Clone, Debug, Encode, Default, Eq, Hash, Decode, PartialEq, TypeInfo)]
      |                        ^^^^^^ the trait `WrapperTypeEncode` is not implemented for `String`
```

Dengjianping:

## Description
The workaround is that I implement `From` trait for DensePullResponse.
```rust
impl From<PullResponse> for DensePullResponse {
    #[inline]
    fn from(resp: PullResponse) -> DensePullResponse {
        Self {
            should_continue: resp.should_continue,
            receivers: base64::encode(resp.receivers.encode()),
            senders: base64::encode(resp.senders.encode()),
            senders_receivers_total: resp.senders_receivers_total,
            next_checkpoint: None,
        }
    }
}
```

When client api calls runtime api, `PullResponse` will be converted into `DensePullResponse`.
```rust
#[async_trait]
impl<B, C> PullApiServer for Pull<B, C>
where
    B: Block,
    C: 'static + ProvideRuntimeApi<B> + HeaderBackend<B>,
    C::Api: PullLedgerDiffApi<B>,
{
    #[inline]
    fn dense_pull_ledger_diff(
        &self,
        checkpoint: Checkpoint,
        max_receivers: u64,
        max_senders: u64,
    ) -> RpcResult<DensePullResponse> {
        let api = self.client.runtime_api();
        let at = BlockId::hash(self.client.info().finalized_hash);
        api.pull_ledger_diff(&at, checkpoint.into(), max_receivers, max_senders)
            .map(Into::into) // PullResponse to DensePullResponse
            .map_err(|err| {
                CallError::Custom(ErrorObject::owned(
                    PULL_LEDGER_DIFF_ERROR,
                    "Unable to compute dense state diff for pull",
                    Some(format!("{err:?}")),
                ))
                .into()
            })
    }
}
```
Now the response will be returned correctly.
```
{
  should_continue: false,
  receivers: 'J0vqVrFQV45G3bIfUnzxdKGRsPRjkbhqScQf0WmosgRP8vqVpKI5YlkgM8468+rxNqpE..`
  senders: 'AA==',
  senders_receivers_total: '0x02000000000000000000000000000000',
  next_checkpoint: null
}
```

So the runtime api `dense_pull_ledger_diff` can be removed, `RuntimeDensePullResponse` is not necessary anymore.
code is more cleaner.

closes #979 
---

Situational Notes:


Dengjianping:
I will submit typescript based client test cases in another pr soon once this pr gets merged.

ghzlatarev:
@Dengjianping description please.

pallets/manta-pay/src/test/payment.rs
```diff
@@ -592,25 +592,34 @@ fn pull_ledger_diff_should_work() {
...
         );
+
         assert_eq!(
-            pull_response.should_continue,
-            dense_pull_response.should_continue
+            pull_response.senders_receivers_total,
+            runtime_dense_pull_response.senders_receivers_total
         );
 
+        // convert runtime response into native response
```
ghzlatarev, line 604: why

flame4 (dismissed):
Cool solution!

pallets/manta-pay/src/test/payment.rs
Dengjianping: Just for testing
```rust
impl From<RuntimeDensePullResponse> for DensePullResponse { ... }
```

Dengjianping:
> @Dengjianping description please.

Added.

ghzlatarev (dismissed):

ferrell-code (dismissed):

pallets/manta-pay/src/lib.rs
```diff
@@ -88,7 +88,7 @@ use manta_util::{
 
 pub use crate::types::{Checkpoint, RawCheckpoint};
 pub use pallet::*;
-pub use types::{DensePullResponse, PullResponse};
+pub use types::{DensePullResponse, PullResponse, RuntimeDensePullResponse};
```
Garandor: do we need `DensePullResponse` in the pallet at all? Seems unused now

pallets/manta-pay/src/lib.rs
```diff
@@ -680,20 +680,19 @@ pub mod pallet {
...
             let senders_receivers_total = (0..=255)
                 .map(|i| ShardTrees::<T>::get(i).current_path.leaf_index as u128)
                 .sum::<u128>()
                 + NullifierSetSize::<T>::get() as u128;
-            DensePullResponse {
+            RuntimeDensePullResponse {
                 should_continue: more_receivers || more_senders,
-                receivers: base64::encode(receivers.encode()),
-                senders: base64::encode(senders.encode()),
+                receivers: receivers.encode(),
```
Garandor: since you're dropping the base64 encoding here, is it still accurate to call it `Dense`  or should it just be `RuntimePullResponse`, since the client does the encoding now.

~~Actually, does it? I didn't see the base64::encode anywhere else in this diff~~

pallets/manta-pay/src/runtime.rs
```diff
@@ -16,12 +16,12 @@
 
 //! MantaPay Runtime APIs
 
-use crate::{DensePullResponse, PullResponse, RawCheckpoint};
+use crate::{PullResponse, RawCheckpoint, RuntimeDensePullResponse};
 
 sp_api::decl_runtime_apis! {
     pub trait PullLedgerDiffApi {
         fn pull_ledger_diff(checkpoint: RawCheckpoint, max_receivers: u64, max_senders: u64) -> PullResponse;
```
Garandor, line 23: you could probably unify these to also return RuntimeDensePullResponse here and just convert to a `PullResponse` in the client.

This gets rid of one type in the RT and you could get rid of `dense_pull_ledger_diff` as a runtime API altogether and handle both non-dense and dense with the same runtime API. (Client API will then introduce the difference depending on the string encoding)

pallets/manta-pay/src/lib.rs
Garandor: This is much better than before actually. The runtime (single threaded, responsible for building/verifying blocks) hands off the base64 encoding workload to the other CPU cores that run the client and RPC now, so it should be much more parallelizable without choking out the runtime.

pallets/manta-pay/src/lib.rs
Dengjianping: We still need `DensePullResponse` which is used from client side.

pallets/manta-pay/src/runtime.rs
Dengjianping: I think about removing `dense_pull_ledger_diff`, not sure it will make people confused if without a dedicated runtime api. And some one might call runtime api by `state_call`.

Dengjianping: Nvm, seems we're the only user, I think we should keep one runtime api `pull_ledger_diff`.

pallets/manta-pay/src/lib.rs
Dengjianping: If rpc request is sent to full node, block production will not be affected from collator.

pallets/manta-pay/src/lib.rs
Dengjianping: `RuntimeDensePullResponse ` will be unused if `dense_pull_ledger_diff` is going to be removed.

Garandor: we need it in the client, but not in the pallet, no? So this import can go away

pallets/manta-pay/src/lib.rs
Garandor: Original question still stands tho

> you're dropping the base64 encoding here, is it still accurate to call it Dense or should it just be RuntimePullResponse

pallets/manta-pay/src/runtime.rs
Garandor: yes. people interfacing with the RT directly can also do the re-serialization themselves

pallets/manta-pay/src/lib.rs
Dengjianping: I removed it from lib.rs

pallets/manta-pay/src/lib.rs
Dengjianping: I have removed `RuntimeDensePullResponse` and `pull_ledger_diff`, so it might not confuse you anymore.

pallets/manta-pay/src/types.rs
```diff
@@ -912,11 +912,11 @@ pub struct DensePullResponse {
     /// Ledger Receiver Chunk
     // we decode the receivers/senders with our own way
     #[codec(skip)]
```
Garandor, line 914: is #[codec(skip)] still needed here?
IIRC the  client <-> frontend/sdk connection also uses SCALE encoding, so wouldn't this mean while the client has receivers/senders now, the frontend/sdk will get an empty senders/receivers pair from the client?

pallets/manta-pay/src/rpc.rs
```diff
@@ -16,7 +16,7 @@
 
 //! MantaPay RPC Interfaces
```
Garandor, line 17: Conceptual question: Should this file better be part of the node/src/rpc path instead?

This is all client code as far as i can tell, where a `pallet` is supposed to be a runtime module

pallets/manta-pay/src/types.rs
Dengjianping: Yes, I think `#[codec(skip)]/Decode/Encode/TypeInfo` can be removed.

As far as I know, frontend team doesn't do [ts generation](https://polkadot.js.org/docs/api/examples/promise/typegen), so we have to define all related types, and import them when create client api. So frontend will know those types information. Just like here
https://github.com/Manta-Network/Manta/blob/manta/tests/rpc_correctness_test.ts#L35

`receivers/senders` won't be empty anymore, I'm writing typescript based test cases for `dense_pull_ledger_diff`, and I can get response from manta node.
For frontend how to know decode response, we have defined related types in frontend, just like here:
https://github.com/Manta-Network/Manta/blob/manta/tests/types.ts

And I have added `DensePullResponse` and `dense_pull_ledger_diff` to this file in this pr #983 
https://github.com/Manta-Network/Manta/pull/983/files#diff-f0d1f2b649e764294393e5674247ce160ff7cb84cb0761307a1a852da7968539R41

pallets/manta-pay/src/rpc.rs
Dengjianping: I think it's more like a convention from substrate, check this example
https://github.com/paritytech/substrate/tree/master/frame/transaction-payment/rpc

Once we have more our own rpc methods, I'm fine if we decouple rpc code from pallet.

pallets/manta-pay/src/types.rs
Dengjianping: I just tested `dense_pull_ledger_diff ` without `#[codec(skip)]/Decode/Encode/TypeInfo`, still works

Dengjianping: And I have to remove `TypeInfo`, due to 
```
the trait `TypeInfo` is not implemented for `manta_pay::config::utxo::Checkpoint`
```
`#[codec(skip)]` will solve it, but `Decode/Encode` have to be derived again.

pallets/manta-pay/src/rpc.rs
Garandor: Ok, makes sense. That pallet also interestingly decouples the runtime_api into its own subfolder of /rpc/
https://github.com/paritytech/substrate/tree/master/frame/transaction-payment/rpc/runtime-api

We don't need to do this now since we only have one (not yet released) version of this API, but it's good to keep in mind for future changes

pallets/manta-pay/src/types.rs
Garandor: cant you just `#[derive(TypeInfo)]` for it?

Similar to 
```
#[derive(Clone, Debug, Encode, Default, Eq, Hash, Decode, PartialEq, TypeInfo)]
pub struct DensePullResponse {
```

But if I understand you correctly, this is irrelevant anyway because these annotations are not used for type generation on the typescript side.
In that case i'm ok with leaving it as `skip` as long as that doesn't mean these fields will be sent empty by the substrate client ( which you verified by testing )

Dengjianping: Seems we cannot skip field if only derive `TypeInfo`.
See this issue
https://github.com/paritytech/scale-info/issues/165#issuecomment-1223930340

Garandor (commented):
LGTM except for the test failure

ghzlatarev (approved):
LGTM , the unit tests should pass after https://github.com/Manta-Network/Manta/pull/981 is merged.

Garandor (approved):

Issue #348, opened 2022-01-12
[CI/CD]Codecov support


Dengjianping:
Now we don't have any tool for testing coverage for manta.
It would be nice if we use a tool like [codecov](https://app.codecov.io/gh).

PR #1012, opened 2023-02-03
Add zkSBTs

### LINKED ISSUE #936: SBT Pallet Prototype
Support nontransferable private NFTs

Create separate UTXO ledger for this purpose

ferrell-code:
## Description
SBT feature for Manta

Changes:
   - Pulls types in MantaPay into a support crate for common use in `pallet-manta-pay` and `pallet-manta-sbt`
   - Adds SBT functunality in `pallet-manta-sbt`
   - Adds `MantaSbt` to Dolphin and Calamari runtime along with sbt rpc enpoints in each client
   
### Workflow of `pallet-manta-sbt`: 

MantaSbt has two calls `reserve_sbt` and `to_private`. 

step 1: `reserve_sbt`
   - Reserves unique `AssetIds` by incrementing `NextSbtId`
   - Maps the above unique `AssetIds` to the polkadot address in `ReservedId` to whitelist that account. This allows this specific polkadot address to call `to_private` on those specific `AssetIds`
   
step 2: `to_private`
   - Checks that caller has `ReservedIds` in storage. And ensures that the `TransferPost` is a `ToPrivate` corresponding to the correct `AssetId` from `ReservedIds`. Then it inserts metadata and UTXO into `MantaSbt` storage.

closes: #936 
---

Situational Notes:


pallets/manta-sbt/src/benchmark/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+
+mod precomputed_coins;
+
+benchmarks! {
+    to_private {
+        let caller: T::AccountId = whitelisted_caller();
+        let factor = 1_000u32;
+        <T as crate::Config>::Currency::make_free_balance_be(&caller, T::ReservePrice::get() * factor.into());
+        Pallet::<T>::reserve_sbt(RawOrigin::Signed(caller.clone()).into())?;
+        let mint_post = TransferPost::decode(&mut &*TO_PRIVATE).unwrap();
```
ferrell-code, line 34: Is there a way to generate the `TransferPost` here? 

Would be nice to delete the `precompute_coins.rs` file

pallets/manta-pay/src/lib.rs
```diff
@@ -1402,3 +1404,275 @@ where
         Ok(())
     }
 }
+
+struct SBTLedger<T>(PhantomData<T>)
```
zqhxuyuan: Is it possible move `SBTLedger` implementation to `manta-sbt` pallet?

pallets/manta-sbt/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+                Error::<T>::NoSenderLedger
+            );
+
+            let (start_id, end_id) =
+                ReservedIds::<T>::get(&origin).ok_or(Error::<T>::NotReserved)?;
+            let asset_id: StandardAssetId = post
+                .asset_id
+                .map(id_from_field)
+                .ok_or(Error::<T>::InvalidAssetId)?
+                .ok_or(Error::<T>::InvalidAssetId)?;
```
zqhxuyuan, line 233: duplicate error?

primitives/manta/src/assets.rs
```diff
@@ -157,9 +136,29 @@ where
...
 }
 
+/// Bound for SBT Metadata Size (Could make generic over runtimes, but simpler to just hardcode)
+pub type SbtBound = ConstU32<100>;
+
+/// Metadata for each Asset Type
+///
+/// Currently supports SBT (Soul Bound Token) and Fungible Tokens
+#[derive(Clone, Debug, Decode, Encode, Eq, PartialEq, TypeInfo, MaxEncodedLen)]
+pub enum AssetMetadata<Balance> {
```
zqhxuyuan: Do we need migration for already existed AssetMetadata?

pallets/manta-pay/src/lib.rs
ferrell-code: It could be, but it is challenging to not end up with a bunch of duplicate code in error handling.

primitives/manta/src/assets.rs
ferrell-code: Yes we do. For Calamari

I think follow up PR, will make it better to review to keep diff smaller

pallets/manta-sbt/src/lib.rs
ferrell-code: yeah I think `flatten()` here is better. Edit: actually `and_then` is flatten and map together :)

One Error is if `TransferPost` is None for `asset_id` field and other is if `id_from_field` fails

pallets/manta-pay/src/test/payment.rs
```diff
@@ -695,3 +693,140 @@ fn transfer_ledger_errors_should_shut_down() {
...
+                None,
+                vec![],
+                vec![],
+                PalletTransferPost::try_from(private_transfer.clone()).unwrap(),
+                AssetType::SBT
+            ),
+            Error::<Test>::AssetSpent
+        );
+
+        // Private transfer succeeds, but this does not matter as ownership of sbt is only determined by reconstructing `ToPrivate` UTXO
```
ferrell-code: Cannot filter out `PrivateTransfer` as it is, well, private. 

The only way to avoid this to my understanding would to be have the SBT UTXOs in different storage from `MantaPay`. 

I opted to not do this because then we would need separate `pull_ledger_diff` rpc, duplicate indexer infrastructure  and a lot of additional downstream code. 

I think allowing dummy `private_transfers` of sbts is better than the alternative

pallets/manta-sbt/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+            ensure!(asset_id == start_id, Error::<T>::InvalidAssetId);
+
+            // Updates SbtMetadata
+            T::UpdateMetadata::update_metadata(&asset_id, AssetMetadata::SBT(metadata))?;
+            // Increments id by one, remove from storage if reserved asset_ids are exhausted
+            let increment_start_id = start_id
+                .checked_add(One::one())
+                .ok_or(ArithmeticError::Overflow)?;
+            if increment_start_id == end_id {
+                ReservedIds::<T>::remove(&origin)
```
ghzlatarev: This remove is not really clear to me.

ghzlatarev: Also can you add the workflow overview in the PR description please.

pallets/manta-sbt/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+                &who,
+                &Self::account_id(),
+                T::ReservePrice::get(),
+                ExistenceRequirement::KeepAlive,
+            )?;
+
+            // Reserves uniques AssetIds to be used later to mint SBTs
+            let start_id =
+                T::UpdateMetadata::create_asset(AssetMetadata::SBT(BoundedVec::default()))?;
+            for _ in 1..T::MintsPerReserve::get() {
```
ghzlatarev: Why start from 1 here ?

pallets/manta-sbt/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+            source: T::AccountId,
+        },
+        /// Reserve SBT
+        SBTReserved {
+            /// Public Account reserving sbt mints
+            who: T::AccountId,
+            /// Start of reserved ids
+            start_id: StandardAssetId,
+            /// end of reserved ids
+            stop_id: StandardAssetId,
```
ghzlatarev: these comments here are redundant.

primitives/manta/src/assets.rs
ghzlatarev: From what I understood we're not developing this on Calamari, so no need.

pallets/manta-sbt/src/tests.rs
```diff
@@ -0,0 +1,257 @@
...
+    parameters::load_proving_context(
+        tempfile::tempdir()
+            .expect("Unable to create temporary directory.")
+            .path(),
+    )
+}
+
+/// Samples a [`Mint`] transaction of `asset` with a random secret.
+#[inline]
+fn sample_to_private<R>(asset_id: AssetId, value: AssetValue, rng: &mut R) -> PalletTransferPost
```
ghzlatarev, line 71: Have you ran a test coverage tool before ? I know it's not something we've put into a process but it would be nice to see the coverage overall and of this pallet. This is what we used to do way back https://github.com/Manta-Network/pallet-manta-pay#test-coverage

primitives/manta/src/assets.rs
ghzlatarev: Buut actually this still needs to change in Calamari...

pallets/manta-sbt/src/benchmark/mod.rs
ghzlatarev: i think all you might need is
```
let mut rng = ChaCha20Rng::from_seed([0; 32]);
    let (proving_context, _, parameters, utxo_accumulator_model) =
        load_parameters(directory.path()).expect("Unable to load parameters.");
    let mut utxo_accumulator = UtxoAccumulator::new(utxo_accumulator_model);
    let asset_id = 0.into();

    let to_private = to_private_example(
        &proving_context.to_private,
        &parameters,
        &mut utxo_accumulator,
        asset_id,
        1,
        &mut rng,
    );
```
may be worth a try

pallets/manta-pay/src/lib.rs
ghzlatarev: Just for clarity can you write in the description what part of the logic is done in manta-sbt and what here.

pallets/manta-sbt/src/tests.rs
ferrell-code: Ive used https://about.codecov.io/ in CI, but tbh it gets annoying to maintain.

I think in this PR running a before and after of the tool you linked seems like a good idea to see if this decreases code coverage from these changes

pallets/manta-pay/src/lib.rs
ferrell-code: The SBTLedger allows for ToPrivate txs without any checks on FungibleLedger balances. This means the AssetId used must always be validated to be an SBT if not you could insert arbitrary `ToPrivate` to any pool and drain a FungibleToken Pool. This is the worst case bug that cannot happen.

The validations around the AssetId is done in `pallet-manta-sbt`, hence the giant warning comment above

Added some more to description but maybe I need to write in-depth README

pallets/manta-sbt/src/lib.rs
ferrell-code: Yeah I think I can make this cleaner. 

I need a range of `AssetIds` to reserve and store. These are the range of `AssetIds` user reserved that can now be used in MantaPay as they are registered to be SBTs not FT

pallets/manta-sbt/src/lib.rs
ferrell-code: Maybe a bit better now

pallets/manta-sbt/src/lib.rs
ferrell-code: This is once we reached the end of reserved `AssetIds` for use on `MantaPay`. 

The `AssetId` must always correspond to SBT in `pallet-asset-manager` and `pallet-manta-pay`... Perhaps I should put a defensive check to ensure that `AssetIdMetadata` is SBT in this call

pallets/manta-sbt/src/tests.rs
ghzlatarev: Yeah we don't have to integrate in the ci right, I think there's an issue already for this https://github.com/Manta-Network/Manta/issues/348
Running it manually will be more than enough.

pallets/asset-manager/src/lib.rs
```diff
@@ -296,15 +306,11 @@ pub mod pallet {
...
     #[pallet::storage]
     #[pallet::getter(fn asset_id_metadata)]
-    pub(super) type AssetIdMetadata<T: Config> = StorageMap<
-        _,
-        Blake2_128Concat,
-        T::AssetId,
-        <T::AssetConfig as AssetConfig<T>>::AssetRegistryMetadata,
-    >;
+    pub(super) type AssetIdMetadata<T: Config> =
+        StorageMap<_, Blake2_128Concat, T::AssetId, AssetMetadata<T::Balance>>;
```
Dengjianping: Do we need migration here? Not sure any change in this metadata type.

pallets/manta-pay/src/lib.rs
```diff
@@ -729,48 +725,55 @@ pub mod pallet {
...
+                            &load_transfer_parameters(),
+                            &mut SBTLedger(PhantomData),
+                            &(),
+                            sources,
+                            sinks,
+                        )
+                        .map_err(Error::<T>::from)?
+                        .convert(origin),
+                );
+                Ok(().into())
```
Dengjianping: You can move line 775 after line 776, and remove line 759.

pallets/manta-pay/src/lib.rs
```diff
@@ -1402,3 +1406,278 @@ where
...
+    #[inline]
+    fn spend_all<I>(
+        &mut self,
+        _super_key: &Self::SuperPostingKey,
+        _iter: I,
+    ) -> Result<(), Self::Error>
+    where
+        I: IntoIterator<Item = (Self::ValidUtxoAccumulatorOutput, Self::ValidNullifier)>,
+    {
+        Ok(())
```
Dengjianping: why we put a dummy implementation here? Just feel a little confused by this function name.

pallets/manta-pay/src/lib.rs
```diff
@@ -1402,3 +1406,278 @@ where
...
+    #[inline]
+    fn update_public_balances(
+        &mut self,
+        _super_key: &TransferLedgerSuperPostingKey<config::Config, Self>,
+        _asset_id: config::AssetId,
+        _sources: Vec<SourcePostingKey<config::Config, Self>>,
+        _sinks: Vec<SinkPostingKey<config::Config, Self>>,
+        _proof: Self::ValidProof,
+    ) -> Result<(), TransferLedgerError<T>> {
+        Ok(())
```
Dengjianping: Same here, confused by the function name.

pallets/manta-pay/src/lib.rs
ferrell-code: It is a trait that needs to be implemented from `manta-rs`. Hence the dummy implementation

pallets/manta-pay/src/lib.rs
ferrell-code: same it is a trait from `manta-rs`

pallets/asset-manager/src/lib.rs
ferrell-code: Yes we do, this PR turns the storage item into the enum `AssetMetadata`, Storage is currently `FungibleAssetRegistryMetadata` on calamari.

ferrell-code: same as https://github.com/Manta-Network/Manta/pull/1012#discussion_r1097409189

primitives/manta/src/assets.rs
ferrell-code: Yeah  I think this is the best way to accumulate no tech debt. Its fine if some variants are not used on both Calamari/Manta

pallets/manta-pay/src/test/payment.rs
ferrell-code: After talking with Fransisco I think we need separate storage for items for SBT Private ledger. Unfortunately this will require a `sbt_pull_ledger_diff` rpc and indexer infrastructure.

Currently, it would only be possible to prove ownership by extracting the `to_private` extrinsic from historical state which requires block number and needs archive node.

pallets/manta-sbt/src/tests.rs
```diff
@@ -0,0 +1,276 @@
+// Copyright 2020-2022 Manta Network.
```
Dengjianping: 2022 -> 2023,  seem there's some 2022s in others files.

pallets/manta-sbt/src/lib.rs
```diff
@@ -0,0 +1,268 @@
...
+                &who,
+                &Self::account_id(),
+                T::ReservePrice::get(),
+                ExistenceRequirement::KeepAlive,
+            )?;
+
+            // Reserves uniques AssetIds to be used later to mint SBTs
+            let asset_id_range: Vec<StandardAssetId> = (0..T::MintsPerReserve::get())
+                .map(|_| T::UpdateMetadata::create_asset(AssetMetadata::SBT(BoundedVec::default())))
+                .collect::<Result<Vec<StandardAssetId>, _>>()?;
```
Dengjianping, line 273: Maybe `collect` can be remove here.
Memory will be allocated once `collect` is called, but iterator won't due to laziness. And seems you just iterate `asset_id_range`.

pallets/manta-sbt/src/lib.rs
```diff
@@ -0,0 +1,268 @@
...
+    #[pallet::config]
+    pub trait Config: frame_system::Config {
+        /// The overarching event type.
+        type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+
+        /// Weight information for extrinsics in this pallet.
+        type WeightInfo: WeightInfo;
+
+        /// Balance Type
+        type Balance: Default + Member + Parameter + TypeInfo;
```
Dengjianping: Not sure why add new `Balance` type here.

primitives/manta/src/assets.rs
```diff
@@ -157,9 +136,29 @@ where
     >;
 }
 
+/// Bound for SBT Metadata Size (Could make generic over runtimes, but simpler to just hardcode)
+pub type SbtBound = ConstU32<200>;
```
ghzlatarev: Another reason to make this generic is to be less coupled with primitives.
Up to you though, or we can just log it for now.

pallets/manta-pay/src/lib.rs
ghzlatarev: Ok im a little hesitant of leaving this here. Let's at least log it.

pallets/asset-manager/src/lib.rs
```diff
@@ -52,7 +52,7 @@ pub mod pallet {
     use frame_system::pallet_prelude::*;
```
ghzlatarev: Up to you, but it would make reviewing much simpler if we can split this out to:
1. asset-manager changes PR
2. manta-support changes PR
3. add manta-sbt PR

pallets/asset-manager/src/lib.rs
```diff
@@ -296,15 +306,11 @@ pub mod pallet {
...
+    /// AssetId to AssetMetadata Map.
     #[pallet::storage]
     #[pallet::getter(fn asset_id_metadata)]
-    pub(super) type AssetIdMetadata<T: Config> = StorageMap<
-        _,
-        Blake2_128Concat,
-        T::AssetId,
-        <T::AssetConfig as AssetConfig<T>>::AssetRegistryMetadata,
-    >;
+    pub(super) type AssetIdMetadata<T: Config> =
```
ghzlatarev: does this even need to be (super) what is accessing it ?

pallets/asset-manager/src/lib.rs
```diff
@@ -341,23 +347,16 @@ pub mod pallet {
...
-                asset_id,
-                metadata.clone().into(),
-                metadata.min_balance().clone(),
-                metadata.is_sufficient(),
-            )
-            .map_err(|_| Error::<T>::ErrorCreatingAsset)?;
+
+            let asset_id = Self::create_asset(metadata.clone())?;
             AssetIdLocation::<T>::insert(asset_id, &location);
-            AssetIdMetadata::<T>::insert(asset_id, &metadata);
```
ghzlatarev, line 360: Why not inserting the metadata anymore ? Is something else doing this

pallets/manta-sbt/src/tests.rs
```diff
@@ -0,0 +1,276 @@
...
+        let id = field_from_id(10);
+        let post = sample_to_private(id, value, &mut rng);
+        assert_noop!(
+            MantaSBTPallet::to_private(MockOrigin::signed(ALICE), Box::new(post), bvec![0]),
+            Error::<Test>::NotReserved
+        );
+    });
+}
+
+/// Private transfer will fail
```
ghzlatarev: can you add why to this comment.

pallets/manta-sbt/src/tests.rs
```diff
@@ -0,0 +1,276 @@
...
+
+        let post = PalletTransferPost::try_from(private_transfer).unwrap();
+        assert_noop!(
+            MantaSBTPallet::to_private(MockOrigin::signed(ALICE), Box::new(post), bvec![]),
+            Error::<Test>::NoSenderLedger
+        );
+    });
+}
+
+/// ToPublic Post should fail
```
ghzlatarev: Please add why to the comment.

pallets/asset-manager/src/lib.rs
ferrell-code: Just a way to explicitly disallow outside modules from accessing storage. `pub(super)` is more restrictive than just `pub`, I think it is good idea for a piece of storage like this as extra warning

primitives/manta/src/assets.rs
ferrell-code: Is generic now

pallets/manta-sbt/src/lib.rs
ferrell-code: Yeah idk, must  of been leftover.

You are correct is unneeded

pallets/manta-sbt/src/benchmark/mod.rs
ferrell-code: Don't think it is possible because benchmarks need a `no_std` environment.
 
And the test code uses `std`. Best action is to follow up with #1022 and move precompute coins to its own crate

pallets/manta-pay/src/lib.rs
ferrell-code: SBTLedger is now in `pallet-manta-sbt`. Better for security as if there is bug it only effects SBTs

pallets/asset-manager/src/lib.rs
ferrell-code: Now is only

2. manta-support changes PR
3. add manta-sbt PR

pallets/manta-sbt/src/lib.rs
ferrell-code: I think collect is needed as it changes the Range value to the incremented `NextSbtId` in the `map`.

Could be mistaken but I don't see how to, could you show me snippit if I am missing something

pallets/asset-manager/src/lib.rs
ghzlatarev: I mean does it even need to be pub ?

pallets/asset-manager/src/lib.rs
ghzlatarev: I forget what the asset-manager changes were exactly, but do we need to log them separately ?

ghzlatarev:
Please don't forget to add some integration tests to the release QA checklist (until we turn them into TS/JS integration tests)

pallets/asset-manager/src/lib.rs
ferrell-code: Well then it is out of scope anywhere outside `pub mod pallet`. Most of the time it is needed to be in scope for a test or maybe some  trait implementation

(This particular example is now reverted)

pallets/asset-manager/src/lib.rs
ferrell-code: I reverted this change and moved the logic to `pallet-manta-sbt` as it is a different ledger and now can have overlapping `AssetIds` with `MantaPay`

The `asset-manager` change was mainly adding an `AssetMetadata`  enum to allow for different types of metadata in storage. This change maybe still useful in the future.

pallets/manta-sbt/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+
+        /// Pallet ID
+        type PalletId: Get<PalletId>;
+
+        /// Number of unique Asset Ids reserved per `reserve_sbt` call, is the amount of SBTs allowed to be minted
+        #[pallet::constant]
+        type MintsPerReserve: Get<u16>;
+
+        /// Price to reserve Asset Ids
+        type ReservePrice: Get<BalanceOf<Self>>;
```
ghzlatarev: Can't the palletid and reserveprice be #[pallet::constant] ?

pallets/manta-sbt/src/mock.rs
```diff
@@ -0,0 +1,322 @@
...
+                b"MantaPay".to_vec(),
+                vec![
+                    b"to_private".to_vec(),
+                    b"private_transfer".to_vec(),
+                    b"to_public".to_vec(),
+                ],
+            )],
+        );
+    }
+}
```
ghzlatarev: Should we add a suspender for the SBT pallet as well ? Separate issue though.

pallets/manta-support/src/lib.rs
```diff
@@ -0,0 +1,23 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Manta.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Manta Support
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+extern crate alloc;
```
ghzlatarev, line 21: Is this needed ?

pallets/manta-sbt/src/lib.rs
```diff
@@ -0,0 +1,1029 @@
...
+                    *receivers_pulled += 1;
+                    receivers.push(next);
+                }
+                _ => return false,
+            }
+        }
+        Shards::<T>::contains_key(shard_index, max_receiver_index)
+    }
+
+    /// Pulls sender data from the ledger starting at the `sender_index`.
```
ghzlatarev: This comment is wrong now. Maybe worth scanning this page again .

pallets/manta-sbt/src/bin/precompute_coins.rs
```diff
@@ -0,0 +1,146 @@
...
+    )?;
+
+    let directory = tempfile::tempdir().expect("Unable to generate temporary test directory.");
+    println!("[INFO] Temporary Directory: {directory:?}");
+
+    let mut rng = ChaCha20Rng::from_seed([0; 32]);
+    let (proving_context, _, parameters, utxo_accumulator_model) =
+        load_parameters(directory.path()).expect("Unable to load parameters.");
+    let mut utxo_accumulator = UtxoAccumulator::new(utxo_accumulator_model);
+    let asset_id = 0.into();
```
ghzlatarev: The benchmarks don't fail with asset id 0 ? That's a little worrisome (although there could be a valid reason for it)
Also, let's log an issue how to dedupe this and the manta-pay version of the code.

pallets/manta-sbt/src/mock.rs
ferrell-code: We could, less important than manta pay tho, as there are no funds to be drained when Ledger get borked

pallets/manta-sbt/src/lib.rs
ferrell-code: Underlying Function is unneeded too

pallets/manta-sbt/src/mock.rs
ghzlatarev: Actually is the suspender even a good idea at this point, as MantaPay is not the only ledger on the blockchain ?

pallets/manta-support/src/lib.rs
ferrell-code: Yeah needed for use of `alloc` in `manta_pay.rs`

pallets/manta-sbt/src/mock.rs
ferrell-code: suspender would only stop MantaPay, I think it still make sense.

And a suspender of SBT should only stop SBT Ledger.

We may want to change the trait names to be more generic like `SuspendLedgerManager` or something, but should work fine just need an separate SBT implementation

ghzlatarev: Oh yeah my bad

pallets/manta-sbt/src/lib.rs
ferrell-code: yes they should be

pallets/manta-sbt/src/bin/precompute_coins.rs
ferrell-code: Ya this is weird, should start the counter at a value of one for this reason.

But I would think this should panic?

ferrell-code: deduping this somewhat of an opened issue #1022

ferrell-code: Made the counter initialize with value of 1. But I'm confused how this worked in the first place with a value of 0.

pallets/manta-sbt/src/lib.rs
Dengjianping: `MintsPerReserve` is really small, so it's fine that we still use `collect`.

runtime/calamari/src/assets_config.rs
```diff
@@ -175,3 +182,17 @@ impl pallet_manta_pay::Config for Runtime {
...
+}
+
+impl pallet_manta_sbt::Config for Runtime {
+    type Event = Event;
+    type PalletId = MantaSbtPalletId;
+    type Currency = Balances;
+    type MintsPerReserve = ConstU16<5>;
+    type ReservePrice = ConstU128<{ 100 * KMA }>;
+    type SbtMetadataBound = ConstU32<300>;
+    type WeightInfo = ();
```
zqhxuyuan: change to manta-sbt weights

ferrell-code: Added, still need to actually benchmark it using CI, which will be done on release.

pallets/manta-sbt/src/lib.rs
```diff
@@ -0,0 +1,905 @@
...
+            <T as pallet::Config>::Currency::transfer(
+                &who,
+                &Self::account_id(),
+                T::ReservePrice::get(),
+                ExistenceRequirement::KeepAlive,
+            )?;
+
+            // Reserves uniques AssetIds to be used later to mint SBTs
+            let asset_id_range: Vec<StandardAssetId> = (0..T::MintsPerReserve::get())
+                .map(|_| Self::next_sbt_id_and_increment())
```
zqhxuyuan, line 272: Can we get rid of this loop, and return only tuple of (start_id,end_id). In each tx, start_id=next_sbt_id, end_id=NextSbtId+MintsPerReserve-1. then store NextSbtId only once.

zqhxuyuan: Another situation is: If user reserve asset ids but not mint sbt, and call reserve asset ids again, we should return original reserved asset ids, instead reserve new asset ids.
This is also suitable for: If user reserve 5 assets(1,5), but mint 1 sbt(1), then next time user want to reserve again, we should return (2,5).

ferrell-code: In that case as its currently written user doesn't need to call reserve_sbt again. Just mint ids 2-5

ghzlatarev (commented):
I guess you can put `ignore` on the fee test, we already have an issue.

Garandor:
> I guess you can put `ignore` on the fee test, we already have an issue.

https://github.com/Manta-Network/Manta/pull/1053 is ready to get merged after review and will resolve this

ghzlatarev (approved):

ghzlatarev:
> ⚠️ Congestion test: 1-day congestion cost (calamari) is NOT above target_daily_congestion_cost_kma

nice

zqhxuyuan (approved):

PR #1015, opened 2023-02-09
Re-new MantaPay precomputed coins and re-enable tests

ghzlatarev:
## Description

* Recomputed the MantaPay coins used for testing using the latest MantaPay
* Updated the tests folder to use the newly generated files
* Re-enabled the CI tests that used the precomputed coins

---

Situational Notes:


tests/insert_coins.js
```diff
@@ -0,0 +1,130 @@
...
+const mints = [];
+
+async function main() {
+    let nodeAddress = "ws://127.0.0.1:9801";
+    
+    const api = await createPromiseApi(nodeAddress);
+    let sender = keyring.addFromMnemonic("bottom drive obey lake curtain smoke basket hold race lonely fit walk//Alice");
+    let receiver = keyring.addFromMnemonic("bottom drive obey lake curtain smoke basket hold race lonely fit walk//Bob");
+
+    const mints_file = '/home/georgi/Desktop/workspace/Manta/precomputed-15k-iterations/precomputed_mints_v3-5';
```
ferrell-code, line 21: Was this the script run to generate the precompute coins? Maybe add comment saying that it does that

ferrell-code (commented):
Nice

tests/insert_coins.js
ghzlatarev: Yes it was, I need to write a README of the whole process, right now it's arcane knowledge that only I possess.
But let me split the README out of this PR, because I don't have capacity to write it in the next 2 - 3 days.

ghzlatarev: https://github.com/Manta-Network/Manta/issues/1016

ferrell-code (dismissed):

tests/insert_coins.js
```diff
@@ -0,0 +1,130 @@
...
+
+    let mints_offset = 2;
+    let transfers_offset = 4;
+    let reclaims_offset = 4;
+    let total_iterations = 15000;
+    let collected_for_batch = 0;
+
+    let mint_size = 552;
+    var mints_buffer = Buffer.alloc(mint_size * total_iterations);
+    fs.open(mints_file, 'r',  async  function(status, fd1) {
```
Garandor: This looks like http://callbackhell.com/

Why not use promise API for fs access, ur in a `async fn` anyway

`var file = await fs.promises.open;`
havent tested the exact syntax, but there has to be a promise API for the fs module

Garandor: https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_promises_api

ghzlatarev: Should be done now.

tests/insert_coins.js
```diff
@@ -0,0 +1,114 @@
...
+        if (status.isDropped) {
+            console.log("tx %i dropped.", status.nonce);
+            unsub();
+        }
+        if (status.isUsurped) {
+            console.log("tx %i usurped.", status.nonce);
+            unsub();
+        }
+        if (status.isFinalityTimeout) {
+            console.log("tx %i finality timeout.", status.nonce);
```
Garandor: aren't these error cases? Why not abort if this happens?

tests/insert_coins.js
```diff
@@ -0,0 +1,114 @@
...
+
+        let transfers_start = transfers_offset + i * (2 * mint_size + transfer_size);
+        const transfer_mint_1 = await api.tx.mantaPay.toPrivate(transfers_buffer.subarray(transfers_start, transfers_start + mint_size));
+        const transfer_mint_2 = await api.tx.mantaPay.toPrivate(transfers_buffer.subarray(transfers_start + mint_size, transfers_start + 2 * mint_size));
+        const transfer = await api.tx.mantaPay.privateTransfer(transfers_buffer.subarray(transfers_start + 2 * mint_size, transfers_start + 2 * mint_size + transfer_size));
+        transactions.push(transfer_mint_1);
+        transactions.push(transfer_mint_2);
+        transactions.push(transfer);
+
+        await api.tx.utility.forceBatch(transactions).signAndSend(sender, {nonce: -1}, ({ events = [], status }) => {
```
Garandor: by awaiting here you're sending one transaction and waiting for it to return before sending the next one, i don't think this is what you want, it'll be slow AF. Also `unsub` is meaningless with promises.
Also your arrow fn will only get executed *once* per batch transaction

tests/insert_coins.js
ghzlatarev: The unsub() terminates.
These were used to catch to debug something.
At one point during testing the transactions were being dropped, but the script would continue.
I can remove these altogether, tehy are not needed.

tests/insert_coins.js
ghzlatarev: This is supposed to be run once only, when there's a change to the circuit, its ok to be slow.
Ill add a nice write-up separately as Charlie already asked for.

tests/insert_coins.js
Garandor: i see this is a weird mix of promise and callback API in this fn.I'm mainly confused because i don't see the promise resolving anywhere. so it looks to me that this should wait forever. if that's not the case it's fine

Garandor: i think it's fine to keep these logs, but i'm missing termination of this js with error. if one of them occurs. I guess its just for testing so no need to be too careful, but normally i'd like to detect the sequence failed from some error code without looking through 14k lines of logs.

Garandor: maybe at least promote these to log.err()

ghzlatarev: I cleaned it up some more and using log.err() now

tests/insert_coins.js
```diff
@@ -0,0 +1,97 @@
...
+            nonce: -1
+        }, ({
+            events = [],
+            status
+        }) => {
+            if (status.isFinalized) {
+                err.log("tx %i success.", status.nonce);
+            }
+            if (status.isDropped || status.isUsurped || status.isFinalityTimeout || status.isRetracted) {
+                err.log(`tx %i ${status.type}.`, status.nonce);
```
Garandor: i don't get where the `err` comes from, i meant `console.log()` for success and `console.err()` for failures, but if this works im ok with it

Garandor (dismissed):

tests/insert_coins.js
ghzlatarev: Ah ok now I get it, that makes more sense.

Garandor (approved):

tests/insert_coins.js
```diff
@@ -0,0 +1,97 @@
...
+    });
+    await api.isReady;
+    return api;
+}
+
+async function main() {
+    const api = await createPromiseApi(nodeAddress);
+    const sender = keyring.addFromMnemonic("bottom drive obey lake curtain smoke basket hold race lonely fit walk//Alice");
+
+    const mints_file = '/home/georgi/Desktop/workspace/Manta/precomputed-15k-iterations/precomputed_mints_v3-5';
```
Dengjianping, line 21: Is the file path from your machine?

tests/insert_coins.js
```diff
@@ -0,0 +1,97 @@
...
+    const reclaims_file = '/home/georgi/Desktop/workspace/Manta/precomputed-15k-iterations/precomputed_reclaims_v3-5';
+
+    const mints_buffer = await fs.readFile(mints_file);
+    const transfers_buffer = await fs.readFile(transfers_file);
+    const reclaims_buffer = await fs.readFile(reclaims_file);
+
+    const mints_offset = 2;
+    const transfers_offset = 4;
+    const reclaims_offset = 4;
+    const total_iterations = 15000;
```
Dengjianping, line 32: I suppose` total_iterations` can be calculated by transfers_buffer/reclaims_offset/mint_size.
But it's ok to use a hard code here.

tests/insert_coins.js
Garandor: Will address in #1016

tests/insert_coins.js
Garandor: Will address in #1016

Dengjianping (approved):

PR #1036, opened 2023-03-06
Pallet Staking Lottery

Garandor:
Implements an APY-maximizing no-loss lottery for Manta/Calamari

- `liquidate_lottery` and `rebalance_stake` extrinsics are left unimplemented for later
- [x] reverify `DrawingFreezeout` is long enough for both Polkadot and Kusama

Note that the test `winner_distribution_should_be_equal_with_equal_deposits` is nondeterministic and can sometimes fail.
I have set the boundary low enough for random failures to be rare, but nonetheless, this test failing is not necessarily indication of an error if it occurs  isolated.

---

Situational Notes:


pallets/randomness/src/lib.rs
```diff
@@ -0,0 +1,201 @@
+// Copyright 2019-2022 PureStake Inc.
+// This file is part of Moonbeam.
```
ghzlatarev: moonbeam or manta here ?

pallets/randomness/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+        ///
+        /// Note: This needs to be updated when asynchronous backing is in effect,
+        ///       as it will be unsafe.
+        fn random(subject: &[u8]) -> (T::Hash, BlockNumberFor<T>) {
+            let relay_epoch_index = <RelayEpoch<T>>::get();
+            let randomness_output =
+                RandomnessResults::<T>::get(RequestType::BabeEpoch(relay_epoch_index))
+                    .unwrap()
+                    .randomness
+                    .unwrap(); // TODO: Handle unavailable randomness
```
ghzlatarev: A bunch of TODOs in the code. Not sure which of them are supposed to be done now and which are issues for the future.

pallets/pallet-lottery/src/lib.rs
```diff
@@ -0,0 +1,1074 @@
...
+    pub struct Request<AccountId, BlockNumber, Balance> {
+        pub user: AccountId,
+        pub block: BlockNumber,
+        pub balance: Balance,
+    }
+
+    #[pallet::storage]
+    #[pallet::getter(fn withdrawal_request_queue)]
+    pub(super) type WithdrawalRequestQueue<T: Config> =
+        StorageValue<_, Vec<Request<T::AccountId, T::BlockNumber, BalanceOf<T>>>, ValueQuery>;
```
ghzlatarev, line 259: Maybe this should be bounded vec ? Someone could do a bunch of min-value requests to spam

ghzlatarev: I guess it can also grow infinitely since there's no time limit to withdrawing

pallets/pallet-lottery/src/staking.rs
```diff
@@ -0,0 +1,495 @@
...
+                e.error
+            })?;
+        }
+        StakedCollators::<T>::mutate(&collator, |balance| *balance += amount);
+
+        log::debug!("Delegated {:?} tokens to {:?}", amount, collator);
+        Ok(())
+    }
+
+    #[named]
```
ghzlatarev: why was this attribute needed btw ?

pallets/pallet-lottery/src/staking.rs
```diff
@@ -0,0 +1,495 @@
...
+        let now = <frame_system::Pallet<T>>::block_number();
+        let info = pallet_parachain_staking::Pallet::<T>::round();
+        let active_collators: BTreeSet<_> =
+            pallet_parachain_staking::Pallet::<T>::selected_candidates()
+                .into_iter()
+                .filter(|collator| {
+                    // being selected is not enough, it must also be actively participating in block production
+                    // but we need to ensure it had a chance to produce blocks, so we only check this if we're at least 25% into a round
+                    if now < info.first + (info.length / 4u32).into() {
+                        return true;
```
ghzlatarev: Does the logic rely on each collator producing at least one block ? The more the better ? Or that you will keep producing blocks ?

pallets/pallet-lottery/src/staking.rs
```diff
@@ -0,0 +1,495 @@
...
+
+        // second concern: We want to maximize staking APY earned, so we want to balance the staking pools with our deposits while conserving gas
+        // We only consider active collators for deposits
+        // TODO: Small optimization: Also consider points / pointsAwarded to not stake to collators missing blocks
+
+        let mut top_collator_accounts =
+            pallet_parachain_staking::Pallet::<T>::compute_top_candidates(); // XXX: This can select collators that are joined but not yet producing blocks
+        if top_collator_accounts.is_empty() {
+            // Use `SelectedCandidates` as fallback (should basically never happen)
+            log::warn!("Lottery falling back to selected candidates");
```
ghzlatarev: maybe put a why this is a warning in the message

pallets/pallet-lottery/src/staking.rs
```diff
@@ -0,0 +1,495 @@
...
+            }
+            return deposits;
+        }
+
+        // second concern: We want to maximize staking APY earned, so we want to balance the staking pools with our deposits while conserving gas
+        // We only consider active collators for deposits
+        // TODO: Small optimization: Also consider points / pointsAwarded to not stake to collators missing blocks
+
+        let mut top_collator_accounts =
+            pallet_parachain_staking::Pallet::<T>::compute_top_candidates(); // XXX: This can select collators that are joined but not yet producing blocks
```
ghzlatarev: why is the comment important

ghzlatarev: also maybe it would be better to just divide evenly in v1, this adds a lot of complexity

pallets/pallet-lottery/src/lib.rs
```diff
@@ -0,0 +1,1077 @@
...
+        /// Time in blocks until a collator is done unstaking
+        #[pallet::constant]
+        type UnstakeLockTime: Get<Self::BlockNumber>; // XXX: could maybe alculate this from staking LeaveDelayRounds * DefaultBlocksPerRound
+        /// Weight information for extrinsics in this pallet.
+        type WeightInfo: WeightInfo;
+    }
+
+    #[pallet::pallet]
+    #[pallet::storage_version(STORAGE_VERSION)]
+    // #[pallet::generate_store(pub(super) trait Store)]
```
Dengjianping: Removable?

pallets/pallet-lottery/src/lib.rs
```diff
@@ -0,0 +1,1077 @@
...
+            MinWithdraw::<T>::set(self.min_withdraw);
+        }
+    }
+
+    #[pallet::event]
+    #[pallet::generate_deposit(fn deposit_event)]
+    pub enum Event<T: Config> {
+        LotteryStarted,
+        LotteryStopped,
+        LotteryWinner(T::AccountId, BalanceOf<T>),
```
Dengjianping: Why not to use structured events?

Dengjianping:
I doubt whether we need customized rpc methods, there's `state_call`, once runtime apis are defined and implemented(already done), we can call runtime apis by `state_call`, and no need to maintain those rpc anymore.

But if we concern about the performance of rpc method, like mantapay's rpc methods, we can still stick to native rpc implementation.

pallets/pallet-lottery/Cargo.toml
```diff
@@ -0,0 +1,78 @@
+[package]
+authors = ["Adam Reif"]
+edition = "2021"
+name = "pallet-lottery"
+version = "0.1.0"
```
Dengjianping: version number

Dengjianping: And more metadata about this pallet

pallets/pallet-lottery/Cargo.toml
```diff
@@ -0,0 +1,78 @@
...
+
+[dependencies]
+codec = { version = '3.1.2', default-features = false, features = ['derive'], package = 'parity-scale-codec' }
```
Dengjianping: seems pretty old version

pallets/pallet-lottery/src/lib.rs
```diff
@@ -0,0 +1,1077 @@
...
+        Deposited(T::AccountId, BalanceOf<T>),
+        ScheduledWithdraw(T::AccountId, BalanceOf<T>),
+        Withdrawn(T::AccountId, BalanceOf<T>),
+    }
+
+    #[pallet::error]
+    pub enum Error<T> {
+        LotteryNotStarted,
```
Dengjianping, line 325: Same here, comments are missing. Once error happens, we can see some more details from blocks(polkadot-js)

pallets/pallet-lottery/src/lib.rs
```diff
@@ -0,0 +1,1077 @@
...
+                                balance: amount,
+                            })
+                        });
+                        // store reduced balance
+                        *maybe_balance = match balance
+                            .checked_sub(&amount)
+                            .ok_or(Error::<T>::ArithmeticUnderflow)?
+                        {
+                            new_balance if new_balance.is_zero() => {
+                                TotalUsers::<T>::mutate(|users| *users -= 1);
```
Dengjianping: `-` seems kind of unsafe to me here.

pallets/pallet-lottery/src/lib.rs
```diff
@@ -0,0 +1,1077 @@
...
+                    }
+                    _ => {
+                        Err("not enough left to handle this request from current unstaking funds")
+                    }
+                }
+            })
+            .or_else(|_| {
+                // Withdrawal needs extra collators to unstake to have enough funds to serve withdrawals, do it
+                let reserve = RemainingUnstakingBalance::<T>::get();
+                let mut remaining_to_withdraw = amount - reserve;
```
Dengjianping, line 514: is it possible that `remaining_to_withdraw` is negative? I found `amount` is user's input.
Maybe I ask a stupid question.

pallets/pallet-lottery/src/lib.rs
```diff
@@ -0,0 +1,1077 @@
...
+                        }
+                        _ => {
+                            Err("not enough left to handle this request after unstaking additional collators")
+                        }
+                    }
+                })
+            })?;
+            // END UNSTAKING SECTION
+            Self::deposit_event(Event::ScheduledWithdraw(caller, amount));
+            Ok::<(), DispatchError>(())
```
Dengjianping: `Ok(().into())` should work here?

pallets/pallet-lottery/src/lib.rs
```diff
@@ -0,0 +1,1077 @@
...
+                            .ok_or(ArithmeticError::Underflow)?;
+                        Ok::<(), ArithmeticError>(())
+                    })?;
+
+                    <T as pallet_parachain_staking::Config>::Currency::transfer(
+                        &Self::account_id(),
+                        &caller,
+                        winnings,
+                        KeepAlive,
+                    ) // NOTE: If the transfer fails, the TXN get rolled back and the winnings stay in the map for claiming later
```
Dengjianping: Can you add a event here?

pallets/pallet-lottery/src/lib.rs
```diff
@@ -0,0 +1,1077 @@
...
+            // NOTE: If more than gas_reserve is in the pallet, the full excess will be paid out to the winner of the next drawing! This is intended to dope the winning balance with extra rewards
+
+            let drawing_interval = <T as Config>::DrawingInterval::get();
+            ensure!(
+                drawing_interval > 0u32.into(),
+                Error::<T>::PalletMisconfigured
+            );
+            let lottery_drawing_call: CallOf<T> = Call::draw_lottery {}.into();
+            T::Scheduler::schedule_named(
+                Self::lottery_schedule_id(),
```
Dengjianping, line 662: I'm not sure we should use the same name for every scheduled task, because if we use the same name, the previous scheduled task storage will be overwritten.

pallets/pallet-lottery/src/lib.rs
```diff
@@ -0,0 +1,1077 @@
...
+            T::LotteryPot::get().into_account_truncating()
+        }
+        /// Get an identifier for scheduling drawings from the `PotId`.
+        fn lottery_schedule_id() -> Vec<u8> {
+            T::LotteryPot::get().0.to_vec()
+        }
+        fn select_winning_balance(
+            max_winning_balance: BalanceOf<T>,
+        ) -> Result<BalanceOf<T>, Error<T>> {
+            const MAX_NUMBER_OF_RESAMPLES: u8 = 3;
```
Dengjianping, line 841: constant should not be here.

pallets/pallet-lottery/src/lib.rs
```diff
@@ -0,0 +1,1077 @@
...
+                log::debug!(
+                    "Finished unstaking {:?} out of {:?} requests",
+                    original_len - unstaking.len(),
+                    original_len
+                );
+                UnstakingCollators::<T>::put(unstaking);
+            }
+        }
+
+        #[named]
```
Dengjianping, line 1039: Interesting, I never see such macro before.

runtime/manta/src/lib.rs
```diff
@@ -331,6 +336,81 @@ impl pallet_timestamp::Config for Runtime {
...
+        if cfg!(feature = "runtime-benchmarks") {
+            // storage reads as per actual reads
+            let _relay_storage_root = ParachainSystem::validation_data();
+            let _relay_chain_state = ParachainSystem::relay_state_proof();
+            let benchmarking_babe_output = Hash::default();
+            return Some(benchmarking_babe_output);
+        }
+        relay_chain_state_proof()
+            .read_optional_entry(
+                cumulus_primitives_core::relay_chain::well_known_keys::TWO_EPOCHS_AGO_RANDOMNESS,
```
ghzlatarev: can you put a comment why two epochs ago

primitives/session-keys/src/inherent.rs
```diff
@@ -0,0 +1,80 @@
...
+//! Inherents used for randomness
+use parity_scale_codec::{Decode, Encode};
+use sp_inherents::{Error, InherentData, InherentIdentifier, IsFatalError};
+use sp_runtime::RuntimeString;
+
+#[derive(Encode)]
+#[cfg_attr(feature = "std", derive(Debug, Decode))]
+/// TODO document
+pub enum InherentError {
+    /// TODO document
```
ghzlatarev: todo: docs

pallets/pallet-lottery/src/lib.rs
```diff
@@ -0,0 +1,1077 @@
...
+        pub min_withdraw: BalanceOf<T>,
+    }
+
+    #[cfg(feature = "std")]
+    impl<T: Config> Default for GenesisConfig<T> {
+        fn default() -> Self {
+            Self {
+                min_deposit: 1u32.into(),
+                min_withdraw: 1u32.into(),
+                gas_reserve: 10_000u32.into(),
```
ghzlatarev, line 280: we should log a task to not forget to update these once we launch on both Manta and Calamari

pallets/pallet-lottery/src/lib.rs
ghzlatarev: what do you mean exactly ?

Dengjianping: Like
```rust
LotteryWinner { winner: T::AccountId, amount: BalanceOf<T> }
```

pallets/pallet-lottery/src/lib.rs
```diff
@@ -0,0 +1,1077 @@
...
+                    <frame_system::Pallet<T>>::block_number()
+                );
+                return Ok(());
+            }
+            // Match random number to winner. We select a winning **balance** and then just add up accounts in the order they're stored until the sum of balance exceeds the winning amount
+            // IMPORTANT: This order and active balances must be locked to modification after the random seed is created (relay BABE randomness, 2 epochs ago)
+            let winning_balance = Self::select_winning_balance(Self::total_pot())?;
+            let mut winner: Option<T::AccountId> = None;
+            let mut count: BalanceOf<T> = 0u32.into();
+            for (account, balance) in ActiveBalancePerUser::<T>::iter() {
```
ghzlatarev, line 937: we should make this storage item bounded because of this loop ActiveBalancePerUser

pallets/pallet-lottery/src/lib.rs
```diff
@@ -0,0 +1,1077 @@
...
+            )
+            .map_err(|_| Error::<T>::CouldNotSchedule)?;
+
+            Self::deposit_event(Event::LotteryStarted);
+            Ok(())
+        }
+
+        /// Stops the ongoing lottery and cancels the scheduled and any future drawings.
+        ///
+        /// This function cancels the scheduled drawing and cleans up bookkeeping.
```
ghzlatarev: How does it cleanup bookkeeping ? Looks like it just cancels the draw

pallets/pallet-lottery/src/lib.rs
```diff
@@ -0,0 +1,1077 @@
...
+                let mut left_overs: Vec<Request<_, _, _>> = Vec::new();
+                for request in request_vec.iter() {
+                    // Don't pay anyone still timelocked
+                    if now < request.block + <T as Config>::UnstakeLockTime::get() {
+                        left_overs.push((*request).clone());
+                        continue;
+                    }
+                    // stop paying people if we've run out of free funds.
+                    // The assumption is the collators serving these requests will
+                    // finish unstaking next round ( next lottery drawing )
```
ghzlatarev, line 1122: this doesn't stop paying though, rather continues.

pallets/pallet-lottery/src/lib.rs
```diff
@@ -0,0 +1,1077 @@
...
+                    if request.balance > funds_available_to_withdraw {
```
ghzlatarev, line 1123: we should update funds_available_to_withdraw after each successful transfer below

pallets/pallet-lottery/src/lib.rs
```diff
@@ -0,0 +1,1077 @@
...
+                    "Have {:?} requests left over after transfers",
+                    left_overs.len()
+                );
+                // Update T::WithdrawalRequestQueue if we paid at least one guy
+                if left_overs.len() != (*request_vec).len() {
+                    request_vec.clear();
+                    for c in left_overs {
+                        request_vec.push(c);
+                    }
+                }
```
ghzlatarev: if left_overs.len() != (*request_vec).len() {
    request_vec.clear();
    request_vec.append(&mut left_overs);
}

pallets/pallet-lottery/src/lib.rs
ghzlatarev: is this edge case unit tested ?

pallets/pallet-lottery/src/tests.rs
ghzlatarev, line 1: We can add more error case unit tests. Some simple ones like trying to withdraw more than deposited and trying to withdraw without having any deposit

ghzlatarev: Or depositing less than the min

pallets/pallet-lottery/src/lib.rs
Garandor: `continue` is semantically identical with "stop paying".
there may be later requests of lower balance later in the list

pallets/pallet-lottery/src/lib.rs
Garandor: thanks for this good catch. this was a fairly large oversight. moved available_balance into the loop

pallets/pallet-lottery/src/lib.rs
Garandor: that would bound the number of users, doesn't seem desirable.
If this needs to be limited then it should be done by the "min_deposit" configurable

pallets/pallet-lottery/src/lib.rs
Garandor: good point, the comment was old and misleading. a stopped lottery is now just that, a lottery that does not have drawings but all other operations continue as normal (deposit/withdraw/rewards balance accrual)

pallets/pallet-lottery/src/lib.rs
ghzlatarev: yeah you're right the min-deposit should help with this. We should maybe add a ci to check that the min-deposit doesn't get something absurdly cheap though

pallets/pallet-lottery/src/lib.rs
ghzlatarev: oh you mean stop paying the current, ok

pallets/randomness/Cargo.toml
ghzlatarev, line 1: adding vrf could be worth prioritizing in the near future

pallets/pallet-lottery/src/lib.rs
Apokalip: Shouldnt be an issue, if this is called, users would be obviously 1 or more

pallets/pallet-lottery/src/lib.rs
Garandor: thanks Georgi, this reminded me that we need to have extrinsics that can actually set these in the pallet lol

pallets/pallet-lottery/src/lib.rs
Garandor: @Apokalip could you help with implementing this pls?

Actually, this doesnt need to be done now. It's ok to log a followup

pallets/randomness/Cargo.toml
Garandor: i see no action item here, marking resolved.

pallets/pallet-lottery/src/tests.rs
Garandor: this was not done due to time constraints. I agree

pallets/pallet-lottery/src/lib.rs
Garandor: updated comment, resolving

pallets/randomness/src/lib.rs
Garandor: substituted with manta headers. resolving

pallets/pallet-lottery/src/staking.rs
Garandor: for `log::trace!(function_name!());`

pallets/pallet-lottery/src/lib.rs
Garandor: it enables `log::trace!(function_name!());`

pallets/pallet-lottery/Cargo.toml
Garandor: the code was originally forked from their 0.9.27 codebase, updated

pallets/pallet-lottery/src/lib.rs
Garandor: done thx

pallets/pallet-lottery/src/lib.rs
Garandor: cool thx @Dengjianping, i wasn't aware that syntax was possible. Changed

pallets/pallet-lottery/src/lib.rs
Garandor: oh yeah thanks, sounds like a good idea.
also this reminded me i created the Event type for `Withdrawn` but forgot to actually deposit it :sweat_smile:

pallets/pallet-lottery/src/lib.rs
Garandor: it's a constant that only impacts this `fn` and isnt needed anywhere else. where would you put it. i think its fine there

pallets/pallet-lottery/src/lib.rs
Garandor: there can not be multiple lotteries running in parallel. This usecase is simply not supported for now

pallets/pallet-lottery/src/lib.rs
Garandor: thx, changed, i had type resolution issues in earlier versions of the code

pallets/pallet-lottery/src/lib.rs
Garandor: yes it is possible but this case is handled in the first branch starting L464 so it can't happen in this `or_else` branch

pallets/pallet-lottery/src/lib.rs
Garandor: TotalUsers is only used as a weight hint anyway, no code logic is impacted by this.

Can't hurt to do it properly tho, changed

pallets/pallet-lottery/src/lib.rs
Garandor: Thx Jamie, added the comments and removed a few unused error types

pallets/pallet-lottery/src/lib.rs
Garandor: These requests are time-bounded because they will be paid out after timelock expiry, no way for users to prevent that other than by stopping the lottery or if drawings fail to execute.
The anti-spam bound on this is `min_withdraw` again so it needs to be reasonably set

pallets/pallet-lottery/src/lib.rs
Garandor: that seems clearer, thx!

pallets/pallet-lottery/src/staking.rs
Garandor: > why is the comment important

because the goal here is APY maximization. It's a minor point though so i left it out of scope.

> also maybe it would be better to just divide evenly in v1, this adds a lot of complexity

i agree, but that complexity is contained and largely provided by parachain_staking.
Open to removing it, but that would also remove one of the main selling points and thus becomes *a product decision*

pallets/pallet-lottery/Cargo.toml
Garandor: true, updated it

pallets/pallet-lottery/src/staking.rs
Garandor: only collators that have produced at least 1 block are considered active.
unless we're less than 25% into a round, then *all* selected collators are considered active.

pallets/pallet-lottery/src/staking.rs
Garandor: done

runtime/manta/src/lib.rs
Garandor: done - and changed to 1 epoch ago

primitives/session-keys/src/inherent.rs
Garandor: done

pallets/randomness/src/lib.rs
Garandor: thanks, the randomness ones were actually important. Although these unwraps should be impossible to get triggered.

pallets/pallet-lottery/src/tests.rs
ghzlatarev: Yeah totally, let's just log it. Someone else could impl to get more familiar with the code.

pallets/randomness/Cargo.toml
ghzlatarev: We could log it, that's what i meant. Seems it like it could be worth spending time on it in the future ?

pallets/pallet-lottery/src/lib.rs
ghzlatarev: Can we add this as a comment

pallets/pallet-lottery/src/staking.rs
ghzlatarev: aha ok i get it now, but isn't 1/4 of a round too long . Can't we use the number of the candidates amount of blocks into the round ?

runtime/manta/src/lib.rs
```diff
@@ -331,6 +336,85 @@ impl pallet_timestamp::Config for Runtime {
...
+            // storage reads as per actual reads
+            let _relay_storage_root = ParachainSystem::validation_data();
+            let _relay_chain_state = ParachainSystem::relay_state_proof();
+            let benchmarking_babe_output = Hash::default();
+            return Some(benchmarking_babe_output);
+        }
+        relay_chain_state_proof()
+            .read_optional_entry(
+                // We use randomness from one relaychain epoch ago (4hrs Polkadot, 1hr Kusama) in combination with a (longer) `DrawingFreezeout` in pallet lottery
+                // to prevent manuipulation of the winning set by participants of the lottery
```
ghzlatarev: nit: spelling (manuipulation)

ghzlatarev: the link is kinda broken - doesn't lead to the risks section

runtime/manta/src/lib.rs
```diff
@@ -331,6 +336,85 @@ impl pallet_timestamp::Config for Runtime {
...
+                // NOTE: We operate under the following assumption https://github.com/Manta-Network/Manta/blob/f7b1a74e0261cc6a73081a9a86f8cd87aef8ea27/pallets/randomness/README.md#L85,
```
ghzlatarev: `The only incentive to prevent the validator from skipping the block is the block rewards. So the randomness value is only economically safe if the value at stake is lower than a block reward.` Doesn't that mean that if the lottery is higher than a block reward, then the randomess is not good ?

pallets/pallet-lottery/src/lib.rs
ghzlatarev: oof yeah of course, a user requests a withdraw which fills up this queue, and it gets freed up after however many blocks, so it can't grow infinitely.

pallets/pallet-lottery/src/tests.rs
```diff
@@ -0,0 +1,407 @@
+// 202
```
ghzlatarev: this line is broken

pallets/pallet-lottery/src/lib.rs
```diff
@@ -0,0 +1,1175 @@
...
+    #[pallet::without_storage_info]
+    pub struct Pallet<T>(PhantomData<T>);
+
+    // Configurable (constant) storage items
+
+    /// NOTE: how much KMA to keep in the pallet for gas
+    /// This must be initialized at genesis, otherwise the pallet will run out of gas at the first drawing
+    #[pallet::storage]
+    #[pallet::getter(fn gas_reserve)]
+    pub(super) type GasReserve<T: Config> = StorageValue<_, BalanceOf<T>, ValueQuery>;
```
ghzlatarev, line 174: what actually reduces this ?

runtime/manta/src/lib.rs
Garandor: it means the last validator of the epoch before the drawing gets two tickets instead of one because he can choose to skip producing his block if the resulting random seed is better for him.
- This is fairly low-impact
- I've been told not to worry about fairness of the randomness algorithm itself

runtime/manta/src/lib.rs
Garandor: thx, fixed

pallets/pallet-lottery/src/tests.rs
Garandor: oh lol, fixed thx

pallets/pallet-lottery/src/staking.rs
Garandor: people can randomly drop blocks sometimes, no need to unstake if they missed their first block or two for whatever reason

pallets/pallet-lottery/src/lib.rs
Garandor: Nothing, it's a configurable constant

runtime/manta/src/lib.rs
ghzlatarev: ahaa ok i see, was thinking something else

pallets/pallet-lottery/src/staking.rs
Garandor: discussion is obsolete, i changed it to check if this collator has earned any points in the last round instead

pallets/pallet-lottery/src/tests.rs
Garandor: https://github.com/Manta-Network/Manta/issues/1184

pallets/pallet-lottery/src/staking/mod.rs
```diff
@@ -0,0 +1,344 @@
...
+        remaining_balance = remaining_balance.saturating_sub(balance_unstaked);
+
+        if !remaining_balance.is_zero() {
+            log::error!(
+                "FATAL: We have {:?} left that COULD NOT BE UNSTAKED",
+                remaining_balance
+            );
+            return vec![]; // NOTE: Ensure we do not continue with the partial withdrawal
+        }
+        // TODO: Add random fallback
```
ghzlatarev: is there an issue for this ?

pallets/pallet-lottery/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+
+            // TODO: Unstake all collators, schedule return of all user deposits
+            // for collator in collators_we_staked_to {
+            //     do_unstake(collator);
+            // }
+            // TODO: Lock everything until this process is finished
+
+            // TODO: return user deposits and pay out winnings, deposit event
+
+            // Ok(())
```
ghzlatarev, line 797: issue for this

pallets/pallet-lottery/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+                    random = T::RandomnessSource::random(&[n; 1]);
+                    log::debug!("select-winner using randomness {:?}", random);
+                    // TODO: The following check needs a change to pallet randomness but is static,
+                    //       so this can be done manually on deployment of the pallet
+                    // ensure!(
+                    //     random.1 = randomness_established_at_block
+                    //         .saturating_add(<T as Config>::DrawingFreezeout::get())
+                    //         < <frame_system::Pallet<T>>::block_number(),
+                    //     Error::<T>::PalletMisconfigured
+                    // );
```
ghzlatarev, line 872: issue for this ?

ghzlatarev (dismissed):

Apokalip (dismissed):
LGMT! Just add issues for all TODOs so we don't forget them

pallets/pallet-lottery/src/staking/mod.rs
Garandor: i'll remove this, that comment was a spur-of-the-moment thing and is not actually necessary, the previous strategy cannot actually fail

pallets/pallet-lottery/src/lib.rs
Garandor: #1185

pallets/pallet-lottery/src/lib.rs
Garandor: #1186

ghzlatarev (approved):

Apokalip (approved):

PR #937, opened 2022-12-20
Set collator minimum bond to 4M KMA

Garandor:
## Description

Calamari Functional change:
- collators below 4mil are no longer considered for collator assignment
https://github.com/Manta-Network/Manta/pull/937/files#diff-b616758683d3c1926eb58a1b16a2b439122f36e4738af354b42addc5daf2cc83R22

Testing changes:
- add regression test for 400k-bonded collators
- add `parachain_staking` to integration_test mock
- refactored integration test for `lazy_static` ALICE-FERDIE keys

relates to OR closes: #933

---

Before we can approve this PR for merge, please make sure that **all** the following items have been checked off:
- [x] Linked to Github issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [x] Added **one** label out of the `L-` group to this PR
- [x] Added **one or more** labels from the `A-` and `C-` groups to this PR
- [x] Explicitly labelled `A-calamari`, `A-dolphin` and/or `A-manta` if your changes are meant for/impact either of these (CI depends on it)
- [ ] This PR is targeted against the *current*  Milestone ( otherwise discuss if it can be added in time)
- [ ] Re-reviewed `Files changed` in the Github PR explorer.

Situational Notes:
- If adding functionality, write unit tests!
- If importing a new pallet, choose a proper module index for it, and allow it in `BaseFilter`. Ensure **every** extrinsic works from front-end. If there's corresponding tool, ensure both work for each other.
- If needed, update our Javascript/Typescript APIs. These APIs are officially used by exchanges or community developers.
- If modifying existing runtime storage items, make sure to implement storage migrations for the runtime and test them with `try-runtime`. This includes migrations inherited from upstream changes, and you can search the diffs for modifications of `#[pallet::storage]` items to check for any.
- If runtime changes, need to update the version numbers properly:
   * `authoring_version`: The version of the authorship interface. An authoring node will not attempt to author blocks unless this is equal to its native runtime.
   * `spec_version`: The version of the runtime specification. A full node will not attempt to use its native runtime in substitute for the on-chain Wasm runtime unless all of spec_name, spec_version, and authoring_version are the same between Wasm and native.
   * `impl_version`: The version of the implementation of the specification. Nodes are free to ignore this; it serves only as an indication that the code is different; as long as the other two versions are the same then while the actual code may be different, it is nonetheless required to do the same thing. Non-consensus-breaking optimizations are about the only changes that could be made which would result in only the impl_version changing.
[...]

PR #1043, opened 2023-03-14
Add self-bond filter condition when computing new set of collators

Garandor:
## Description

This will result in collators on calamari below 4M self-bond no longer being eligible to produce blocks.

- Previously, tests were not handling the 6th collator correctly, causing the fact that it wasn't offboarded in the old PR to be missed.
- A take-N-before-filter sequence bug in `parachain-staking` led to fewer-than-expected collators being selected
- Added regression test for correct offboarding

Followup to #937

---

Situational Notes:

### DIFF
--- a/pallets/parachain-staking/src/lib.rs
+++ b/pallets/parachain-staking/src/lib.rs
@@ -1688,7 +1688,14 @@ pub mod pallet {
                 .into_iter()
                 .rev()
                 .take(top_n)
-                .filter(|x| x.amount >= T::MinCollatorStk::get())
+                .filter(|x| {
+                    // Only consider collators above minimum total stake and self-bond
+                    x.amount >= T::MinCollatorStk::get()
+                        && Self::candidate_info(x.owner.clone())
+                            .expect("looping candidates. therefore canidateinfo exists. qed")
+                            .bond
+                            > T::MinCandidateStk::get()
+                })
                 .map(|x| x.owner)
                 .collect::<Vec<T::AccountId>>();
             collators.sort();
--- a/pallets/parachain-staking/src/lib.rs
+++ b/pallets/parachain-staking/src/lib.rs
@@ -1690,7 +1690,7 @@ pub mod pallet {
                 .filter(|x| {
                     // Only consider collators above minimum total stake and self-bond
                     x.amount >= T::MinCollatorStk::get()
-                        && Self::is_candidate(x.owner.clone())
+                        && Self::is_candidate(&x.owner)
                         && Self::candidate_info(x.owner.clone())
                             .expect("is_candidate => canidateinfo exists. qed")
                             .bond
--- a/pallets/parachain-staking/src/lib.rs
+++ b/pallets/parachain-staking/src/lib.rs
@@ -1689,12 +1689,13 @@ pub mod pallet {
                 .rev()
                 .filter(|x| {
                     // Only consider collators above minimum total stake and self-bond
-                    x.amount >= T::MinCollatorStk::get()
-                        && Self::is_candidate(&x.owner)
-                        && Self::candidate_info(x.owner.clone())
-                            .expect("is_candidate => canidateinfo exists. qed")
-                            .bond
-                            >= T::MinCandidateStk::get()
+                    x.amount >= T::MinCollatorStk::get() &&
+                    if let Some(info) = Self::candidate_info(x.owner.clone()) {
+                        info.bond >= T::MinCandidateStk::get()
+                    } else {
+                        log::error!("Candidate did not have CandidateInfo in storage, this should not happen");
+                        false
+                    }
                 })
                 .take(top_n)
                 .map(|x| x.owner)
--- a/pallets/parachain-staking/src/lib.rs
+++ b/pallets/parachain-staking/src/lib.rs
@@ -1694,7 +1694,7 @@ pub mod pallet {
                         && Self::candidate_info(x.owner.clone())
                             .expect("looping candidates. therefore canidateinfo exists. qed")
                             .bond
-                            > T::MinCandidateStk::get()
+                            >= T::MinCandidateStk::get()
                 })
                 .map(|x| x.owner)
                 .collect::<Vec<T::AccountId>>();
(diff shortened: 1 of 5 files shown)


pallets/parachain-staking/src/lib.rs
```diff
@@ -1687,8 +1687,15 @@ pub mod pallet {
             let mut collators = candidates
                 .into_iter()
                 .rev()
+                .filter(|x| {
+                    // Only consider collators above minimum total stake and self-bond
+                    x.amount >= T::MinCollatorStk::get()
+                        && Self::candidate_info(x.owner.clone())
+                            .expect("looping candidates. therefore canidateinfo exists. qed")
```
ferrell-code: Note that this is looping through `CandidatePool`, but then reads `CandidateInfo`. 

So maybe error handle this. It maybe an invariant state that shouldn't happen, but its not proven to be impossible

Garandor: > but its not proven to be impossible 

It kinda is, an address is only added to CandidatePool after its CandidateInfo is created. Check the `fn join_candidates`
This invariant is used in multiple places throughout parachain_staking in the same manner

Garandor: Also, this filter and `fn` do not propagate a `Result`, so i see no easy way to handle this

runtime/calamari/tests/integrations_mock/integration_tests.rs
```diff
@@ -604,6 +604,73 @@ fn collator_cant_join_below_standard_bond() {
...
+                assert_ok!(ParachainStaking::delegate(
+                    Origin::signed(USER.clone()),
+                    collator,
+                    100_000_000 * KMA,
+                    50,
+                    50
+                ));
+            }
+
+            // Ensure ALICE is not selected despite having a large stake although it has a large total stake
```
ghzlatarev: This wording doesn't make sense to me.

pallets/parachain-staking/src/lib.rs
ferrell-code: What about `schedule_leave_candidate` seems to put it in a state that could panic?

You could default to zero instead of expect:

```rust
let mut collators = candidates
                .into_iter()
                .rev()
                .filter(|x| {
                    // Only consider collators above minimum total stake and self-bond
                    x.amount >= T::MinCollatorStk::get()
                        && Self::candidate_info(x.owner.clone())
                            .map(|x| x.bond)
                            .unwrap_or(Zero::zero())
                            >= T::MinCandidateStk::get()
                })
```

ghzlatarev: Check out what ChatGPT is suggesting:

Thank you for providing more context. Based on the code you linked, it appears that candidate_info() retrieves an item from a Vec<Option<...>> and returns it. The Vec is a storage item that stores the candidate information for the staking pallet.

When items are added to the Vec, they are first added as None values, and then updated later with the actual candidate information. The expect() call in question is used to unwrap the Option value returned by candidate_info(), but only after it has been updated with the candidate information.

Given this context, the use of expect() may be justified in this case. Since the code first adds an entry to the Vec as None, and then later updates it with the candidate information, it's possible that the candidate_info() function could be called before the item has been updated. In that case, candidate_info() would return None, and the expect() call would panic.

However, assuming that the item is always updated before it is accessed by candidate_info(), the expect() call may be unnecessary. It may be better to use a match or if let statement to handle the Option value returned by candidate_info() more robustly, especially if the program needs to continue running in the event that the candidate information is not available.

Garandor: I'm in camp 
> Given this context, the use of expect() may be justified in this case.

Also, using if let or match would lead to 
```
if let Some(Self::candidate_info(x.owner.clone()))
{ do stuff}
else
{ panic!() }
```
Which is identical.
~~candidate info not existing is *not* a situation that can be handled gracefully but `parachain_staking`~~
Actually, it might be better not to panic here upon more thinking. Returning empty vec will let the chain continue producing blocks with the old set of collators potentially hiding a problem but at least it won't halt.

It'll panic in other parts of the code if that happens anyway, but at least we don't have to add to this

runtime/calamari/tests/integrations_mock/integration_tests.rs
Garandor: fixed, thx

pallets/parachain-staking/src/lib.rs
Garandor: > What about `schedule_leave_candidate` seems to put it in a state that could panic?

that code removes the entry from `CandidatePool`, not `CandidateInfo`

Your alternative looks nice but hides the error.
Also, missing CandidateINfo is a **failure**, setting bond to 0 can still include that collator if and when MinCandidateStk is set to 0

Garandor: I added an explicit `is_candidate` check that isn't needed but provides clarity in preventing the panic as a compromise

ferrell-code: Ya agreed maybe this is better:

```rust
            let mut collators = candidates
                .into_iter()
                .rev()
                .filter(|x| {
                    // Only consider collators above minimum total stake and self-bond
                    x.amount >= T::MinCollatorStk::get()
                        && if let Some(info) = Self::candidate_info(x.owner.clone()) {
                            info.bond >= T::MinCandidateStk::get()
                        } else {
                            log::warn("Candidate did not have CandidateInfo in storage, this should not happen")
                            false
                        }
                })
```

ferrell-code (dismissed):

pallets/parachain-staking/src/lib.rs
Garandor: Ok, took your suggestion since y'all seem to *really* dislike `expect`s ;)

ghzlatarev (dismissed):

ferrell-code (dismissed):
seems clippy needs a older nightly version

pallets/parachain-staking/src/lib.rs
ferrell-code: I identify as a qed hardliner :p

ferrell-code (dismissed):

ghzlatarev (dismissed):

ferrell-code (approved):

ghzlatarev (approved):

PR #1047, opened 2023-03-18
Transaction fees bump

ghzlatarev:
## Description

---

Situational Notes:


runtime/calamari/src/fee.rs
```diff
@@ -31,30 +26,6 @@ pub const FEES_PERCENTAGE_TO_TREASURY: u8 = 45;
...
-    fn polynomial() -> WeightToFeeCoefficients<Self::Balance> {
-        // Refer to the congested_chain_simulation() test for how to come up with the coefficient.
-        smallvec![WeightToFeeCoefficient {
-            coeff_integer: 5000u32.into(),
-            coeff_frac: Perbill::zero(),
-            negative: false,
-            degree: 1,
-        }]
-    }
-}
```
ghzlatarev, line 56: This basically expands to constant mutliplier so just using the ConstantMultiplier type

runtime/calamari/src/fee.rs
```diff
@@ -125,133 +65,119 @@ mod multiplier_tests {
...
-    fn congested_chain_simulation() {
-        // Configure the target cost depending on the current state of the network.
-        let target_daily_congestion_cost_usd = 250000;
+    #[ignore]
+    fn multiplier_growth_simulator_and_congestion_budget_test() {
+        let target_daily_congestion_cost_usd = 250_000;
         let kma_price = fetch_kma_price().unwrap();
         println!("KMA/USD price as read from CoinGecko = {kma_price}");
         let target_daily_congestion_cost_kma =
-            (target_daily_congestion_cost_usd as f32 * kma_price * KMA as f32) as u128;
```
ghzlatarev, line 145: this was wrong

runtime/calamari/src/fee.rs
```diff
@@ -125,133 +65,119 @@ mod multiplier_tests {
...
+
+        let mut should_fail = false;
+        while multiplier <= Multiplier::from_u32(1) {
+            t.execute_with(|| {
+                // Give the attacker super powers to not pay tx-length fee
+                // The maximum length fo a block is 3_670_016 on Calamari
+                let len = 0;
+                // Imagine this tx was called. This means that he will pay a single base-fee,
+                // which is additional super powers for the attacker
+                let fee = TransactionPayment::compute_fee(len, &info, 0);
```
ghzlatarev, line 113: We're giving the attacker some super powers to pay almost no length and base fees.

runtime/calamari/src/lib.rs
```diff
@@ -371,13 +370,14 @@ impl pallet_balances::Config for Runtime {
 }
 
 parameter_types! {
-    /// Relay Chain `TransactionLengthToFeeCoeff` / 10
-    pub const TransactionLengthToFeeCoeff: Balance = mKMA / 100;
+    /// Relay Chain `TransactionLengthToFeeCoeff` / 10 (1_000_000 on Kusama)
+    pub const TransactionLengthToFeeCoeff: Balance = mKMA;
+    pub const WeightToFeeCoeff: Balance = 150_000;
```
ghzlatarev: This is a 30x increase from before and is one of the changes that makes a significant difference. Users will feel this directly eg: transfers will increase from ~0.4KMA to ~18KMA

runtime/common/src/lib.rs
```diff
@@ -66,11 +66,11 @@ parameter_types! {
     pub const TargetBlockFullness: Perquintill = Perquintill::from_percent(25);
     /// The adjustment variable of the runtime. Higher values will cause `TargetBlockFullness` to
     /// change the fees more rapidly.
-    pub AdjustmentVariable: Multiplier = Multiplier::saturating_from_rational(3, 100_000);
+    pub AdjustmentVariable: Multiplier = Multiplier::saturating_from_rational(125, 1_000_000);
     /// Minimum amount of the multiplier. This value cannot be too low. A test case should ensure
     /// that combined with `AdjustmentVariable`, we can recover from the minimum.
     /// See `multiplier_can_grow_from_zero`.
-    pub MinimumMultiplier: Multiplier = Multiplier::saturating_from_rational(1, 1_000_000u128);
+    pub MinimumMultiplier: Multiplier = Multiplier::saturating_from_rational(1, 5u128);
```
ghzlatarev: On Kusama this is 1/10. This increase makes a significant difference.

runtime/calamari/src/lib.rs
```diff
@@ -371,13 +370,14 @@ impl pallet_balances::Config for Runtime {
 }
 
 parameter_types! {
-    /// Relay Chain `TransactionLengthToFeeCoeff` / 10
-    pub const TransactionLengthToFeeCoeff: Balance = mKMA / 100;
+    /// Relay Chain `TransactionLengthToFeeCoeff` / 10 (1_000_000 on Kusama)
+    pub const TransactionLengthToFeeCoeff: Balance = mKMA;
```
ghzlatarev: This 100x makes almost no difference. Even if we I add the maximum length in the test.

runtime/common/src/lib.rs
```diff
@@ -66,11 +66,11 @@ parameter_types! {
     pub const TargetBlockFullness: Perquintill = Perquintill::from_percent(25);
     /// The adjustment variable of the runtime. Higher values will cause `TargetBlockFullness` to
     /// change the fees more rapidly.
-    pub AdjustmentVariable: Multiplier = Multiplier::saturating_from_rational(3, 100_000);
+    pub AdjustmentVariable: Multiplier = Multiplier::saturating_from_rational(125, 1_000_000);
```
ghzlatarev: With this it takes 2.3844445 days to reach 1x multiplier and 3.41125 days to reach 2x
To cooldown from 1x it takes ~7 days and from 2x it takes ~10 days

1 day cost is 320k and to reach 1x it would cost ~1 million USD

runtime/calamari/src/lib.rs
ferrell-code: could do a 10x rather than 30x and rely more on the `AdjustmentVariable` to prevent attacks

ferrell-code (commented):
Changes look reasonable to me

runtime/common/src/lib.rs
Garandor: this is with the minimum multiplier set to 0.25, right?
These reaction times seem reasonable to me

I do feel the absolute number is too high tho. with KMA price rising, maybe we should target 100k / day at current price.
That would imply a lower minimum factor. The adjustment *speed* seems ok if not a bit too slow, we're talking about a fully congested chain here after all, so maybe a faster increase&cooldown would be better.
e.g. from 0.1x -> 1x in 1 day, 2x in 2 days with cooldown in 2 / 4 days respectively

ghzlatarev: I think I have better numbers now, with some data to back it up. I'll update and organize a meeting later today or tomorrow to review.

runtime/dolphin/src/lib.rs
```diff
@@ -360,11 +359,12 @@ impl pallet_balances::Config for Runtime {
 parameter_types! {
     /// Relay Chain `TransactionLengthToFeeCoeff` / 10
     pub const TransactionLengthToFeeCoeff: Balance = mDOL / 100;
+    pub const WeightToFeeCoeff: Balance = 150_000;
```
Garandor: why is this higher on DOL than on KMA?

ghzlatarev: leftover from testing, let me revert.

runtime/calamari/src/fee.rs
```diff
@@ -125,133 +65,123 @@ mod multiplier_tests {
...
-        run_with_system_weight(target * 101 / 100, || {
-            let next = SlowAdjustingFeeUpdate::<Runtime>::convert(minimum_multiplier);
-            assert!(
-                next > minimum_multiplier,
-                "{next:?} !>= {minimum_multiplier:?}"
-            );
-        })
+        println!(
+            "It will take {:?} days to reach multiplier of 1",
+            (blocks * 12) as f32 / (60 * 60 * 24) as f32,
```
Garandor: Parametrize the block time in all tests, otherwise this will break with async backing

runtime/calamari/src/fee.rs
```diff
@@ -125,133 +65,123 @@ mod multiplier_tests {
...
                 multiplier = next;
+
+                println!("block = {} / multiplier {:?}", blocks, multiplier);
             });
             blocks += 1;
-            println!("block = {blocks} multiplier {multiplier:?}");
+        }
+
+        let cooldown_target = 10.0f32;
+        let days = (blocks * 12) as f32 / (60 * 60 * 24) as f32;
```
Garandor: Better use the `1 * DAYS` constant

runtime/calamari/src/fee.rs
```diff
@@ -125,133 +65,123 @@ mod multiplier_tests {
...
-        // the weight is 1/100th bigger than target.
-        run_with_system_weight(target * 101 / 100, || {
-            let next = SlowAdjustingFeeUpdate::<Runtime>::convert(minimum_multiplier);
-            assert!(
-                next > minimum_multiplier,
-                "{next:?} !>= {minimum_multiplier:?}"
-            );
-        })
+        println!(
+            "It will take {:?} days to reach multiplier of 1",
```
Garandor: the loop above breaks after a day even if multiplier is not yet one. Either this comment or the loop condition is wrong

ghzlatarev: The loop condition is correct, but i forgot to update this. It needs to be the blocks at 1 day.

Garandor: sry but this makes no sense, this will always print
`It will take 1 days to reach multiplier of 1`
because your blocks_at_one_day = 7200

Garandor: or if the multiplier crosses 1 first, it will be `0` which is also incorrect

Garandor: Just change
`        while multiplier <= Multiplier::from_u32(1) || blocks <= 7200 {`
to `        while multiplier <= Multiplier::from_u32(1) {`
and leave the rest as it originally was

ghzlatarev: Yeah that was a brain fart.
But the test needs to wait for both reaching 1 multiplier and 7200 days in order to report what it needs.
Just need to make that pretty.

Garandor: i think in that case `while multiplier <= Multiplier::from_u32(1) && blocks <= 7200 {` is what youre looking for

Garandor: oh mea culpa, i read the loop condition wrong, it looks correct to leave it as you originally had it you're right.

Garandor (dismissed):
LGTM!

stechu (dismissed):

ferrell-code (approved):

stechu (approved):

Garandor (approved):

PR #1055, opened 2023-03-29
Allow xcm-transacts from all chains

ghzlatarev:
## Description

* Allow xcm Transact instructions to be executed
* Added barrier that makes sure such xcm messages have DescenOrigin first so they and pays for their own execution via WithdrawAsset|BuyExecution with an address that is not the sovereign of the sender chain. 
* That address is created by hashing the origin of the sender location so it should be unique for each address on each chain. 

---

Situational Notes:


runtime/manta/src/xcm_config.rs
```diff
@@ -153,9 +156,9 @@ match_types! {
 pub type Barrier = (
     // Allows local origin messages which call weight_credit >= weight_limit.
     TakeWeightCredit,
-    // Allows non-local origin messages, for example from from the xcmp queue,
-    // which have the ability to deposit assets and pay for their own execution.
-    AllowTopLevelPaidExecutionFrom<Everything>,
```
zqhxuyuan, line 158: Why remove this barrier?

ghzlatarev: oh good catch, this was a bad merge

zqhxuyuan (approved):

ferrell-code (approved):

PR #1057, opened 2023-03-30
Add Evm Allowlist Feature to SBT Pallet

ferrell-code:
## Description

Adds evm allowlist feature. This gives a privileged account the ability to allowlist eth based addresses to mint a free zkSBT (still needs to pay for tx fee). 

Minting free zkSBT requires user to generate a signature by signing the valid zkp with their eth wallet. The eth address used to sign must be on allowlist for this signature to be considered valid.

---

Situational Notes:


pallets/manta-sbt/src/lib.rs
```diff
@@ -284,6 +351,116 @@ pub mod pallet {
...
+                EvmAddressType::Bab(eth_address) | EvmAddressType::Galxe(eth_address) => {
+                    ensure!(eth_address == address, Error::<T>::BadSignature);
+                }
+            }
+
+            let mint_status =
+                EvmAddressAllowlist::<T>::get(address_type).ok_or(Error::<T>::NotAllowlisted)?;
+            let asset_id = match mint_status {
+                MintStatus::Available(asset) => asset,
+                MintStatus::AlreadyMinted => return Err(Error::<T>::NotAllowlisted.into()),
```
zqhxuyuan: This error type should be `AlreadyMinted`

pallets/manta-sbt/src/lib.rs
```diff
@@ -113,7 +176,7 @@ pub mod pallet {
 
     /// The module configuration trait.
     #[pallet::config]
-    pub trait Config: frame_system::Config {
+    pub trait Config: frame_system::Config + pallet_timestamp::Config {
```
Dengjianping: Seems you can add a new associated type to `Config` instead of inheriting `pallet_timestamp`
```rust
type Now: frame_support::traits::Time;
```

pallets/manta-sbt/src/lib.rs
```diff
@@ -146,6 +216,20 @@ pub mod pallet {
...
+
+    /// Allowlist for Evm Accounts
+    #[pallet::storage]
+    pub(super) type EvmAddressAllowlist<T: Config> =
+        StorageMap<_, Blake2_128Concat, EvmAddressType, MintStatus, OptionQuery>;
+
+    /// Range of time at which evm mints for each `MintType` are possible.
+    #[pallet::storage]
+    pub(super) type MintTimeRange<T: Config> =
+        StorageMap<_, Blake2_128Concat, MintType, (T::Moment, Option<T::Moment>), OptionQuery>;
```
Dengjianping: Not sure parachain based timestamp is safe enough here.

pallets/manta-sbt/src/lib.rs
```diff
@@ -284,6 +351,116 @@ pub mod pallet {
...
+
+        /// Sets the time range of which a `MintType` will be valid. Requires `AdminOrigin`
+        #[pallet::call_index(5)]
+        #[pallet::weight(<T as pallet::Config>::WeightInfo::set_mint_time())]
+        #[transactional]
+        pub fn set_mint_time(
+            origin: OriginFor<T>,
+            mint_type: MintType,
+            start_time: T::Moment,
+            end_time: Option<T::Moment>,
```
Dengjianping: Do we need to add one line of code for checking ` if end_time.is_some() { end_time > start_time }` though it's a governance call?

pallets/manta-sbt/src/lib.rs
```diff
@@ -602,6 +821,151 @@ where
...
+    fn verify_eip712_signature(proof: &Proof, sig: &[u8; 65]) -> Option<EvmAddress> {
+        let msg = Self::eip712_signable_message(proof);
+        let msg_hash = keccak_256(msg.as_slice());
+
+        recover_signer(sig, &msg_hash)
+    }
+
+    /// Eip-712 message to be signed
+    #[inline]
+    fn eip712_signable_message(proof: &Proof) -> Vec<u8> {
```
Dengjianping: Seems we can use a fixed size of array here.

pallets/manta-sbt/src/lib.rs
ferrell-code: I tried that but I had trouble getting the generic `UnixTime::Moment` to work.

Ultimately it seems unlikely we will ever use anything other than `pallet_timestamp`, so I did this. 

I am cool if you want to try and push that fix, I agree it is better to be `UnixTime`, but I couldn't get it to compile

Dengjianping: I tried it locally, seems we can get rid of pallet inheriting.
```rust
pub type Moment<T> = <<T as Config>::Timestamp as frame_support::traits::Time>::Moment;
```

ferrell-code: Nice thanks!

pallets/manta-sbt/src/lib.rs
ferrell-code: Yeah I have that question too. How possible is it to get a grossly wrong timestamp included on-chain. Only other option would be blocknumber

pallets/manta-sbt/src/lib.rs
ferrell-code: Yeah a check can be added, would just cause `mint_sbt_eth`  to always fail for that mint type. 

still an invariant that we can block from going on-chain

pallets/manta-sbt/src/lib.rs
ferrell-code: Not sure there is an easy way to avoid allocating the vec.

Could do this but in practice I don't think there is any tangible benefit. Maybe you could share a snippit if you have an idea of best way to do it:

```rust
    /// Eip-712 message to be signed
    #[inline]
    fn eip712_signable_message(proof: &Proof) -> [u8; 66] {
        let domain_separator = Self::evm_account_domain_separator();
        let payload_hash = Self::evm_account_payload_hash(proof);

        let mut msg = b"\x19\x01".to_vec();
        msg.extend_from_slice(&domain_separator);
        msg.extend_from_slice(&payload_hash);
        msg.try_into().expect("bla")
    }
```

pallets/manta-sbt/src/lib.rs
zqhxuyuan: We don't need very precise timestamp onchain. one or two block time error is fine.

pallets/manta-sbt/src/lib.rs
Dengjianping: Agreed, there's almost no benefit we can get.

pallets/manta-sbt/src/lib.rs
```diff
@@ -98,6 +123,57 @@ const ENCODED_ONE: [u8; 16] = 1u128.to_le_bytes();
...
+/// zkSBT mint Status of `EvmAddressType`. This has flag `AlreadyMinted` to put into storage after succesful mint
+#[derive(Copy, Clone, PartialEq, Eq, Encode, Decode, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+pub enum MintStatus {
+    Available(StandardAssetId),
+    AlreadyMinted,
+}
+
+#[derive(Clone, PartialEq, Eq, Encode, Decode, RuntimeDebug, TypeInfo)]
+#[scale_info(skip_type_params(T))]
+pub struct Metadata<T: Config> {
```
Dengjianping: maybe this name is kind of general, we already have metadata like asset metadata, runtime metadata.

zqhxuyuan: SbtMetadata? IMO, this `Metadata` only exist in this pallet, so it wouldn't make confuse.

runtime/calamari/src/assets_config.rs
```diff
@@ -194,5 +194,11 @@ impl pallet_manta_sbt::Config for Runtime {
     type MintsPerReserve = ConstU16<5>;
     type ReservePrice = ConstU128<{ 100_000 * KMA }>;
     type SbtMetadataBound = ConstU32<300>;
+    type ChainId = ConstU64<2084>;
```
zqhxuyuan: This chainId configuration will not work when user use Metamask to sign message, because Metamask need to make sure the chainId is evm based. 
- Make this as dynamic configuration set by AdminOrigin for different MintType, because different evm chain has different chainId. 
- ~~But seems the chainId do not play important role in the verification process, maybe we can set it to Ethereum mainnet chainId. And we're suppose user should connect to Ethereum mainnet.~~

Dengjianping:
Like `ReservePrice`, which can be configured onchain, but we can do it in another pr.

pallets/manta-sbt/src/lib.rs
```diff
@@ -98,6 +123,66 @@ const ENCODED_ONE: [u8; 16] = 1u128.to_le_bytes();
...
+/// Enum with each possible type of allowlisted Eth Address
+#[derive(Copy, Clone, PartialEq, Eq, Encode, Decode, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+pub enum EvmAddressType {
+    /// BAB holder
+    Bab(EvmAddress),
+    /// Galxe holder
+    Galxe(EvmAddress),
+}
+
+/// Different mint types
```
ghzlatarev, line 141: can you add a comment about the differences.

runtime/calamari/src/assets_config.rs
```diff
@@ -194,5 +194,10 @@ impl pallet_manta_sbt::Config for Runtime {
     type MintsPerReserve = ConstU16<5>;
     type ReservePrice = ConstU128<{ 100_000 * KMA }>;
     type SbtMetadataBound = ConstU32<300>;
+    type AdminOrigin = EitherOfDiverse<
+        EnsureRoot<AccountId>,
+        pallet_collective::EnsureMembers<AccountId, TechnicalCollective, 2>,
```
ghzlatarev: Maybe 2/3 is better than flat 2, in case we add people to the tech committee ?

pallets/manta-sbt/src/lib.rs
ferrell-code: added comments below for each type

ghzlatarev (approved):

zqhxuyuan (approved):

PR #1062, opened 2023-04-04
wip: Don't let user provide `min_balance` or `is_sufficient` as parameters to `update_asset_metadata`

Garandor:
## Description

Alternative 2 (see #1059) - prevent changing these 

---

- [ ] Added **one** label out of the `L-` group to this PR
- [ ] Added **one or more** labels from the `A-` and `C-` groups to this PR

Situational Notes:


pallets/asset-manager/src/lib.rs
```diff
@@ -461,11 +461,14 @@ pub mod pallet {
                 Error::<T>::UpdateNonExistentAsset
             );
             <T::AssetConfig as AssetConfig<T>>::AssetRegistry::update_asset_metadata(
-                &asset_id,
-                metadata.clone().into(),
+                &asset_id, metadata,
             )?;
-            AssetIdMetadata::<T>::insert(asset_id, &metadata);
```
Garandor, line 467: All tests are passing but I'm 90% sure it's *not* safe to skip this insert. The `Assets` metadata got updated above and it never seems to be used from here (not sure why we store the `AssetStorageMetadata` anyway since we never do anything with it outside of `Assets` code) but frontend code might depend on `asset_id_metadata` returning the correct values somewhere

The interface has no mutators unfortunately, so not sure how to update the field otherwise

PR #1064, opened 2023-04-06
add chain id

zqhxuyuan:
## Description

---

Situational Notes:


ghzlatarev (dismissed):

pallets/manta-sbt/src/lib.rs
```diff
@@ -434,9 +435,10 @@ pub mod pallet {
             // check that mint type is within time window
             Self::check_mint_time(&chain_info)?;
 
-            let address =
-                Self::verify_eip712_signature(&post.proof, &eth_signature, chain_info.chain_id)
-                    .ok_or(Error::<T>::BadSignature)?;
+            let chain_id = chain_id.unwrap_or(chain_info.chain_id);
```
Dengjianping: What the difference between `chain_id` and `chain_info.chain_id`?

zqhxuyuan: chain_id on the method is as parameters. if not passing chain_id, then use chain_info's chain_id.

Dengjianping: What if user inputs `chain_id` and `chain_info.chain_id` both but different, will it affect line 441? I mean will the signature be valid?

zqhxuyuan: as long as the chainId when user sign the message match user input's chain id, the signature is valid.

Garandor: I'm not clear on when it makes sense to sign from a different chain ID than the one set by the minting admin.

Like, why would i sign a zkp from BSC when the mint is meant for Galxe - and why should we accept a signature like that?

zqhxuyuan: The signature is only used for eth address verification. If we set chainId fixed to admin setting. When user change to other network which is almost frequency operation, their signature will be invlid, they'll need to switch back to BSC chain and sign again, this is not a good UE.

zqhxuyuan: If user first is not connect to BSC(and also our admin setting is BSC chain id), frontend will tip the user to switch to BSC. But later on he can still switch to ETH, in this time, if he sign the message, we still allow the signature valid, frontend will use ETH chain id. That's why we add chain_id as parameter. For UE, we don't want user have to switch back to BSC to sign the valid signature.

zqhxuyuan: maybe remove the admin setting, and only accept chain id as parameters, that will not cause any confuse?

ferrell-code: From a security perspective this is fine as the signature can never be successfully replayed.

Although IMO it is a little weird to allow an signature from ETH chain_id to pass verification for a BSC item

Dengjianping (dismissed):

ferrell-code (dismissed):
I suppose it is more of a product decision whether we should accept any chain_id. This is fine on runtime side

ferrell-code (approved):

Garandor:
~~The BAB owner generates a ZKP, signs it from his BSC account and makes an SBT.~~
~~Now he gives the ZKP to a friend, who signs it from his ETH account and gets the same SBT (e.g. by manually calling the backend if the frontend doesn't allow him to).~~
prevented by the eip712 signature, so attacker must hold the corresponding private key

~~or~~

~~a fake frontend tricks the BAB owner into generating the ZKP and then provides a signature from some fake chain ID where the attacker controls the BAB owner's public address to steal the minted SBT from its owner.
You can add custom networks to metamask, so creating a fake chain like that which will provide a valid signature shouldn't be hard~~
~~Or even better, a fake chain that just send `true` on any signing request for any account.~~
prevented by it having to be a valid eip712 signature, so attacker must hold the corresponding private key

~~What makes this impossible?~~


> maybe remove the admin setting, and only accept chain id as parameters, that will not cause any confuse?

i like this. the `chain_info.chainid` setting is meaningless in this case. just make the parameter non-`Option`

Garandor (approved):

PR #1080, opened 2023-04-16
Rust integration tests for Manta runtime

### LINKED ISSUE #1070: Rust integration tests for Manta runtime
Similar to the Calamari integration tests.

Dengjianping:
## Description

1. Created new crate `integration-tests` in runtime folder, that will be only be compiled when run integration tests.
1. Introduced two features to control which runtime runs integration tests, just like this:
```sh
cargo test -p integration-tests --release --features=calamari --no-default-features
cargo test -p integration-tests --release --features=manta --no-default-features
```
So that minimized redundant test code.

3. Ignored `6` parachain-staking integration tests.

Closes #1070 
---

Situational Notes:


runtime/manta/tests/mod.rs
ghzlatarev, line 1: what about the calamari xcm tests ?

ghzlatarev (commented):
I think we also need an additional issue to add missing tests.
Also maybe somehow move all tests to a common module with tests that are generic by the runtime.

runtime/manta/tests/mod.rs
Dengjianping: I found some different xcm configures for each runtime xcm tests, so I separated them by features `--features=calamari` and `--features=manta`.
All xcm tests will be run for each runtime.

Dengjianping:
> I think we also need an additional issue to add missing tests. Also maybe somehow move all tests to a common module with tests that are generic by the runtime.

You mean moving common integration tests for common crate?

.github/workflows/check_tests.yml
```diff
@@ -82,6 +82,16 @@ jobs:
...
+      - name: Run Integration Tests
+        env:
+          RUST_BACKTRACE: full
+          RUSTC_WRAPPER: sccache
+          SCCACHE_CACHE_SIZE: 2G
+          SCCACHE_DIR: /home/runner/.cache/sccache
+        run: |
+          source ${HOME}/.cargo/env
+          RUSTC_BOOTSTRAP=1 cargo test -p integration-tests --release --features=calamari --no-default-features
+          RUSTC_BOOTSTRAP=1 cargo test -p integration-tests --release --features=manta --no-default-features
```
Garandor, line 98: this will compile the whole thing twice, could you run it `--features=manta,calamari`?

Garandor (commented):
Foders `integrations_mock` and `xcm_mock` contain the tests along the mocks, so they should be renamed to
`local` and `xcm`.

I would also rename `calamari.rs` and `manta.rs` to something like `test_manta.rs`

Love the idea to split into features :thumbsup:

.github/workflows/check_tests.yml
ferrell-code: Don't think that's possible cause in xcm-test it uses 

```rust 
if #[feature = "calamari"] {
...
} else {
...
}
```

runtime/integration-tests/src/integrations_mock/mod.rs
```diff
@@ -102,3 +137,18 @@ pub fn initialize_collators_through_whitelist(collators: Vec<AccountId>) {
         candidate_count
     ));
 }
+
+#[allow(dead_code)]
+fn seal_header(mut header: Header, author: AccountId) -> Header {
```
ferrell-code, line 142: why do we need this?

runtime/integration-tests/src/lib.rs
```diff
@@ -14,7 +14,5 @@
 // You should have received a copy of the GNU General Public License
 // along with Manta.  If not, see <http://www.gnu.org/licenses/>.
 
-#![cfg(test)]
```
ferrell-code: keep this?

ferrell-code (commented):
some nits, very nice!

.github/workflows/check_tests.yml
Garandor: hm, i guess that's the cost of having these in unified code then.
I guess we can live with that, maintainability wins out.
In this case however, we could do the same thing with the integration tests, unify them in a file and enable each testcase with the feature flag

Dengjianping: I added part to here, just because there's a lint check running `SKIP_WASM_BUILD=1 cargo check --no-default-features`, `no-default-features` will cause compilation error without this part of code.
https://github.com/Manta-Network/Manta/blob/manta/.github/workflows/run_linters.yml#L97

ghzlatarev (dismissed):
I think we also need an additional issue to add missing tests.
Also maybe somehow move all tests to a common module with tests that are generic by the runtime.

ferrell-code (dismissed):

runtime/integration-tests/src/integrations_mock/test_common.rs
```diff
@@ -0,0 +1,985 @@
...
+    },
+    VersionedMultiLocation,
+};
+use xcm_executor::traits::WeightBounds;
+
+use super::{ALICE, ALICE_SESSION_KEYS};
+use sp_core::sr25519;
+use sp_runtime::{DispatchError, ModuleError};
+
+// currently, we ignore all parachain staking tests in integration tests
```
ferrell-code, line 56: Are these still being called somewhere else?

Dengjianping: Do you mean these imports? or the ignored test cases?

ferrell-code: Ignored tests

Dengjianping: No, they're not being called, @Garandor  told me that these parachain-staking test cases doesn't make sense here with no finalized related staking params, so I followed his advice to ignore them.

ferrell-code: Ok , could add feature flags to separate calamari and manta if we use different parameters.

Perhaps a future PR

Dengjianping: Sure.

ghzlatarev (approved):

zqhxuyuan (approved):
👍🏻

PR #1083, opened 2023-04-21
tx fees diff support

### LINKED ISSUE #807: Automate Weight Comparison previous->current release
We are currently comparing weight files manually on each release.

Write a small tool that can
- read a weights file
- execute each of the `fn`s contained to obtain their weight output
- compare these with the weight outputs from a second weight file

Fail if any of the detected Weights differ by more than some significance barrier, e.g. 30% to alert a dev to check these manually or there are weights missing from one of the two files

Dengjianping:
## Description

Closes #807

With this pr merge, we will no need to look at new weights when they're updated.
Maybe it's not a smart implementation, at least, we can get rid of reviewing weight numbers.

- The code about estimating all tx fees locates at `diff_gas_fees.rs`. All computations base on actual [fee multiplier](https://github.com/Manta-Network/Manta/blob/manta/runtime/common/src/lib.rs#L76).
- Whatever the extrinsic are open to community or not, I added them to `diff_gas_fees.rs`.
- All tx fees history locates at `tx-fees-data`.
- If any tx fees fluctuates over **10%**, the test case `diff_gas_fees::diff_gas_fees` will fail.
---

Situational Notes:


runtime/calamari/src/diff_gas_fees.rs
```diff
@@ -0,0 +1,2166 @@
...
+            .unwrap()
+            .path()
+            .file_name()
+            .unwrap()
+            .to_str()
+            .unwrap()
+            .split("-tx-fees.csv")
+            .collect::<Vec<&str>>()[0]
+            .to_string();
+        let _version = version_compare::Version::from(&_version).unwrap();
```
ghzlatarev: this unwrap is failing

runtime/calamari/src/diff_gas_fees.rs
ghzlatarev, line 1: can you add a line in the release template to update this file if there are new weights.

runtime/calamari/src/diff_gas_fees.rs
Dengjianping: I just committed a fix, should be fixed.

runtime/calamari/src/diff_gas_fees.rs
Dengjianping: sure

runtime/calamari/tx-fees-data/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+- The code about estmating all tx fees locates at [`diff_gas_fees.rs`](../src/diff_gas_fees.rs).
+- All tx fees history locates at [`tx-fees-data`](.).
+- If any extrinsic fluctuates over **10**, the test case `diff_gas_fees::diff_gas_fees` will fail.
+
+## Generate latest tx fees
+
+Run the command.
+```sh
+
```
ferrell-code: extra line

runtime/calamari/src/diff_gas_fees.rs
```diff
@@ -0,0 +1,2169 @@
...
+    let all_extrinsics_gas_fees = calculate_all_current_extrinsic_gas_fee();
+
+    for extrinsic in all_extrinsics_gas_fees {
+        wtr.serialize(extrinsic).unwrap();
+    }
+    wtr.flush().unwrap();
+}
+
+fn calculate_all_current_extrinsic_gas_fee() -> Vec<GasFeeDetail> {
+    let multiplier = MinimumMultiplier::get();
```
ferrell-code: The diff is going to be small if we use `MinimumMultiplier` as most of the cost is in base fee

Dengjianping: It's not just base fee, this line of code will calculate the whole tx fee according to current `MinimumMultiplier`
```rust
let fee = TransactionPayment::compute_fee(call_len, &dispatch_info, 0);
```
If I'm wrong, @ghzlatarev please correct me.

Dengjianping:
One more question, should I add it to manta now?

runtime/calamari/src/diff_gas_fees.rs
ferrell-code: Yeah correct but `MinimumMultipler` makes it so almost the entire cost of the tx is in the base fee, if the `Multiplier` increases from congestion the weight of the extrinsic becomes much more heavily weighted in the calculation of the tx fee

runtime/calamari/src/diff_gas_fees.rs
```diff
@@ -0,0 +1,2169 @@
...
+        let call = crate::Call::Utility(pallet_utility::Call::force_batch { calls: vec![] });
+        let (dispatch_info, call_len) = get_call_details(&call);
+        calamari_runtime_calls.push(("pallet_utility", "force_batch", dispatch_info, call_len));
+    }
+
+    let mut all_extrinsics_gas_fees = vec![];
+    t.execute_with(|| {
+        for (pallet_name, extrinsic_name, dispatch_info, call_len) in calamari_runtime_calls {
+            let fee = TransactionPayment::compute_fee(call_len, &dispatch_info, 0);
+            let float_gax_fees = Multiplier::try_from(fee).unwrap().div(decimal).to_float();
```
Garandor: This does integer division and creates a float after, truncating at DECIMAL.
Not a big deal i guess but it could raise false positives for small/cheap TXNs
maybe convert to float before dividing

Garandor: also `gax`

runtime/calamari/src/diff_gas_fees.rs
```diff
@@ -0,0 +1,2169 @@
...
+        let (dispatch_info, call_len) = get_call_details(&call);
+        calamari_runtime_calls.push((
+            "pallet_parachain_staking",
+            "cancel_candidate_bond_less",
+            dispatch_info,
+            call_len,
+        ));
+
+        // delegate
+        let call = crate::Call::ParachainStaking(pallet_parachain_staking::Call::delegate {
```
Garandor: This method is great, but it can't substitute weight comparison as it is.

In this TX you check for almost the cheapes TX possible and guarantee that this *specific* tx signature is not changed.
What about a delegate call with `candidate_delegation_count: 5000` and `delegation_count:10000` ?
This is part of the dynamic component of the weights and changes in that are not captured this way, not allowing to determine whether the extrinsic with above parameters has changed in weight or not.

To get "good enough"-level coverage i guess we could have multiple invocations per extrinsic
- the "base case" simplest possible call as it is now
- One call with each of the dynamic weight components ( as specified in the benchmarking fn for this extrinsic) at its max/a very large setting
- Technically we'd also need all permutations of high/low values for each dynamic parameter, but that explodes the  complexity so i'd recommend to leave that out

For this example this would be
- pallet_parachain_staking::Call::delegate(ALICE, 1,2,3)
- pallet_parachain_staking::Call::delegate(ALICE, 1,25,3)
- pallet_parachain_staking::Call::delegate(ALICE, 1,2,100)
- (optionally, but leave it out) pallet_parachain_staking::Call::delegate(ALICE, 1,25,100)

Max values taken from `benchmarks.rs` and calamari `lib.rs` config items

Sry for this additional effort :|

Garandor (dismissed):
need to expand coverage of multiple extrinsic call params

runtime/calamari/src/diff_gas_fees.rs
Garandor: `Multiplier::one()` seems more reasonable, i agree

runtime/calamari/src/diff_gas_fees.rs
Dengjianping: I will add one more filed that would be more readable, but for fees comparison, I will use that tx fees without removing decimal.

Dengjianping:
> need to expand coverage of multiple extrinsic call params

I would cover those extrinsics which are open to users, but like governance extrinsics, I will ignore them, just leave one case there.

ghzlatarev (approved):

ghzlatarev (dismissed):

ghzlatarev (dismissed):

ferrell-code (dismissed):

ghzlatarev (commented):
Linter is broken, otherwise lgtm
https://github.com/Manta-Network/Manta/actions/runs/5285745613/jobs/9564520599?pr=1083

ghzlatarev:
Also the unit tests are failing after the latest manta-sbt change

ghzlatarev (approved):

ghzlatarev (commented):
Seems we need to bump the timeouts of the integration tests jobs from 180 to 240 or something like that.

ghzlatarev (approved):

Garandor (approved):

PR #1084, opened 2023-04-24
Use safe-math in xcm `buy_weight` impl

ghzlatarev:
## Description

* This was not going to happen with current values but just in case for future additions we protect against accidental overflows.
* The division is impossible to underflow because the dividend will be saturated to the max value of u128. And u128::Max can be divided by the second smallest u128 number.
* A divide by zero is possible but we can assert that `WEIGHT_PER_SECOND` is never 0 with an integration test.

---

Situational Notes:

### DIFF
--- a/primitives/manta/src/xcm.rs
+++ b/primitives/manta/src/xcm.rs
@@ -180,9 +180,8 @@ where
                     XcmError::TooExpensive
                 })?;
 
-                let amount = (units_per_second.saturating_mul(weight as u128))
-                    .checked_div(WEIGHT_PER_SECOND as u128)
-                    .ok_or(XcmError::TooExpensive)?;
+                let amount =
+                    (units_per_second.saturating_mul(weight as u128)) / (WEIGHT_PER_SECOND as u128);
 
                 // we don't need to proceed if amount is zero.
                 // This is very useful in tests.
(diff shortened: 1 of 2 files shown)


ghzlatarev:
Question: is it actually impossible to get an underflow from the division ? I'm not sure about that, but we can use `divide_checked`

primitives/manta/src/xcm.rs
```diff
@@ -180,7 +180,9 @@ where
                     XcmError::TooExpensive
                 })?;
 
-                let amount = units_per_second * (weight as u128) / (WEIGHT_PER_SECOND as u128);
+                let amount =
+                    (units_per_second.saturating_mul(weight as u128)) / (WEIGHT_PER_SECOND as u128);
```
ferrell-code, line 184: still uses `\`, could be `saturating_div`

primitives/manta/src/xcm.rs
```diff
@@ -248,8 +250,9 @@ where
     fn refund_weight(&mut self, weight: Weight) -> Option<MultiAsset> {
         if let Some((id, prev_amount, units_per_second)) = &mut self.refund_cache {
             let weight = weight.min(self.weight);
-            self.weight -= weight;
-            let amount = *units_per_second * (weight as u128) / (WEIGHT_PER_SECOND as u128);
+            self.weight = self.weight.saturating_sub(weight);
+            let amount =
+                ((*units_per_second).saturating_mul(weight as u128)) / (WEIGHT_PER_SECOND as u128);
```
ferrell-code, line 255: here too

ferrell-code (approved):
one comment fine as is

ferrell-code (approved):

Garandor (approved):
this division can't overflow, so it's fine as is

ferrell-code:
Good point, why does `saturating_div` even exist?

Garandor:
> Good point, why does `saturating_div` even exist?

Dividing by a float [0,1) increases the value so it can overflow in the positive direction which is handled by saturating...
on signed types it can also overflow in the negative direction in the same way.
Also, division-by-zero is a panic condition.
Georgi uses a `u128` type and checks for nonzero of the divisor in the integration test, so *for these inputs* `saturating_div` and `/` have identical semantics

PR #1086, opened 2023-04-26
Add SBT Registry

### LINKED ISSUE #1081: Create New zkSBT MintTypes Without a Runtime Upgrade
Right now to add new mint type we need update this enum: https://github.com/Manta-Network/Manta/blob/ceb9e46cd53b77eb914ba6c17452fc238bc3a28f/pallets/manta-sbt/src/lib.rs#L134

We should add an `MintType` registry to add new zkSBT `MintTypes` onchain dynamically. Migrating existing storage will be necessary, I think doing it lazily using `on_idle` hook is best way to go due to number of existing zkSBT mints already completed

ferrell-code:
## Description

Adds a `MintRegistry` and identifies mint types using `u32`. 

Migrates storage to new scheme using on_idle hook.

Manually bumps `NextMintId` to 3 using `on_runtime_upgrade`

closes: #1081 

---

Situational Notes:


runtime/calamari/src/migrations/mod.rs
```diff
@@ -15,5 +15,6 @@
 // along with Manta.  If not, see <http://www.gnu.org/licenses/>.
 
 pub mod asset_id;
+pub mod init_sbt_counter;
```
ghzlatarev, line 18: what do you guys think about commenting out the executed migrations, so we dont have to maintain

pallets/manta-sbt/src/tests.rs
```diff
@@ -450,25 +458,28 @@ fn allowlist_account_works() {
...
+            EvmAccountAllowlist::<Test>::get(bab_id, EvmAddress::default()).unwrap(),
             MintStatus::Available(1)
         );
 
         assert_noop!(
             MantaSBTPallet::allowlist_evm_account(
                 MockOrigin::signed(ALICE),
-                EvmAddressType::Bab(EvmAddress::default())
+                bab_id,
+                EvmAddress::default(),
```
ghzlatarev, line 473: is this default only for tests or is it actually needed in production ?
If it's not needed we should make this for tests only.

pallets/manta-sbt/src/lib.rs
```diff
@@ -239,23 +279,57 @@ pub mod pallet {
...
+        MintId,
+        RegisteredMint<Moment<T>, T::RegistryBound>,
+        OptionQuery,
+    >;
+
     /// SBT Metadata maps `StandardAsset` to the correstonding SBT metadata
     ///
     /// Metadata is raw bytes that correspond to an image
+    ///
+    /// Deprecated delete after migration
```
ghzlatarev, line 332: issue for this.

pallets/manta-sbt/src/lib.rs
```diff
@@ -497,37 +573,128 @@ pub mod pallet {
...
 
+    #[pallet::hooks]
+    impl<T: Config> Hooks<T::BlockNumber> for Pallet<T>
+    where
+        T::AccountId: From<AccountId> + Into<AccountId>,
+    {
+        /// Migrates sbt pallet to new storage scheme
+        ///
+        /// remove once migration is complete
+        fn on_idle(_n: T::BlockNumber, remaining_weight: Weight) -> Weight {
```
ghzlatarev, line 656: when is the on-idle actually executed.
after the extrinsics phase, if there's more weight available ?

ghzlatarev (commented):
... didn't click submit on Saturday, so these were all pending.

pallets/manta-sbt/src/lib.rs
ferrell-code: IIRC, After the extrinsics and it is the last hook executed other than the `on_finalize` hook

pallets/manta-sbt/src/tests.rs
ferrell-code: I used default in tests where the value of `EvmAddress` was irrelevant.

I don't think there is a way to feature gate the Default trait as it originates from `H160` type. Shouldn't use default in prod

pallets/manta-sbt/src/lib.rs
ferrell-code: #1090

runtime/calamari/src/migrations/mod.rs
ferrell-code: I think that is a good solution, could do in #1090

pallets/manta-sbt/src/lib.rs
ferrell-code: https://github.com/paritytech/substrate/blob/e699560d2037d155e6054c05b6334a6e9fa64926/frame/support/src/traits/hooks.rs#L77

pallets/manta-sbt/src/tests.rs
ghzlatarev: Oh it's not our own impl, ok then,

ghzlatarev (approved):

zqhxuyuan (approved):

PR #1104, opened 2023-05-12
Fix off by 1 erorrs in pull-ledger-diff rpc

ghzlatarev:
## Description

* The total receivers calculation in pull-ledger-diff rpc is off by one, because we start indexing from 0 not 1.
* The exception is for the first 256 utxos, for which we can use the sum of the pulled receivers and senders

---

Situational Notes:


pallets/manta-pay/src/lib.rs
```diff
@@ -610,10 +610,18 @@ pub mod pallet {
...
                 .map(|i| ShardTrees::<T>::get(i).current_path.leaf_index as u128)
-                .sum::<u128>()
-                + NullifierSetSize::<T>::get() as u128;
+                .sum::<u128>();
+            // Workaround for the fact that we index the receivers from 0
+            if senders_receivers_total == 0u128 {
+                // The cast is fine because the vector sizes are limited by PULL_MAX_RECEIVER_UPDATE_SIZE and PULL_MAX_SENDER_UPDATE_SIZE
+                senders_receivers_total = receivers.len() as u128;
+            } else {
+                senders_receivers_total += 256u128;
```
ferrell-code, line 621: should this be `saturating_add`?

Not sure if an overflow is possible, so may be fine as is

pallets/manta-pay/src/lib.rs
```diff
@@ -610,10 +610,18 @@ pub mod pallet {
...
+            }
+            senders_receivers_total += NullifierSetSize::<T>::get() as u128;
```
ferrell-code, line 623: same here

ghzlatarev: Each tree has 2^20 leaves max and there are 256 trees, so in total 268_435_456 utxos.
NullifierSetSize is u64 so even at max the sum shouldn't go over u128::max

ferrell-code (approved):

DanielZhangReal (approved):

PR #1106, opened 2023-05-12
Remove timeout for check_tests pipeline

Dengjianping:
## Description

---

Situational Notes:

### DIFF
--- a/.github/workflows/check_tests.yml
+++ b/.github/workflows/check_tests.yml
@@ -16,7 +16,6 @@ env:
 jobs:
   start-unit-test-checks:
     runs-on: ubuntu-20.04
-    timeout-minutes: 150
     outputs:
       runner-label: ${{ steps.start-self-hosted-runner.outputs.runner-label }}
       aws-region: ${{ steps.start-self-hosted-runner.outputs.aws-region }}


ghzlatarev (approved):

ferrell-code (commented):
Can't we just make it a larger number like 300?

Dengjianping:
> 

In last PR #1080, I had a issue about sscache, I thought I could extend the timeout to fix it, but failed. Finally, Georgi increased the size of cache, and I separated features to solve it. So now, it's unnecessary now.

ferrell-code (approved):

Garandor:
> > 
> 
> In last PR #1080, I had a issue about sscache, I thought I could extend the timeout to fix it, but failed. Finally, Georgi increased the size of cache, and I separated features to solve it. So now, it's unnecessary now.

@Dengjianping If it's not necessary to remove the time limit, why did you merge it? I don't like the thought of having one or more perpetually stuck runners over time

ghzlatarev:
> > > 
> > 
> > 
> > In last PR #1080, I had a issue about sscache, I thought I could extend the timeout to fix it, but failed. Finally, Georgi increased the size of cache, and I separated features to solve it. So now, it's unnecessary now.
> 
> @Dengjianping If it's not necessary to remove the time limit, why did you merge it? I don't like the thought of having one or more perpetually stuck runners over time

We didn't use the limit here and the amount that i added was just killing the jobs prematurely.
So that's why we removed it.
We can add a proper limit in a different PR, but it was out of scope for this one.

Garandor:
> We didn't use the limit here and the amount that i added was just killing the jobs prematurely. So that's why we removed it. We can add a proper limit in a different PR, but it was out of scope for this one.

This statement is self-contradictory, if it killed stuff it was obviously "being used", no?
I do get that the tests run longer than 2 hrs so the limit is too low, i agree there.
Can we get this readded for the next release? @Dengjianping / @ghzlatarev could you propose a new PR?
It can be as simple as readding the deleted line with something like 300min timeout.

ghzlatarev:
> > We didn't use the limit here and the amount that i added was just killing the jobs prematurely. So that's why we removed it. We can add a proper limit in a different PR, but it was out of scope for this one.
> 
> This statement is self-contradictory, if it killed stuff it was obviously "being used", no? I do get that the tests run longer than 2 hrs so the limit is too low, i agree there. Can we get this readded for the next release? @Dengjianping / @ghzlatarev could you propose a new PR? It can be as simple as readding the deleted line with something like 300min timeout.

I meant we didn't use a time limit in this workflow before the last commit which added the limit for an unrelated reason, which it didn't affect in the end. But yeah definitely a proper limit for maybe 2.5 -3 hours would be useful.

PR #1107, opened 2023-05-15
Add ledger total count api

DanielZhangReal:
## Description

Add ledger total count api

Goes together with https://github.com/Manta-Network/sdk/pull/128

Refer to this PR: https://github.com/Manta-Network/Manta/pull/1109

---

Situational Notes:


pallets/manta-sbt/src/lib.rs
```diff
@@ -885,21 +897,11 @@ where
...
-        let mut receivers_total = (0..=255)
-            .map(|i| ShardTrees::<T>::get(i).current_path.leaf_index as u128)
-            .sum::<u128>();
-        // Workaround for the fact that we index the receivers from 0
-        if receivers_total == 0u128 {
-            // The cast is fine because the vector sizes are limited by PULL_MAX_RECEIVER_UPDATE_SIZE
-            receivers_total = receivers.len() as u128;
-        } else {
-            receivers_total += 256u128;
-        }
```
ghzlatarev, line 897: where is this part ?

DanielZhangReal: // In the initial state of the ledger, the total value will be 256;
// if we want to get an accurate value, we need to request `pull_receivers` to fix this value;
// but a simple total count interface does not need to add more complicated logic.

I added some notes on `pull_ledger_total_count`

ghzlatarev: i'd leave the logic here, to make it easier on the users.

DanielZhangReal: Make sure `pull_ledger_total_count` is consistent with `senders_receivers_total` returned by `pull_ledger_diff`, directly returns the value of `pull_ledger_total_count`

ghzlatarev: Sorry i meant where is the other part `(if total_senders_receivers == 256) total_senders_receivers = receivers.count()`

DanielZhangReal: I removed this logic, The above explains why we don't need this logic

DanielZhangReal: I don’t think it’s necessary to add that logic for a new chain with only a few ledger data. If you think we need to add, then I may need to request `pull_receivers` on the `pull_ledger_total_count` interface to be compatible with the correct data, as I said above.

ghzlatarev: So you're essentially leaving to the caller ?
It seems better to do the correction here, to make it easier for the users of this RPC. That was the idea initially.
Or am i missing something else ?

DanielZhangReal: This is what I considered when adding `pull_ledger_total_count`

ghzlatarev: im fine either way, let's see what the others say

Apokalip: I think either way is fine, just make sure to optimize easier usage of RPCs

ferrell-code: Looks fine to me, early value doesn't seem to be relevant.

ferrell-code (commented):
Looks good to me.

Need to make ci happy

PR #1112, opened 2023-05-18
Manta assets genesis storage migration

ghzlatarev:
## Description

* The problem: because we added AssetManager and Assets after genesis on Manta runtime, genesis storage config like registering the native asset and setting next-asset-id to 8, were not executed. Then when we registered USDT and DOT they took up the wrong asset ids 0 and 1 instead of 8 and 9
* Add a storage migration to reset the Manta runtime AssetManager and Assets storage. No actual storage changes though.
* Firsts clear out all the storage items in both AssetManager and Storage
* Then register the Native Asset and set AssetManager next-asset-id storage item to 8
* Perform some asserts like checking that before we had 2 items in the maps and after the migration there's only 1 and that next-asset id is correct. 
* Update try-runtime CI and the result is https://github.com/Manta-Network/Manta/actions/runs/5020904261:
```
2023-05-19 06:12:43        API/INIT: Not decorating unknown runtime apis: 0xf3ff14d5ab527059/1, 0x2aa62120049dd2d2/1
2023-05-19 06:12:44 since no prefix is filtered, the data for all pallets will be downloaded    
2023-05-19 06:12:44 scraping key-pairs from remote at block height 0x8abe986dacd5d46004c6255c57829e6b50a2248644f9f9c4ac0a7213754709fb    
2023-05-19 06:12:45 Querying a total of 2566 keys from prefix , splitting among 2 threads, 1283 keys per thread    
2023-05-19 06:12:47 adding data for hashed prefix: , took 3s    
2023-05-19 06:12:47 adding data for hashed key: 3a636f6465    
2023-05-19 06:12:47 adding data for hashed key: 26aa394eea5630e07c48ae0c9558cef7f9cce9c888469bb1a0dceaa129672ef8    
2023-05-19 06:12:48 adding data for hashed key: 26aa394eea5630e07c48ae0c9558cef702a5c1b19ab7a04f536c519aca4983ac    
2023-05-19 06:12:48 initialized state externalities with storage root 0x17bdd2603410aa55e24be67cde4d7b1c784958f3a2859b21438bcfe75520dd8f and state_version V1    
2023-05-19 06:12:48 original spec: RuntimeString::Owned("manta")-4060, code hash: 4bde474bbe3f7bb78669eb48648bf238110f690c144c097988314989bf9b3d4e    
2023-05-19 06:12:48 new spec: RuntimeString::Owned("manta")-4070, code hash: 57[33](https://github.com/Manta-Network/Manta/actions/runs/5020904261/jobs/9002789410#step:12:34)790b1d7014c[35](https://github.com/Manta-Network/Manta/actions/runs/5020904261/jobs/9002789410#step:12:36)eae977ee873ea5c5a0e7fe6e698bcb599ed718ca8a[42](https://github.com/Manta-Network/Manta/actions/runs/5020904261/jobs/9002789410#step:12:43)6a2    
2023-05-19 06:12:49 ✅ Migrations pre checks finished!    
2023-05-19 06:12:49 Starting migration for AssetManager...    
2023-05-19 06:12:49 ✅ Storage migration for AssetManager has been executed successfully and storage version has been update to: 1.    
2023-05-19 06:12:49 Starting migration for pallet-assets...    
2023-05-19 06:12:49 ✅ Storage migration for Assets has been executed successfully and storage version has been update to: 1.    
2023-05-19 06:12:49 ✅ Migrations post checks finished!    
2023-05-19 06:12:49 ⚠️ ParachainSystem declares internal migrations (which *might* execute). On-chain `StorageVersion(1)` vs current storage version `StorageVersion(2)`    
2023-05-19 06:12:[49](https://github.com/Manta-Network/Manta/actions/runs/5020904261/jobs/9002789410#step:12:50) ⚠️ XcmpQueue declares internal migrations (which *might* execute). On-chain `StorageVersion(1)` vs current storage version `StorageVersion(2)`    
2023-05-19 06:12:49 ⚠️ PolkadotXcm declares internal migrations (which *might* execute). On-chain `StorageVersion(0)` vs current storage version `StorageVersion(0)`    
2023-05-19 06:12:49 ⚠️ DmpQueue declares internal migrations (which *might* execute). On-chain `StorageVersion(0)` vs current storage version `StorageVersion(1)`    
2023-05-19 06:12:49 TryRuntime_on_runtime_upgrade executed without errors. Consumed weight = ([50](https://github.com/Manta-Network/Manta/actions/runs/5020904261/jobs/9002789410#step:12:51)0800000000 ps, [52](https://github.com/Manta-Network/Manta/actions/runs/5020904261/jobs/9002789410#step:12:53)42880 byte), total weight = (500000000000 ps, 5242880 byte) (100.16 %, 100.00 %).
```
* command used: `./target/release/manta try-runtime --runtime target/release/wbuild/manta-runtime/manta_runtime.compact.compressed.wasm --chain manta-dev -linfo on-runtime-upgrade --checks=all live --at 0x061574be3ce30662ea2ba4442997589c25559b0e40e233a0448f2a8e42e14a06 --uri wss://ws.manta.systems:443`

---

Situational Notes:


.github/workflows/try-runtime-mainnet.yml
```diff
@@ -8,9 +8,9 @@ on:
         description: The WebSocket url of the reference node
         default: wss://ws.calamari.systems:443
         required: true
-      chain_spec:
-        description: The chain spec of the local node
-        default: calamari-dev
+      runtime:
+        description: The runtime of the local node - use `manta` or `calamari` only.
+        default: calamari
```
Garandor: let's try to use `manta` as a default setting for everything from now on

runtime/manta/src/migrations/assets_genesis.rs
```diff
@@ -0,0 +1,397 @@
...
+            || assets_storage_version != INITIAL_PALLET_ASSETS_VERSION
+        {
+            log::info!("Aborting migration due to unexpected on-chain storage versions for pallet-assets-manager: {:?} and pallet-assets: {:?}. Expectation was: {:?} and {:?}.", asset_manager_storage_version, assets_storage_version, INITIAL_PALLET_ASSETS_MANAGER_VERSION, INITIAL_PALLET_ASSETS_VERSION );
+            return T::DbWeight::get().reads(2);
+        }
+
+        log::info!(target: "asset-manager", "Starting migration for AssetManager...");
+
+        let pallet_prefix: &[u8] = b"AssetManager";
+        let _ = clear_storage_prefix(pallet_prefix, b"AssetIdLocation", b"", None, None);
```
Garandor, line 68: Are you being explicit for documentation sake or does
`let _ = clear_storage_prefix(pallet_prefix, b"", b"", None, None);` not work?

runtime/manta/src/migrations/assets_genesis.rs
```diff
@@ -0,0 +1,397 @@
...
+        let _ = clear_storage_prefix(pallet_prefix, b"AssetIdMetadata", b"", None, None);
+        let _ = clear_storage_prefix(pallet_prefix, b"UnitsPerSecond", b"", None, None);
+        let _ = clear_storage_prefix(pallet_prefix, b"MinXcmFee", b"", None, None);
+        let _ = clear_storage_prefix(pallet_prefix, b"AllowedDestParaIds", b"", None, None);
+
+        put_storage_value(
+            pallet_prefix,
+            b"NextAssetId",
+            b"",
+            <T::AssetConfig as AssetConfig<T>>::StartNonNativeAssetId::get(),
```
Garandor, line 79: this puts `8` in here, you probably wanted `StartNativeAssetId`

Garandor: ok nvm, i see the native assert is registered here manually

runtime/manta/src/migrations/assets_genesis.rs
```diff
@@ -0,0 +1,397 @@
...
+        put_storage_value(
+            pallet_prefix,
+            b"LocationAssetId",
+            &Blake2_128Concat::hash(&location.encode()),
+            asset_id,
+        );
+        put_storage_value(
+            pallet_prefix,
+            b"AssetIdMetadata",
+            &Blake2_128Concat::hash(&asset_id.encode()),
```
Garandor, line 99: `setAssetIdMetadata` has a component that writes some metadata in `pallet-assets`. You're not writing anything in `pallet-assets` in the migration later, i think this will leave this metadata incomplete

Garandor: hm, asset manager genesis build does the same thing
https://github.com/Manta-Network/Manta/blob/dc29283c438d99cf9b1cefe23d9e9ccf0e9c7617/pallets/asset-manager/src/lib.rs#L184

Garandor: the part that will be missing for the native asset is https://github.com/Manta-Network/Manta/blob/dc29283c438d99cf9b1cefe23d9e9ccf0e9c7617/runtime/calamari/src/assets_config.rs#L121

Garandor: idk if that means it has always been buggy or it's safe for some reason

runtime/manta/src/migrations/assets_genesis.rs
```diff
@@ -0,0 +1,397 @@
...
+        let pallet_prefix: &[u8] = b"Assets";
+        let _ = clear_storage_prefix(pallet_prefix, b"Asset", b"", None, None);
+        let _ = clear_storage_prefix(pallet_prefix, b"Metadata", b"", None, None);
+        let _ = clear_storage_prefix(pallet_prefix, b"Account", b"", None, None);
+        let _ = clear_storage_prefix(pallet_prefix, b"Approval", b"", None, None);
+
+        log::info!(target: "assets", "✅ Storage migration for Assets has been executed successfully and storage version has been update to: {:?}.", INITIAL_PALLET_ASSETS_MANAGER_VERSION + 1);
+
+        StorageVersion::new(INITIAL_PALLET_ASSETS_VERSION + 1).put::<pallet_assets::Pallet<T>>();
+
```
Garandor, line 119: Why not register the DOT and USDT assets while you're at it? Saves us another multisig coordination headache

runtime/manta/src/migrations/assets_genesis.rs
ghzlatarev: It doesn't unfortunately or i'm misusing it somehow :/

ghzlatarev: Lemme try again though, before i resolve this comment.

.github/workflows/try-runtime-mainnet.yml
ghzlatarev: Good point

runtime/manta/src/migrations/assets_genesis.rs
ghzlatarev: Yeah better to be explicit and not use these magic numbers.

runtime/manta/src/migrations/assets_genesis.rs
ghzlatarev: Correct.
This is actually equivalent in Calamari. The Assets pallet doesn't know anything about the native asset.
This has added benefit of the MANTA asset not showing up in Assets page in the front-end (with 0 issuance)

runtime/manta/src/migrations/assets_genesis.rs
ghzlatarev: I want to keep the migration work low, to review easier and also to not spend more dev time.
Was thinking the mutli-sig would be easier.

runtime/manta/src/migrations/assets_genesis.rs
ghzlatarev: No. We're setting the native asset with the migraiton so the next-asset-id should be the next non-native id.

runtime/manta/src/migrations/assets_genesis.rs
ghzlatarev: Nope :/

ferrell-code (commented):
lgtm, we should make sure to have at least two people to check the storage migration went well on staging when we start the release

ghzlatarev:
> lgtm, we should make sure to have at least two people to check the storage migration went well on staging when we start the release

Indeed

ferrell-code (approved):

runtime/manta/src/migrations/assets_genesis.rs
Garandor: What i'm hoping to get here is assurance that this lack-of-asset-metadata-on-an-otherwise-normally-registered-asset can not lead to some sort of runtime crash somewhere down the road because other code may rely on this existing.

If you can with certainty state this won't happen i'll trust in that, but if not we should have a meeting next week to make sure everyone fully understands the implications of this. whether this is already deployed or not

ghzlatarev: I mean i haven't done formal verification or something, but there are no panics in the assets code, so we should be safe from random crashes.
Only the asset manager and xcm are using the assets pallet, and they both won't have an issue handling missing asests during lookups. For example you remember TooExpensive xcm errors.

ghzlatarev: And if we do add it the total balance will always be 0 (wrong) , and only valid part would be the metadata.

Dengjianping (approved):

PR #1121, opened 2023-05-23
Cleanup Integration Test Import

### LINKED ISSUE #1118: Change MANTA as KMA import in integration tests
https://github.com/Manta-Network/Manta/blob/55f15d96c2056d379f24223e52db4904387e9d42/runtime/integration-tests/src/integrations_mock/mod.rs#L54

This line is confusing and could cause bad tests to be run due to the non-intuitiveness of it. We should use UNIT or some other moniker

ferrell-code:
## Description

close: #1118 

---

Situational Notes:

### DIFF
--- a/runtime/integration-tests/src/integrations_mock/mod.rs
+++ b/runtime/integration-tests/src/integrations_mock/mod.rs
@@ -48,10 +48,11 @@ cfg_if::cfg_if! {
         type RuntimeAssetConfig = calamari_runtime::assets_config::CalamariAssetConfig;
         type RuntimeConcreteFungibleLedger =
             calamari_runtime::assets_config::CalamariConcreteFungibleLedger;
+        const UNIT: Balance = KMA;
     } else {
         pub mod test_manta;
         use manta_runtime::{
-            currency::MANTA as KMA,
+            currency::MANTA,
             fee::{FEES_PERCENTAGE_TO_AUTHOR, FEES_PERCENTAGE_TO_TREASURY},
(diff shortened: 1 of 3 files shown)


runtime/integration-tests/src/integrations_mock/mod.rs
```diff
@@ -48,10 +48,11 @@ cfg_if::cfg_if! {
         type RuntimeAssetConfig = calamari_runtime::assets_config::CalamariAssetConfig;
         type RuntimeConcreteFungibleLedger =
             calamari_runtime::assets_config::CalamariConcreteFungibleLedger;
+        const UNIT: Balance = KMA;
     } else {
         pub mod test_manta;
         use manta_runtime::{
-            currency::MANTA as KMA,
+            currency::MANTA,
```
Garandor, line 55: why not just `use currency::MANTA as UNIT` ?

ferrell-code: Well the thought is that then `KMA` can be used in calamari specific tests and `MANTA` can be used in manta tests and then `UNIT` is used for common tests.

But maybe just changing all tests to use `UNIT` is cleaner. I am cool w either way

Garandor: I'd personally prefer `UNIT` across all tests, but that's just my preference. Ok leaving as it is

Garandor (approved):

Apokalip (approved):
LGTM

PR #1127, opened 2023-05-26
Add check that public asset is restricted

ghzlatarev:
## Description

* Add a check that the public asset of a receiver post is not unrestricted, but the default. Otherwise we give a potential attacker control over the nullifier values for no particular reason.
* Currently there's no obvious attack vector but if we end up implementing an exchange where we allow some rules (like swapping the public assets of Utxos) then things may start getting dangerous
* Test cases

---

Situational Notes:


pallets/manta-pay/src/lib.rs
```diff
@@ -270,6 +283,15 @@ pub mod pallet {
                     && post.sink_accounts.is_empty(),
                 Error::<T>::InvalidShape
             );
+            for post in post.receiver_posts.clone() {
```
zqhxuyuan: is it better to use iter() instead of clone()?

ghzlatarev: indeed.

pallets/manta-pay/src/lib.rs
```diff
@@ -236,6 +243,12 @@ pub mod pallet {
                     && post.sink_accounts.len() == 1,
                 Error::<T>::InvalidShape
             );
+            if !post.receiver_posts[0].utxo.is_transparent {
+                ensure!(
+                    post.receiver_posts[0].utxo.public_asset == Asset::default(),
```
ferrell-code: Does Asset not implement `zero()`?

ferrell-code (commented):
one question looks good

pallets/manta-pay/src/lib.rs
ghzlatarev: No but yeah i think its better to implement and avoid default()

ferrell-code (dismissed):

ferrell-code (dismissed):

.github/workflows/run_all_benchmarks.yml
```diff
@@ -67,6 +67,7 @@ jobs:
           rustup toolchain install nightly-2023-03-03
           rustup default nightly-2023-03-03
           rustup target add wasm32-unknown-unknown
+          rustup default nightly-2023-01-01
```
ferrell-code: I think accidental duplicate here

.github/workflows/run_linters.yml
```diff
@@ -78,7 +78,7 @@ jobs:
           curl -s https://sh.rustup.rs -sSf | sh -s -- -y
           source ${HOME}/.cargo/env
           rustup toolchain install nightly-2023-03-13
-          rustup default nightly-2023-03-13
```
ferrell-code: Not sure if this change is onpurpose

ferrell-code (approved):

Apokalip (approved):

PR #1138, opened 2023-06-01
Check manta and calamari lease expiration

Dengjianping:
## Description
- If only two leases remains, the test will fail. One lease on kusama is 42 days, 84 days on polkadot.
```
// kusama
leasePeriod = 604800 // blocks
oneLease = leasePeriod / (3600 * 24) = 42

// polkadot
leasePeriod = 1209600
oneLease = leasePeriod / (3600 * 24) = 84
```
- If not, will print how the lease has many days remaining.
- Calamari lease checking is being skipping, because it will must fail for now.

---

Situational Notes:


tests/check_parachain_lease_expiration.ts
```diff
@@ -0,0 +1,78 @@
...
+            const remainingDays = leasePeriod * relaychainBlockTime * remainingLeaseLength.length / oneDay;
+            console.log(`Manta parachain lease has ${remainingDays} days remaining.`);
+        } else {
+            const currentLeaseRemainedBlocks = leasePeriod - firstBlock;
+            const remainingDays = leasePeriod * relaychainBlockTime * (remainingLeaseLength.length - 1) / oneDay + (currentLeaseRemainedBlocks * relaychainBlockTime / oneDay);
+            console.log(`Manta parachain lease has ${remainingDays} days remaining.`);
+        }
+    }).timeout(2000000);
+
+    it.skip('Check Calamari Parachain Lease', async () => {
```
ghzlatarev: what does this `skip` mean

tests/check_parachain_lease_expiration.ts
```diff
@@ -0,0 +1,78 @@
...
+        const bestNumber = (await polkadotApi.derive.chain.bestNumber()).toNumber();
+        const startNumber = bestNumber - polkadotApi.consts.slots.leaseOffset.toNumber();
+        const leasePeriod = polkadotApi.consts.slots.leasePeriod.toNumber();
+        const currentPeriod = Math.floor(startNumber / leasePeriod);
+
+        const remainingLeaseLength = await polkadotApi.query.slots.leases(mantaParaId);
+        const endPeriod = currentPeriod + remainingLeaseLength.length - 1;
+
+        await polkadotApi.disconnect();
+        assert(remainingLeaseLength.length >= 1, 'Manta lease is going to expire, please do auction.');
```
ghzlatarev: id make this 2 to be on the safe side

tests/check_parachain_lease_expiration.ts
ferrell-code: This skips calamari test? I suppose it would fail right now so it needs to be activated when calamari is back online?

Dengjianping: Yes, it will fail. I have an issue #1139  to track it, once we get the slot again, this test case will be enabled.

ferrell-code (dismissed):

tests/check_parachain_lease_expiration.ts
Garandor: i dont get how this fails at all, the only output i see is a `console.log`

tests/check_parachain_lease_expiration.ts
```diff
@@ -0,0 +1,78 @@
...
+        const firstBlock = startNumber % leasePeriod;
+        if (firstBlock === 0) {
+            const remainingDays = leasePeriod * relaychainBlockTime * remainingLeaseLength.length / oneDay;
+            console.log(`Manta parachain lease has ${remainingDays} days remaining.`);
+        } else {
+            const currentLeaseRemainedBlocks = leasePeriod - firstBlock;
+            const remainingDays = leasePeriod * relaychainBlockTime * (remainingLeaseLength.length - 1) / oneDay + (currentLeaseRemainedBlocks * relaychainBlockTime / oneDay);
+            console.log(`Manta parachain lease has ${remainingDays} days remaining.`);
+        }
+    }).timeout(2000000);
```
Garandor: whats the timeout number for, maybe declare this as a named var?

tests/check_parachain_lease_expiration.ts
ferrell-code: It has `it.skip(` to avoid failure right now. 

Will need to be update when we get calamari a slot again

Garandor: can you explain how it will fail? there's no return value

ferrell-code: Line 65

```ts
 assert(remainingLeaseLength.length >= 1, 'Calamari lease is going to expire, please do auction.');
```

tests/check_parachain_lease_expiration.ts
Dengjianping: I think the problem is the test framework `mocha` doesn't support `async` well, so I have to add a timeout for each test case. And `2000` sec is unreasonable, now it's 5 minutes.

tests/check_parachain_lease_expiration.ts
Dengjianping: This test case will be not executed. I will enable it after calamari is alive again, and there's an issue to track it.

tests/check_parachain_lease_expiration.ts
Dengjianping: Last lease for calamari is 42 days, 84 days for manta. Not sure it's too early to trigger it if  two leases remain.

ferrell-code: I think 2 is good, we can always disable this ci test once we are reminded

Dengjianping: Now it's 2.

ferrell-code (dismissed):

Apokalip (dismissed):

Garandor (approved):

ferrell-code (approved):

Issue #624, opened 2022-06-21
Adapt all our pallets to new BoundedVec ( tentatively polkadot v0.9.25 )


Garandor:
https://github.com/paritytech/substrate/pull/11645

Garandor:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Runtime-635aba84a463c200116d03d9/issues/Manta-Network/Manta/624?planning-poker) with Zenhub @ghzlatarev @zqhxuyuan @Dengjianping @ferrell-code

Btw. When you're voting, consider other already estimated issues as guideposts, i'll try making this more concrete tomorrow

Garandor:
This is also possibly audit-relevant but we'll only pick it up if doable

PR #1141, opened 2023-06-05
manta-farming

zqhxuyuan:
## Description
farming protocol:
- deposit: user deposit token into farming pool, and get share
- withdraw: user remove share from farming pool
- claim: user can get rewards when they stake token into farming pool
- withdraw_claim: user can un-stake token from farming pool.

---

Situational Notes:


pallets/farming/src/lib.rs
```diff
@@ -0,0 +1,820 @@
...
+                *id = id.checked_add(1).ok_or(ArithmeticError::Overflow)?;
+                Ok(())
+            })?;
+
+            Self::deposit_event(Event::FarmingPoolCreated { pid });
+            Ok(())
+        }
+
+        #[pallet::call_index(1)]
+        #[pallet::weight(0)]
```
ghzlatarev: 0 weight ?

pallets/farming/Cargo.toml
```diff
@@ -0,0 +1,63 @@
+[package]
+authors = ['Manta Network']
+description = 'Pallet implementing farming protocol.'
+edition = "2021"
+homepage = "https://manta.network"
+license = "GPL-3.0"
+name = "manta-farming"
+repository = 'https://github.com/Manta-Network/Manta/'
+version = "4.0.8"
```
Dengjianping: It shoud be 4.1.0 now.

pallets/farming/rpc/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+authors = ['Manta Network']
+edition = "2021"
+homepage = "https://manta.network"
+license = "GPL-3.0"
+name = "manta-farming-rpc-api"
+repository = 'https://github.com/Manta-Network/Manta/'
+version = "4.0.8"
```
Dengjianping: Same here

pallets/farming/rpc/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+use sp_api::ProvideRuntimeApi;
+use sp_blockchain::HeaderBackend;
+use sp_rpc::number::NumberOrHex;
+use sp_runtime::{generic::BlockId, traits::Block as BlockT};
+
+#[rpc(client, server)]
+pub trait FarmingRpcApi<BlockHash, AccountId, CurrencyId, PoolId> {
+    /// rpc method for getting farming rewards
+    #[method(name = "farming_getFarmingRewards")]
+    fn get_farming_rewards(
```
Dengjianping: If it runs without `blocking`, I suppose this method can be `async`.
https://docs.rs/jsonrpsee-proc-macros/0.18.2/jsonrpsee_proc_macros/attr.rpc.html#method-attribute

pallets/farming/src/gauge.rs
```diff
@@ -0,0 +1,494 @@
...
+    BlockNumberFor: Default,
+{
+    fn new(who: AccountIdOf) -> Self {
+        Self {
+            who,
+            gauge_amount: Default::default(),
+            total_time_factor: Default::default(),
+            latest_time_factor: Default::default(),
+            claimed_time_factor: Default::default(),
+            gauge_start_block: Default::default(),
```
Dengjianping, line 53: Because `BlockNumberFor` derives from `Default`, I suppose we can save some line of code here by
```
..Default::default()
```

pallets/farming/src/gauge.rs
```diff
@@ -0,0 +1,494 @@
...
+            pool_info.keeper.clone(),
+            pool_info.reward_issuer.clone(),
+            gauge_basic_rewards,
+            max_block,
+            current_block_number,
+        );
+
+        GaugePoolInfos::<T>::insert(gid, &gauge_pool_info);
+        GaugePoolNextId::<T>::mutate(|id| -> DispatchResult {
+            *id = id.checked_add(1).ok_or(ArithmeticError::Overflow)?;
```
Dengjianping: Can yo use One::one()?

pallets/farming/src/gauge.rs
```diff
@@ -0,0 +1,494 @@
...
+
+impl<BalanceOf, BlockNumberFor, AccountIdOf> GaugeInfo<BalanceOf, BlockNumberFor, AccountIdOf>
+where
+    BalanceOf: Default + HasCompact,
+    BlockNumberFor: Default,
+{
+    fn new(who: AccountIdOf) -> Self {
+        Self {
+            who,
+            gauge_amount: Default::default(),
```
Dengjianping, line 49: I'm not against `Default` in runtime, but yeah, we should be careful about these `Default`.

pallets/farming/src/gauge.rs
```diff
@@ -0,0 +1,494 @@
...
+                );
+
+                ensure!(
+                    gauge_pool_info.max_block
+                        >= gauge_info.gauge_stop_block - gauge_info.gauge_start_block + gauge_block,
+                    Error::<T>::GaugeMaxBlockOverflow
+                );
+
+                let incease_total_time_factor = if gauge_info.gauge_amount.is_zero() {
+                    gauge_info.gauge_stop_block = current_block_number;
```
Dengjianping: If `claimed_time_factor` still uses default `claimed_time_factor`, does it affect gauge claiming?
Because I found this field is not being set.

pallets/farming/src/gauge.rs
```diff
@@ -0,0 +1,494 @@
...
+        let total_shares = U256::from(pool_info.total_shares.to_owned().saturated_into::<u128>());
+        let mut result_vec = Vec::<(CurrencyIdOf<T>, BalanceOf<T>)>::new();
+
+        pool_info.rewards.iter().try_for_each(
+            |(reward_currency, (total_reward, total_withdrawn_reward))| -> DispatchResult {
+                let withdrawn_reward = share_info
+                    .withdrawn_rewards
+                    .get(reward_currency)
+                    .copied()
+                    .unwrap_or_default();
```
Dengjianping: Does `unwrap_or_default` mean `0`?

pallets/farming/src/gauge.rs
```diff
@@ -0,0 +1,494 @@
...
+
+                let total_reward_proportion: BalanceOf<T> =
```
Dengjianping: I feel confused about the way to calculate portion, Why is it a integer instead if percent?
Dividing `total_shares` should return a a float number (0, 1), right?

pallets/farming/src/lib.rs
```diff
@@ -0,0 +1,839 @@
...
+        pub fn create_farming_pool(
+            origin: OriginFor<T>,
+            tokens_proportion: Vec<(CurrencyIdOf<T>, Perbill)>,
+            basic_rewards: Vec<(CurrencyIdOf<T>, BalanceOf<T>)>,
+            gauge_init: Option<GaugeInitType<T>>,
+            min_deposit_to_start: BalanceOf<T>,
+            #[pallet::compact] after_block_to_start: BlockNumberFor<T>,
+            #[pallet::compact] withdraw_limit_time: BlockNumberFor<T>,
+            #[pallet::compact] claim_limit_time: BlockNumberFor<T>,
+            withdraw_limit_count: u8,
```
Dengjianping: #[pallet::compact]

pallets/farming/src/lib.rs
```diff
@@ -0,0 +1,839 @@
...
+    {
+        /// `ControlOrigin` create the farming pool.
+        #[pallet::call_index(0)]
+        #[pallet::weight(T::WeightInfo::create_farming_pool())]
+        pub fn create_farming_pool(
+            origin: OriginFor<T>,
+            tokens_proportion: Vec<(CurrencyIdOf<T>, Perbill)>,
+            basic_rewards: Vec<(CurrencyIdOf<T>, BalanceOf<T>)>,
+            gauge_init: Option<GaugeInitType<T>>,
+            min_deposit_to_start: BalanceOf<T>,
```
Dengjianping: #[pallet::compact]

pallets/farming/src/lib.rs
```diff
@@ -0,0 +1,839 @@
...
+                    &mut pool_info,
+                    gauge_token,
+                    gauge_basic_rewards_map,
+                    max_block,
+                )?;
+            };
+
+            PoolInfos::<T>::insert(pid, &pool_info);
+            PoolNextId::<T>::mutate(|id| -> DispatchResult {
+                *id = id.checked_add(1).ok_or(ArithmeticError::Overflow)?;
```
Dengjianping: One::one()

pallets/farming/src/lib.rs
```diff
@@ -0,0 +1,839 @@
...
+                Self::create_gauge_pool(
+                    pid,
+                    &mut pool_info,
+                    gauge_token,
+                    gauge_basic_rewards_map,
+                    max_block,
+                )?;
+            };
+
+            PoolInfos::<T>::insert(pid, &pool_info);
```
Dengjianping: Do we need to check whether it's a duplicated pool?

pallets/farming/src/lib.rs
```diff
@@ -0,0 +1,839 @@
...
+            })?;
+
+            Self::deposit_event(Event::FarmingPoolCreated { pid });
+            Ok(())
+        }
+
+        /// After create farming pool, need to deposit reward token into the pool.
+        #[pallet::call_index(1)]
+        #[pallet::weight(0)]
+        pub fn charge(
```
Dengjianping, line 393: So anyone here can deposit reward tokens multiple times, right?

pallets/farming/src/lib.rs
```diff
@@ -0,0 +1,839 @@
...
+            origin: OriginFor<T>,
+            pid: PoolId,
+            rewards: Vec<(CurrencyIdOf<T>, BalanceOf<T>)>,
```
Dengjianping, line 396: Some one might split reward tokens into like 1000 rewards here, like
```
[
    (1, 1),
    (1, 1),
    (1, 1),
    ...
]
```

pallets/farming/src/lib.rs
```diff
@@ -0,0 +1,839 @@
...
+                .iter()
+                .try_for_each(|(reward_currency, reward)| -> DispatchResult {
+                    T::MultiCurrency::transfer(
+                        *reward_currency,
+                        &exchanger,
+                        &pool_info.reward_issuer,
+                        *reward,
+                    )
+                })?;
+            pool_info.state = PoolState::Charged;
```
Dengjianping, line 411: If user can deposit rewards multiple times, seems it's unnecessary to insert pool_info

pallets/farming/src/lib.rs
```diff
@@ -0,0 +1,839 @@
...
+        pub fn close_pool(origin: OriginFor<T>, pid: PoolId) -> DispatchResult {
+            T::ControlOrigin::ensure_origin(origin)?;
+
+            let mut pool_info = Self::pool_infos(pid).ok_or(Error::<T>::PoolDoesNotExist)?;
+            ensure!(
+                PoolState::state_valid(Action::ClosePool, pool_info.state),
+                Error::<T>::InvalidPoolState
+            );
+
+            pool_info.state = PoolState::Dead;
```
Dengjianping: What does dead mean? Does it mean no one use it any more?

runtime/calamari/src/lib.rs
```diff
@@ -790,6 +794,29 @@ impl calamari_vesting::Config for Runtime {
...
+/// Zenlink protocol Asset adaptor for orml_traits::MultiCurrency.
+type MantaCurrencies = Currencies<Runtime, assets_config::CalamariAssetConfig, Balances, Assets>;
+
+impl manta_farming::Config for Runtime {
+    type RuntimeEvent = RuntimeEvent;
+    type CurrencyId = CalamariAssetId;
+    type MultiCurrency = MantaCurrencies;
+    type ControlOrigin = EitherOfDiverse<
+        EnsureRoot<AccountId>,
+        pallet_collective::EnsureProportionAtLeast<AccountId, TechnicalCollective, 2, 3>,
```
Dengjianping, line 817: Not sure it's reasonable if TC controls related farming extrinsics.

pallets/farming/src/gauge.rs
ghzlatarev, line 1: btw can someone explain what is the word gauge supposed to mean in this context :D

pallets/farming/src/tests.rs
```diff
@@ -0,0 +1,611 @@
...
+#[test]
+fn claim() {
+    let keeper_account = FarmingKeeperPalletId::get().into_account_truncating();
+
+    ExtBuilder::default()
+        .one_hundred_for_alice_n_bob()
+        .build()
+        .execute_with(|| {
+            let (pid, _tokens) = init_gauge_1000();
+            // assert_eq!(Farming::shares_and_withdrawn_rewards(pid, &ALICE), ShareInfo::default());
```
ghzlatarev: there's commented code in these tests

pallets/farming/src/tests.rs
```diff
@@ -0,0 +1,611 @@
...
+                RuntimeOrigin::signed(ALICE),
+                pid,
+                Some(800)
+            ));
+            assert_eq!(Assets::balance(KSM, &ALICE), 4698);
+
+            // Claim again
+            System::set_block_number(System::block_number() + 6);
+            assert_ok!(Farming::claim(RuntimeOrigin::signed(ALICE), pid));
+            assert_err!(
```
ghzlatarev: not all errors are covered: PoolDoesNotExist, GaugePoolNotExist, GaugeInfoNotExist, ShareInfoNotExists, LastGaugeNotClaim, GaugeMaxBlockOverflow,

pallets/farming/src/tests.rs
```diff
@@ -0,0 +1,611 @@
...
+                Some(200)
+            ));
+            System::set_block_number(System::block_number() + 100);
+            assert_ok!(Farming::withdraw_claim(RuntimeOrigin::signed(ALICE), pid));
+            assert_eq!(Assets::balance(KSM, &ALICE), 4166);
+            assert_ok!(Farming::claim(RuntimeOrigin::signed(ALICE), pid));
+            assert_eq!(Farming::shares_and_withdrawn_rewards(pid, &ALICE), None);
+            assert_eq!(Assets::balance(KSM, &ALICE), 4166);
+            // let ed = <Runtime as Config>::MultiCurrency::minimum_balance(KSM);
+            // assert_eq!(Assets::balance(KSM, &TREASURY_ACCOUNT), ed);
```
ghzlatarev: commented code

pallets/farming/src/tests.rs
```diff
@@ -0,0 +1,611 @@
...
+                RuntimeOrigin::signed(ALICE),
+                pid,
+                Some(200)
+            ));
+            System::set_block_number(System::block_number() + 100);
+            assert_ok!(Farming::withdraw_claim(RuntimeOrigin::signed(ALICE), pid));
+            assert_eq!(Assets::balance(KSM, &ALICE), 4166);
+            assert_ok!(Farming::claim(RuntimeOrigin::signed(ALICE), pid));
+            assert_eq!(Farming::shares_and_withdrawn_rewards(pid, &ALICE), None);
+            assert_eq!(Assets::balance(KSM, &ALICE), 4166);
```
ghzlatarev: im not sure there's a cap but we never test deposting/withdrawing the minimums/maximums

pallets/farming/src/tests.rs
```diff
@@ -0,0 +1,611 @@
...
+            assert_eq!(Assets::balance(KSM, &ALICE), 2000);
+            Farming::on_initialize(0);
+            Farming::on_initialize(0);
+            System::set_block_number(System::block_number() + 1);
+            assert_ok!(Farming::withdraw(
+                RuntimeOrigin::signed(ALICE),
+                pid,
+                Some(800)
+            ));
+            assert_err!(
```
ghzlatarev: we should add an error case where we try to withdraw more than we staked

ghzlatarev (commented):
Also is there an e2e test for QA-ing this ?

pallets/farming/src/gauge.rs
ferrell-code: https://resources.curve.fi/reward-gauges/understanding-gauges

might help, to my understanding its like a gauge for rewards

pallets/farming/src/gauge.rs
zqhxuyuan: The struct itself is not implements Default, compile error when change:
```
the trait `core::default::Default` is not implemented for `gauge::GaugeInfo<BalanceOf, BlockNumberFor, AccountIdOf>`
```

pallets/farming/src/gauge.rs
zqhxuyuan: yes, the default value here is 0 based on the type is Balance. default value of zero is fine here, as when reward is zero, it didn't effect the final result.

pallets/farming/src/gauge.rs
zqhxuyuan: total_reward_proportion = user_share * total_reward / total_share = (user_share/total_share)*total_reward. so it should be reward amount, not percent.

zqhxuyuan: PS: also refactor the reward algorithms into lib.rs

pallets/farming/src/lib.rs
zqhxuyuan: this `charge` operation transfer sender token into reward issuer, unless someone want to donate money to use, this normally is executed by our account so that we can send reward token to user.
Yes, this can execute multiple times, for example, when reward is not enough, we need to inject new rewards into reward issuer account.

pallets/farming/src/lib.rs
zqhxuyuan: same above, if someone want to send token to us, it's fine.

pallets/farming/src/lib.rs
zqhxuyuan: insert operation here is to change the PoolState.
There's a situation that when pool state is already `Ongoing`, then this charge operation change state to `Charged`, but it's fine, because on next block, it'll go back to `Ongoing` again by `on_initialize`.

pallets/farming/src/lib.rs
zqhxuyuan: Suppose we don't want to incentive a farming pool, we need first close the pool, then we can retire and kill the pool.
In Dead state, user can still claim rewards and withdraw tokens that staked into farming. but can't deposit token anymore.

pallets/farming/src/lib.rs
ghzlatarev: could be used for ddos ?

pallets/farming/src/lib.rs
zqhxuyuan: no, We may support incentive multiple pool for one token, for example user stake MANTA, can get different types of reward token from different pool.

pallets/farming/src/lib.rs
zqhxuyuan: add weight to `Charge` should fix that.

pallets/farming/rpc/src/lib.rs
zqhxuyuan: done.

pallets/farming/src/lib.rs
zqhxuyuan: add weight to `charge`.

pallets/farming/src/tests.rs
zqhxuyuan: added a testcase that total share is 1000, first withdraw 800, then withdraw 300.

pallets/farming/src/tests.rs
zqhxuyuan: add `precondition_check_should_work()` testcase.

primitives/manta/src/currencies.rs
```diff
@@ -0,0 +1,185 @@
...
+        }
+        Ok(())
+    }
+
+    fn can_slash(currency_id: Self::CurrencyId, who: &C::AccountId, value: Self::Balance) -> bool {
+        if value == Zero::zero() {
+            return true;
+        }
+        Self::free_balance(currency_id, who) >= value
+    }
```
ghzlatarev, line 171: what is slashing in the context of farming ?

pallets/farming/src/lib.rs
```diff
@@ -0,0 +1,927 @@
...
+    pub type PoolInfos<T: Config> = StorageMap<_, Twox64Concat, PoolId, PoolInfoOf<T>>;
+
+    /// Record gauge farming pool info.
+    ///
+    /// map PoolId => GaugePoolInfo
+    #[pallet::storage]
+    #[pallet::getter(fn gauge_pool_infos)]
+    pub type GaugePoolInfos<T: Config> = StorageMap<_, Twox64Concat, PoolId, GaugePoolInfoOf<T>>;
+
+    ///
```
ghzlatarev, line 246: missing docs

primitives/manta/src/currencies.rs
zqhxuyuan: it's not used by farming. orml::MultiCurrency need to implements `can_slash` and `slash`.

primitives/manta/src/currencies.rs
```diff
@@ -0,0 +1,185 @@
...
+
+    fn transfer(
+        currency_id: Self::CurrencyId,
+        from: &C::AccountId,
+        to: &C::AccountId,
+        amount: Self::Balance,
+    ) -> DispatchResult {
+        if currency_id == A::NativeAssetId::get() {
+            Native::transfer(from, to, amount, ExistenceRequirement::KeepAlive)?;
```
ghzlatarev: Maybe this should be AllowDeath ? Wouldn't it help with https://github.com/Manta-Network/Manta/issues/1172 actually

ghzlatarev: Also on the withdraw, althought it doesn't seem to be used

zqhxuyuan: If we support staking native token, then Stake all operation will make account as
 dust account. I think this has issue that user can't execute furthure transaction as user don't have native token to paying fees.
I'll add a testcase for this case.

pallets/farming/src/lib.rs
```diff
@@ -0,0 +1,927 @@
...
+    /// double_map (PoolId, AccountId) => ShareInfo
+    #[pallet::storage]
+    #[pallet::getter(fn shares_and_withdrawn_rewards)]
+    pub type SharesAndWithdrawnRewards<T: Config> =
+        StorageDoubleMap<_, Twox64Concat, PoolId, Twox64Concat, T::AccountId, ShareInfoOf<T>>;
+
+    #[pallet::hooks]
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        fn on_initialize(n: BlockNumberFor<T>) -> Weight {
+            PoolInfos::<T>::iter().for_each(|(pid, mut pool_info)| match pool_info.state {
```
ghzlatarev: nit: pid -> pool_id

pallets/farming/src/lib.rs
```diff
@@ -0,0 +1,927 @@
...
+
+            Self::deposit_event(Event::WithdrawClaimed {
+                who: exchanger,
+                pid,
+            });
+            Ok(())
+        }
+
+        #[pallet::call_index(6)]
+        #[pallet::weight(0)]
```
ghzlatarev: still dont get why these weights are 0

pallets/farming/src/gauge.rs
```diff
@@ -0,0 +1,494 @@
...
+        gauge_basic_rewards: BTreeMap<CurrencyIdOf<T>, BalanceOf<T>>,
+        max_block: BlockNumberFor<T>,
+    ) -> DispatchResult {
+        let gid = Self::gauge_pool_next_id();
+        pool_info.gauge = Some(gid);
+        let current_block_number = frame_system::Pallet::<T>::block_number();
+        let gauge_pool_info = GaugePoolInfo::new(
+            pid,
+            gauge_token,
+            pool_info.keeper.clone(),
```
ghzlatarev, line 133: not sure this needs to be cloned

pallets/farming/src/gauge.rs
```diff
@@ -0,0 +1,415 @@
...
+            pool_info.reward_issuer.clone(),
```
ghzlatarev, line 134: dont think these need to be cloned

pallets/farming/src/gauge.rs
```diff
@@ -0,0 +1,494 @@
...
```
ghzlatarev, line 134: dont think these need to be cloned

pallets/farming/src/gauge.rs
```diff
@@ -0,0 +1,494 @@
...
+                    .ok_or(Error::<T>::ShareInfoNotExists)?;
+                gauge_pool_info.rewards.iter().try_for_each(
+                    |(
+                        reward_currency,
+                        (reward_amount, total_gauged_reward, _total_withdrawn_reward),
+                    )|
+                     -> DispatchResult {
+                        let reward = reward_amount
+                            .checked_sub(total_gauged_reward)
+                            .ok_or(ArithmeticError::Overflow)?;
```
ghzlatarev: Let's add an underflow error for these instead of reusing Overflow

pallets/farming/src/lib.rs
```diff
@@ -0,0 +1,930 @@
...
+    #[pallet::storage]
+    #[pallet::getter(fn pool_infos)]
+    pub type PoolInfos<T: Config> = StorageMap<_, Twox64Concat, PoolId, PoolInfoOf<T>>;
+
+    /// Record gauge farming pool info.
+    ///
+    /// map PoolId => GaugePoolInfo
+    #[pallet::storage]
+    #[pallet::getter(fn gauge_pool_infos)]
+    pub type GaugePoolInfos<T: Config> = StorageMap<_, Twox64Concat, PoolId, GaugePoolInfoOf<T>>;
```
ghzlatarev, line 243: Feel like this should be a bounded structure as there are for loops in the code that can explode
@Apokalip https://github.com/Manta-Network/Manta/issues/624

pallets/farming/src/lib.rs
```diff
@@ -0,0 +1,930 @@
...
+            #[pallet::compact] withdraw_limit_time: BlockNumberFor<T>,
+            #[pallet::compact] claim_limit_time: BlockNumberFor<T>,
+            #[pallet::compact] withdraw_limit_count: u8,
+        ) -> DispatchResult {
+            T::ControlOrigin::ensure_origin(origin)?;
+
+            let pid = Self::pool_next_id();
+            let keeper = T::Keeper::get().into_sub_account_truncating(pid);
+            let reward_issuer = T::RewardIssuer::get().into_sub_account_truncating(pid);
+            let basic_token = tokens_proportion[0];
```
ghzlatarev, line 342: we should maybe do a check that this is non-empty ?

pallets/farming/src/lib.rs
zqhxuyuan: admin operation can use weight=0. This's original code, but we can benchmark it and don't allow free exectution.

pallets/farming/src/gauge.rs
```diff
@@ -0,0 +1,415 @@
...
+        } else {
+            let time_factor_a = gauge_value
+                .saturated_into::<u128>()
+                .checked_mul(
+                    (gauge_info.gauge_stop_block - current_block_number).saturated_into::<u128>(),
+                )
+                .ok_or(ArithmeticError::Overflow)?;
+            let time_factor_b = gauge_block
+                .saturated_into::<u128>()
+                .checked_mul((gauge_value + gauge_info.gauge_amount).saturated_into::<u128>())
```
ghzlatarev: what about checked_add for the math in the parenthesis ?

pallets/farming/src/gauge.rs
zqhxuyuan: There's a mutation operation before this clone, need clone, otherwise has `Cannot move` error.

pallets/farming/src/gauge.rs
```diff
@@ -0,0 +1,415 @@
...
+    ) -> Result<(Perbill, u128), DispatchError> {
+        let current_block_number: BlockNumberFor<T> = frame_system::Pallet::<T>::block_number();
+        let start_block = if current_block_number > gauge_info.gauge_stop_block {
+            gauge_info.gauge_stop_block
+        } else {
+            current_block_number
+        };
+
+        let latest_claimed_time_factor = gauge_info.latest_time_factor
+            + gauge_info
```
ghzlatarev: safe math ?

pallets/farming/src/rewards.rs
```diff
@@ -0,0 +1,394 @@
...
+                            *withdrawn_reward = withdrawn_reward.saturating_sub(withdrawn_amount);
+                            Ok(())
+                        },
+                    )?;
+                    Ok(())
+                })?;
+
+                share_info
+                    .withdraw_list
+                    .push((n + withdraw_limit_time, remove_amount));
```
ghzlatarev, line 287: safe math ?

pallets/farming/src/lib.rs
ghzlatarev: we could leave it as a separate issue, but i'd give it some non 0 value at least.

pallets/farming/Cargo.toml
```diff
@@ -0,0 +1,63 @@
+[package]
+authors = ['Manta Network']
+description = 'Pallet implementing farming protocol.'
+edition = "2021"
+homepage = "https://manta.network"
+license = "GPL-3.0"
+name = "manta-farming"
```
ghzlatarev: pallet-farming is more consistent with teh rest of the pallets.

pallets/farming/src/gauge.rs
ghzlatarev: Yeah my bad.
Just wondering though, can we implement Copy on the Accoutnid type, it's just a [u8;32] after all

pallets/farming/src/rewards.rs
zqhxuyuan: block number should be safe here.

pallets/farming/src/lib.rs
Apokalip: @ghzlatarev Not sure how you would bound this without loosing the key feature of the StorageMap, do you have example code?

ghzlatarev: actually nvm, i forget this is a map.

pallets/farming/src/lib.rs
zqhxuyuan: make sense, change to 1000.

node/Cargo.toml
```diff
@@ -36,6 +36,8 @@ jsonrpsee = { version = "0.16.2", features = ["server"] }
 pallet-transaction-payment-rpc = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.37" }
 sc-transaction-pool-api = { git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.37" }
 
+pallet-farming-rpc-api = { path = "../pallets/farming/rpc" }
+pallet-farming-rpc-runtime-api = { path = "../pallets/farming/rpc/runtime-api" }
```
ghzlatarev, line 40: should log a task for integration tests of these new rpcs

pallets/farming/src/lib.rs
Apokalip: I mean in maps I can see it being used to limit either size of the object or if you have finite number of things under a key

node/Cargo.toml
zqhxuyuan: https://github.com/Manta-Network/Manta/issues/1183

pallets/farming/src/tests.rs
zqhxuyuan: add a `staking_minimum_kma()` to testing deposit/withdraw ED of KMA, and `staking_all_kma()` to deposit/withdraw all KMA.

runtime/calamari/src/lib.rs
zqhxuyuan: We can remove TC later on if we don't want TC to involve.

ghzlatarev (dismissed):

ghzlatarev:
Btw can you add a description with the major beats and why they are so

pallets/farming/src/gauge.rs
```diff
@@ -0,0 +1,426 @@
...
+        let current_block_number = frame_system::Pallet::<T>::block_number();
+        gauge_pool_info.gauge_last_block = current_block_number;
+        gauge_pool_info.gauge_amount = gauge_pool_info
+            .gauge_amount
+            .checked_add(&gauge_value)
+            .ok_or(ArithmeticError::Overflow)?;
+
+        ensure!(
+            gauge_info.gauge_stop_block >= current_block_number
+                || gauge_info.gauge_stop_block == Default::default(),
```
Dengjianping, line 200: What does `gauge_info.gauge_stop_block == Default::default()` mean here?
I suppose `Default::default()` is zero, right?

pallets/farming/src/lib.rs
```diff
@@ -0,0 +1,941 @@
...
+            T::ControlOrigin::ensure_origin(origin)?;
+
+            let mut pool_info = Self::pool_infos(pool_id).ok_or(Error::<T>::PoolDoesNotExist)?;
+            ensure!(
+                PoolState::state_valid(Action::RetirePool, pool_info.state),
+                Error::<T>::InvalidPoolState
+            );
+
+            let withdraw_limit_time = BlockNumberFor::<T>::default();
+            let retire_limit = RetireLimit::<T>::get();
```
Dengjianping: If I'm correct, you will get 0 if it doesn't set any value, because it will return default value.
Does it affect the rest of code?

zqhxuyuan: If `RetireLimit` is not set, calling `retire_pool` did't take any effect.

Apokalip: Wouldnt this be better set in the Config of the pallet, if we want to make sure its never 0 or Default?

pallets/farming/src/gauge.rs
zqhxuyuan: when initialize a new gauge info, `gauge_stop_block` value is Default. the `process_gauge_add()` will update to either current block number(if gauge_amount=0) or increment `gauge_block` based on old value.
The `LastGaugeNotClaim` error happens when `gauge_stop_block` < current block number.
Because we allow process_gauge_add in initial case. so we need to open gate for `gauge_stop_block` is default.

pallets/farming/src/lib.rs
zqhxuyuan: Using config is constant value, but we may change that limit number.
The retire operation is infrequent, we can add a precondition check.

ghzlatarev (approved):

Dengjianping (approved):

PR #1147, opened 2023-06-07
Add Polkadot Signature Allowlist

ferrell-code:
## Description

Gives ability to `AllowlistAccount` to reserve `AssetIds` for polkadot addresses for free.

Adds feature for `to_private` using signature to be relayed.

Adds check to disallow reserving new `AssetId` if `AccountId` already has `AssetIds` mapped to it.

---

Situational Notes:


pallets/manta-sbt/src/lib.rs
```diff
@@ -299,17 +316,30 @@ pub mod pallet {
...
         ///
         /// `TransferPost` is posted to private ledger and SBT metadata is stored onchain.
+        /// `signature` parameter can be used to relay a tx.
         #[pallet::call_index(0)]
         #[pallet::weight(<T as pallet::Config>::WeightInfo::to_private())]
         #[transactional]
         pub fn to_private(
-            who: OriginFor<T>,
+            origin: OriginFor<T>,
+            signature: Option<SignatureInfoOf<T>>,
```
zqhxuyuan, line 336: I think we need `mintId` as we also have different project on pomp.

pallets/manta-sbt/src/lib.rs
```diff
@@ -343,17 +373,28 @@ pub mod pallet {
         #[pallet::call_index(1)]
         #[pallet::weight(<T as pallet::Config>::WeightInfo::reserve_sbt())]
         #[transactional]
-        pub fn reserve_sbt(origin: OriginFor<T>) -> DispatchResult {
+        pub fn reserve_sbt(origin: OriginFor<T>, reservee: Option<T::AccountId>) -> DispatchResult {
             let who = ensure_signed(origin)?;
+            // Use reservee account, if None then use account whom signed transaction
+            let reserve_account = reservee.unwrap_or(who.clone());
+
+            let allowlist_account = AllowlistAccount::<T>::get();
```
zqhxuyuan: is it ok to add another account to avoid nonce issue as much as possible.

pallets/manta-sbt/src/lib.rs
```diff
@@ -299,17 +316,30 @@ pub mod pallet {
...
         #[transactional]
         pub fn to_private(
-            who: OriginFor<T>,
+            origin: OriginFor<T>,
+            signature: Option<SignatureInfoOf<T>>,
             post: Box<TransferPost>,
             metadata: BoundedVec<u8, T::SbtMetadataBound>,
         ) -> DispatchResultWithPostInfo {
-            let who = ensure_signed(who)?;
+            let mut minting_account = ensure_signed(origin)?;
```
zqhxuyuan, line 340: the relayer can also send to_private tx for user.

pallets/manta-sbt/src/lib.rs
ferrell-code: yeah that's a good idea

pallets/manta-sbt/src/lib.rs
ferrell-code: If we are storing data in the metadata we could also include the mint-id into the metadata

pallets/manta-sbt/src/lib.rs
```diff
@@ -343,17 +377,28 @@ pub mod pallet {
         #[pallet::call_index(1)]
         #[pallet::weight(<T as pallet::Config>::WeightInfo::reserve_sbt())]
         #[transactional]
-        pub fn reserve_sbt(origin: OriginFor<T>) -> DispatchResult {
+        pub fn reserve_sbt(origin: OriginFor<T>, reservee: Option<T::AccountId>) -> DispatchResult {
```
zqhxuyuan, line 392: Do we need change `reservee` to polkadot signature? ~~But even though we use signature here, user can still generate lots of different account and send it to relayer.~~(if using manta wallet, there'll be only one account for each wallet, so it should be fine using signature).
One thing I can think to prevent this case is make some rules on npo backend service just like `allowlist` that we're check eth contract.

pallets/manta-sbt/src/lib.rs
```diff
@@ -343,17 +377,28 @@ pub mod pallet {
...
-                &Self::account_id(),
-                T::ReservePrice::get(),
-                ExistenceRequirement::KeepAlive,
-            )?;
-
+            // ensure account does not have any `AssetId` already reserved
+            ensure!(
+                !ReservedIds::<T>::contains_key(&reserve_account),
+                Error::<T>::AssetIdsAlreadyReserved
+            );
```
ghzlatarev: Maybe we should do this check before the potential state change with the Currency::transfer

pallets/manta-sbt/src/lib.rs
ferrell-code: Either way user can generate infinite accounts we need backend to verify

pallets/manta-sbt/src/lib.rs
ferrell-code: doesn't really matter as it is transactional, but I am fine changing this as it may be a bit more readable

pallets/manta-sbt/src/lib.rs
```diff
@@ -613,6 +613,25 @@ pub mod pallet {
             Self::deposit_event(Event::<T>::ChangeFreeReserveAccount { account });
             Ok(())
         }
+
+        /// Remove allowlist evm account. Requires `AdminOrigin`
+        #[pallet::call_index(8)]
+        #[pallet::weight(<T as pallet::Config>::WeightInfo::change_allowlist_account())]
+        #[transactional]
+        pub fn remove_allowlist_evm_account(
```
ferrell-code, line 639: this is unneeded you can just use `None` in `change_allowlist_account`

zqhxuyuan: `change_allowlist_account` is used to set privilege account. while this `remove_allowlist_evm_account` used to remove user eth address storage, normally this is only used on test environment.

ferrell-code: Ah gotcha I misread this, is fine

zqhxuyuan:
We need `chain_id` because in dapp you have to change to some network to generate the signature, you can't just set chain_id as `0`(We have met this issue before in npo).

Also add `mint_id` and validation.

Dengjianping (dismissed):

zqhxuyuan (approved):

Dengjianping (approved):

ghzlatarev (approved):

PR #1167, opened 2023-06-20
Retire dolphin runtime

### LINKED ISSUE #958: Retire Dolphin Runtime
Our testnet is now using calamari RT and Dolphin does not represent the state of the chain due to e.g. missing staking

Can we retire this RT codebase altogether or do we need to keep it around?
Benefits of eliminating the code should be obvious, drawbacks not so much.

We'd need to merge dolphin and calamari integration tests

ghzlatarev:
## Description

closes: #958 

* Retire all Dolphin runtime related code, as it is not longer used for anything.
* check-tests was excluding pallet-manta-pay tests, when we only wanted to exclude the benchmark unit test (as our code relies on externally passed chain-spec file) . The fix is to not exclude in the CI but just hard-code disable the benchmark unit tests in the pallet code.
* gov v2 imports are required because of upstream code, so moved them outside of integrations test crate, as that crate is excluded in our unit tests job and it will fail compilation

---

Situational Notes:


ghzlatarev:
cargo clippy seems to be broken with any rust version

Dengjianping:
And we might delete dolphin docs from our manta docs

Dengjianping:
And some dolphin genesis files should be deleted

ferrell-code (commented):
looks good ci still seems to be failing

ghzlatarev:
> looks good ci still seems to be failing

Yeah i see why.
Something from upstream pulls the gov v2 pallets, which require us to also include them.
But if we include them in the integrations-test crate (and exclude it in the unit tests) then it won't compile.

This actually showed that we had broken our unit tests in another way. We wanted to exclude the manta-pay benchmarking unit test, but we had excluded all the unit tests. Fixed this now.

runtime/calamari/Cargo.toml
```diff
@@ -59,6 +59,10 @@ pallet-treasury = { git = "https://github.com/paritytech/substrate.git", default
 pallet-utility = { git = 'https://github.com/paritytech/substrate.git', default-features = false, branch = "polkadot-v0.9.37" }
 substrate-fixed = { git = "https://github.com/Manta-Network/substrate-fixed.git", default-features = false }
 
+pallet-conviction-voting = { git = "https://github.com/paritytech/substrate.git", default-features = false, branch = "polkadot-v0.9.37" }
```
Dengjianping, line 62: I know why you import these three pallet here, it's kind of annoying.

Dengjianping (approved):

Dengjianping:
I thought we would have faster ci check once dolphin runtime.
By the way, not sure it will affect change log generation, the dolphin tag `[DO]`. I suppose it won't.

runtime/calamari/Cargo.toml
ghzlatarev: yeah i just moved them from the integrations tests crate because it is excluded in some of our tests

ghzlatarev: added to the description

ghzlatarev:
> I thought we would have faster ci check once dolphin runtime. By the way, not sure it will affect change log generation, the dolphin tag `[DO]`. I suppose it won't.

no, we stopped Dolphin jobs some time ago already. This is just to reduce code base -8000 lines feels good

also re-includes mantapay in the unit tests job as they were not running for a few days

Apokalip (approved):

PR #1212, opened 2023-07-13
Add Permissionless Asset Registry Feature

### LINKED ISSUE #892: Make pallet-assets permissionless
Creating assets should be permissionless.
Reference implementation in Statemine.

ferrell-code:
## Description

Can use native token to create an asset for use on manta pay

Makes `AssetRegistryMetadata` a concrete type rather than generic

Creates an additional counter `NextPermissionlessAssetId` so the `AssetIds` of permissionless asset are separate from governance ones.

closes #892 

---

Situational Notes:


pallets/asset-manager/src/lib.rs
```diff
@@ -678,6 +721,77 @@ pub mod pallet {
             });
             Ok(())
         }
+
+        #[pallet::call_index(8)]
+        #[pallet::weight(T::WeightInfo::permissionless_register_asset())]
+        #[transactional]
+        pub fn permissionless_register_asset(
```
ghzlatarev, line 728: We should maybe decouple the mint into its own permissioned_mint function and add a flag for each asset whether  the asset is mintable or not after the initial creation.

pallets/asset-manager/src/tests.rs
```diff
@@ -647,3 +647,166 @@ fn register_lp_asset_should_work() {
...
+        );
+
+        // overflows when decimal is too high
+        assert_noop!(
+            AssetManager::permissionless_register_asset(
+                RuntimeOrigin::signed(ALICE),
+                "dog token".as_bytes().to_vec().try_into().unwrap(),
+                "dog".as_bytes().to_vec().try_into().unwrap(),
+                u8::MAX,
+                u128::MAX,
```
ghzlatarev: these 2 can be separate fail cases

ghzlatarev: we should also test the limits of the string

ferrell-code: a balance of u128::MAX doesn't fail, but agreed it is confusing

pallets/asset-manager/src/lib.rs
ferrell-code: Yeah that could be useful

pallets/asset-manager/src/tests.rs
ferrell-code: string limit is a little tricky to test as `BoundedVec` with too long of a string won't compile

I suppose trying to decode an encoded call with too long of a string could test

ghzlatarev: Oh wait it's BoundedVec, my bad. No need then.

pallets/asset-manager/src/lib.rs
ghzlatarev: Can be separate improvement issue.

ghzlatarev (dismissed):

runtime/calamari/src/assets_config.rs
```diff
@@ -163,18 +163,19 @@ impl AssetIdType for CalamariAssetConfig {
...
     type NativeAssetLocation = NativeAssetLocation;
     type NativeAssetMetadata = NativeAssetMetadata;
-    type StorageMetadata = AssetStorageMetadata;
     type AssetRegistry = CalamariAssetRegistry;
     type FungibleLedger = CalamariConcreteFungibleLedger;
 }
 
 impl pallet_asset_manager::Config for Runtime {
+    type PermissionlessStartId = ConstU128<1_000_000_000>;
+    type TokenNameMaxLen = ConstU32<200>;
```
Dengjianping: I think both constants should be aligned with pallet-assets, which has a limit on length of string
```rust
type StringLimit = ConstU32<50>;
```
Under the hood, we manipulate pallet-assets.

pallets/asset-manager/src/lib.rs
Dengjianping: Once user creates an asset, how can it be destroyed?

ferrell-code: User cannot destroy only a call from `pallet_assets` can do this. So would need to be governance call

runtime/calamari/src/assets_config.rs
ferrell-code: yeah good call now unified with pallet-assets. It would fail above pallet assets limit

Dengjianping:
Because new extrinsic has been added to this pallet, so you have to add it to tx fee diff test.
```
thread 'diff_tx_fees::diff_tx_fees' panicked at 'assertion failed: `(left == right)`
  left: `9`,
 right: `8`: Please update new extrinsic here.', runtime/calamari/src/diff_tx_fees.rs:1178:9
```

ferrell-code:
> Because new extrinsic has been added to this pallet, so you have to add it to tx fee diff test.
> 
> ```
> thread 'diff_tx_fees::diff_tx_fees' panicked at 'assertion failed: `(left == right)`
>   left: `9`,
>  right: `8`: Please update new extrinsic here.', runtime/calamari/src/diff_tx_fees.rs:1178:9
> ```

Done, also changed the README for the tx fee diff test if you want to take a look at that

ghzlatarev (dismissed):

ghzlatarev (approved):

Dengjianping (approved):

PR #1223, opened 2023-07-17
Block outgoing MANTA transfers on the XCM instruction level

ghzlatarev:
## Description

* Someone could've tried to bypass xTokens outgoing transfers filter by sending a `TransferReserveAsset` instruction with a top level paid XCM message.
* In order to prevent that we modify the `AllowTopLevelPaidExecutionFrom` and `AllowTopLevelPaidExecutionDescendOriginFirst` barriers to check for the `TransferReserveAsset` instruction after the `BuyExecution` and return error if it is present in the message
* Added a test case for to showcase this barrier
* We still need the xTokens outgoing transfers filter because that covers a different code path 

---

Situational Notes:


primitives/manta/src/xcm.rs
```diff
@@ -417,6 +417,69 @@ where
...
+    ) -> Result<(), ()> {
+        log::trace!(
+            target: "xcm::barriers",
+            "AllowTopLevelPaidExecutionFrom origin: {:?}, message: {:?}, max_weight: {:?}, weight_credit: {:?}",
+            origin, message, max_weight, _weight_credit,
+        );
+        ensure!(T::contains(origin), ());
+        let mut iter = message.0.iter_mut();
+        let i = iter.next().ok_or(())?;
+        match i {
```
Dengjianping, line 443: I'm not that familiar to xcm now, so my question is the 1st valid xcm instruction should start like `ReceiveTeleportedAsset`/`WithdrawAsset`/`ClaimAsset`, right?

primitives/manta/src/xcm.rs
```diff
@@ -417,6 +417,69 @@ where
...
+            ReceiveTeleportedAsset(..)
+            | WithdrawAsset(..)
+            | ReserveAssetDeposited(..)
+            | ClaimAsset { .. } => (),
+            _ => return Err(()),
+        }
+        let mut i = iter.next().ok_or(())?;
+        while let ClearOrigin = i {
```
Dengjianping, line 451: Why not to merge line 451-477 into like
```rust
while let Some(instruction) = iter().next() {
	match instruction {
		ClearOrigin => continue,
		BuyExecution { .. } => {
			// handle max weight
		}
		..
	}
}
```
I'm not sure there's another reason to divide such piece of code into many lines.

primitives/manta/src/xcm.rs
ghzlatarev: Yeah the first instruction must be one of these 3, because they actually put tokens into holding that can be used to pay for execution. (in the BuyExecution step)

primitives/manta/src/xcm.rs
ghzlatarev: This forces that there must be only ClearOrigin before any other instructions are executed.
So it force drops the origin of this message from a parachain sovereign account to nothing, in order to reduce the privilege of the message.
Don't think it can be combined in that match.

runtime/integration-tests/src/xcm_mock/xcm_tests.rs
```diff
@@ -3605,3 +3605,50 @@ fn send_disabled_asset_should_fail() {
...
+        let dummy_assets = MultiAssets::from(vec![dummy_asset.clone()]);
+        // Send withdraw and deposit
+        assert_ok!(ParachainPalletXcm::send_xcm(
+            Here,
+            (Parent, Parachain(PARA_B_ID)),
+            Xcm(vec![
+                WithdrawAsset(dummy_assets),
+                BuyExecution {
+                    fees: dummy_asset.clone(),
+                    weight_limit: Unlimited,
```
Dengjianping, line 3630: Can you cover `Limited` weight?

Dengjianping: Or you think it's unnecessary, this pr is fine to me

ghzlatarev: yeah it doesn't make a difference here, the thing we care about is that the TransferReserveAsset instruction never gets executed after the BuyExecution

Dengjianping (approved):

Apokalip (approved):

Garandor (approved):

